Issue #1, opened 2021-11-22
Publish weather-tools on PyPi


alxmrs:
This will help with distribution of the tool. Python users would only have to run `pip install weather-tools` to get the source and updates.

PR #31, opened 2021-12-21
Github action to publish package

saveriogzz:
Hi @alxmrs ! Thanks for this very nice project!!

I created a small github action to publish the package on PyPi after seen #1 . It's very minimal at the moment but could be an idea for the future!
In order for this to work, you should have a Twine username and pw within your repo's secrets.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,25 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  deploy:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.8'
+    - name: Install dependencies
+      run: |
+        pip install setuptools wheel twine
+    - name: Build and publish
+      env:
+        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWRD }}
+      run: |
+        python setup.py sdist
+        twine upload dist/*
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,3 +1,16 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 name: Upload Python Package
 
 on:
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,28 +11,92 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-name: Upload Python Package
+name: Build and Upload weather-tools to PyPI
 
 on:
   release:
     types: [created]
 
 jobs:
-  deploy:
+  build-artifacts:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-    - name: Set up Python
-      uses: actions/setup-python@v2
-      with:
-        python-version: '3.8'
-    - name: Install dependencies
-      run: |
-        pip install setuptools wheel twine
-    - name: Build and publish
-      env:
-        TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
-        TWINE_PASSWORD: ${{ secrets.PYPI_PASSWRD }}
-      run: |
-        python setup.py sdist
-        twine upload dist/*
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2.3.1
+        with:
+          python-version: 3.8
+
+      - name: Install dependencies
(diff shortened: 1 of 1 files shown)


alxmrs:
Hey @saveriogzz, thanks for your contribution! I'll give this change a review in a few hours. 

I really appreciate the help.

.github/workflows/publish.yml
```diff
@@ -0,0 +1,25 @@
+name: Upload Python Package
```
alxmrs, line 1: Please add the Apache License comment to the top of this YAML file (feel free to copy from the other workflow).

alxmrs: Nit: Can you specify where the package is uploaded to? i.e. `PyPi`

.github/workflows/publish.yml
```diff
@@ -0,0 +1,25 @@
...
+
+on:
+  release:
+    types: [created]
```
alxmrs, line 5: I think it's best to release on `published` instead of `created`. WDYT?

alxmrs (changes requested):
Hey @saveriogzz, thanks for your patch! A few general thoughts: 

It would be ideal to include tests for the success of publication to PyPi. It's possible that the tools don't install correctly once uploaded. Thus, it would be ideal for this workflow to tests that an installation round-trip works before publicizing. 

I think that [XArray's PyPi publication Action](https://github.com/pydata/xarray/blob/main/.github/workflows/pypi-release.yaml) is an excellent model for how we can test the release process. Aside from the changes that I listed in this review, consider taking leaves out of XArray's book. 

Last, our tools have an idiosyncratic project structure and type of packaging (a technical summary can be found [here](https://alex.merose.com/blog/beam-cli-hack.html), though it's slightly stale). With that, my final request is to add a check in this workflow to ensure that all the CLIs are runnable after installation from a repository (e.g. calling `weather-* --help` and checking the exit code is successful).  

Again, thank you so much for your contribution! Congratulations on being our first, non-internal contributor. 🥇 🎉

.github/workflows/publish.yml
saveriogzz: That depends! Type ‘published’ will be triggered if the release is published, no matter it’s draft or not. Type ‘created’ will be triggered when a NON-draft release is created and published. Reference: https://github.community/t/what-is-the-created-type-in-on-release/17184

alxmrs: Ah, that makes sense. I assumed that the opposite was the case (from the names only). Thanks for the reference! I agree with you, `created` is the way to go.

.github/workflows/publish.yml
saveriogzz: With the current configuration, the package will be uploaded on PyPi [twine](https://twine.readthedocs.io/en/stable/) takes care of looking for the PyPi's URL within the `dist` directory! You should however have a PyPi username and Password within your GitHub repository secrets.

I have just found a nice github action that looks really useful for us: https://github.com/pypa/gh-action-pypi-publish.
Under the hood, it uses twine!

saveriogzz:
> I think that [XArray's PyPi publication Action](https://github.com/pydata/xarray/blob/main/.github/workflows/pypi-release.yaml) is an excellent model for how we can test the release process.
> 
It's brilliant indeed! I'll soon adjust this action accordingly

.github/workflows/publish.yml
```diff
@@ -11,28 +11,92 @@
...
+          repository_url: https://test.pypi.org/legacy/
+          verbose: true
+
+      - name: Check uploaded package
+        if: github.event_name == 'push'
+        run: |
+          sleep 3
+          python -m pip install --upgrade pip
+          python -m pip install --extra-index-url https://test.pypi.org/simple --upgrade weather-tools
+          weather-dl --help && weather-mv --help && weather-sp --help
```
saveriogzz, line 86: These are simple sanity checks!

.github/workflows/publish.yml
```diff
@@ -11,28 +11,92 @@
...
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: releases
+          path: dist
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@v1.4.2
+        with:
+          user: __token__
+          password: ${{ secrets.PYPI_TOKEN }}
```
saveriogzz, line 101: N.B. This secret is different than [L76](https://github.com/google/weather-tools/pull/31/commits/6b511f4911934f8cce00537f1290fc5cd5e5579e#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R76)

alxmrs:
Thanks @saveriogzz! I'll TAL at the latest version starting next week. Until then, happy holidays!

.github/workflows/publish.yml
```diff
@@ -0,0 +1,102 @@
...
+          repository_url: https://test.pypi.org/legacy/
+          verbose: true
+
+      - name: Check uploaded package
+        if: github.event_name == 'push'
+        run: |
+          sleep 3
+          python -m pip install --upgrade pip
+          python -m pip install --extra-index-url https://test.pypi.org/simple --upgrade weather-tools
+          weather-dl --help && weather-mv --help && weather-sp --help
```
alxmrs, line 86: Nice! This all looks great.

.github/workflows/publish.yml
alxmrs: Cool, that makes sense to me.

alxmrs (approved):
LGTM! Before we merge, I'd like to test the action a bit.

alxmrs:
Hey @saveriogzz! I wanted to let you know the status of this PR. I think this is ready to merge, and I may do so in the next few days. However, we are still deciding on the name in PyPi to use.

saveriogzz:
Hey @alxmrs ! How about 'weather-beam'? 😁

alxmrs:
Not a bad idea. We were thinking of `google-weather-tools`. WDYT?

saveriogzz:
> Not a bad idea. We were thinking of `google-weather-tools`. WDYT?

Catchy! :)

PR #33, opened 2021-12-21
Handle NaNs in data

pramodg:


tox.ini
```diff
@@ -17,7 +17,7 @@ envlist = py37,py38,lint,type
 
 [testenv]
 deps = .[test]
-commands = pytest
+commands = pytest {posargs}
```
alxmrs, line 20: Can you explain why this was added? I don't understand what it does.

weather_mv/loader_pipeline/netcdf_loader_test.py
```diff
@@ -169,6 +169,23 @@ def test_extract_rows__no_duplicate_coordinates(self):
         counts = Counter([tuple(c.values()) for c in get_coordinates(ds)])
         self.assertTrue(all((c == 1 for c in counts.values())))
 
+    def test_extract_rows_nan(self):
+        self.test_data_path = f'{next(iter(weather_mv.__path__))}/test_data/test_data_has_nan.nc'
```
alxmrs, line 173: Nit: Can you extract the test_data folder into a variable?

weather_mv/loader_pipeline/netcdf_loader_test.py
```diff
@@ -169,6 +169,23 @@ def test_extract_rows__no_duplicate_coordinates(self):
...
+        actual = next(extract_rows(self.test_data_path))
+        print(actual)
```
alxmrs, line 175: Is this left over from debugging?

weather_mv/loader_pipeline/netcdf_loader.py
```diff
@@ -202,8 +203,10 @@ def to_row(it: t.Dict) -> t.Dict:
         row_ds = data_ds.loc[it]
 
         # Create a Name-Value map for data columns. Result looks like:
-        # {'d': -2.0187, 'cc': 0.007812, 'z': 50049.8}
-        row: t.Dict = row_ds.to_pandas().apply(to_json_serializable_type).to_dict()
+        # {'d': -2.0187, 'cc': 0.007812, 'z': 50049.8, 'rr': None}
+        temp_row = row_ds.to_pandas().apply(to_json_serializable_type)
+        # Pandas coerces floating type None values back to NaNs, need to do an explicit replace after.
+        row: t.Dict = temp_row.astype(object).where(pd.notnull(temp_row), None).to_dict()
```
alxmrs, line 209: I don't think the type annotation is necessary, since the RHS ends with `to_dict()`. WDYT?

alxmrs (changes requested):
Thanks for the fix! It generally looks good, just a few nits & questions.

tox.ini
pramodg: this allows to test a specific file / test fixture

weather_mv/loader_pipeline/netcdf_loader.py
pramodg: If you are referring to astype(object), that is the only way to make NaNs go to None.
  
Floating types are converted back to NaNs and behavior is inconsistent otherwise depending on pandas version
More details here, https://github.com/pandas-dev/pandas/issues/42423

alxmrs: That makes sense to me. 

I'm referring to the `t.Dict` annotation, specifically. Since the line ends in `.to_dict()`, it's clear to me that `row` is a dictionary without the annotation.

tox.ini
alxmrs: That seems useful! I'm ok with having it in this change. It can also be in its own CL.

weather_mv/loader_pipeline/netcdf_loader.py
pramodg: done

alxmrs (approved):
Thanks for the changes; the fix LGTM.

PR #41, opened 2021-12-24
Set up empty dataset, not table

lakshmanok:
We have to create an empty BigQuery dataset, not an empty table. Because an empty table won't have the right schema, the weather-mv pipeline fails. It happily creates the table if it doesn't exist, though.


alxmrs (approved):
Thanks Lak! LGTM.

Issue #55, opened 2021-12-30
Extend CI to test example configs are valid


alxmrs:
Hey @CillianFn and @pranay101, this issue was inspired by #52 and #47. If either of you would like to pick this up, that would be a great help!

To summarize: we should start treating our configs like code and include them in our automated checks. For this issue to be fixed, we should add to our CI build to test that all the example configs parse correctly (check for correct syntax). For now, I think it's sufficient to run each config with the `-d, --dry-run` flag. Later, we can check that the semantics of the configs are correct.

pranay101:
hey @alxmrs, it'll be great working on this issue. But this issue seems too technical to me right now can you please break it down a bit.

alxmrs:
Given that #56 is merged, we can call this issue fixed.

PR #56, opened 2021-12-30
Test example configs

CillianFn:
#55 

Added test for parsing config files (also included updates to existing config files based on missing required parameters). Contemplated adding a separate CI stage but it seemed sufficient to add a unit test. Calling `run` instead of `process_config` so the arg parsing is also covered in the test

Let me know what you think 👍


weather_dl/download_pipeline/pipeline_test.py
```diff
@@ -632,5 +635,19 @@ def test_target_name_date_and_target_directory_additional_partitions_in_path(sel
         self.assertEqual(target_name, "somewhere/expver-1/2017/01/15-pressure-500.nc")
 
 
+class ConfigTest(unittest.TestCase):
+
+    def setUp(self):
+        self._data_dir = f'{next(iter(weather_dl.__path__))}/../configs'
```
CillianFn, line 641: I feel like this could be cleaner but it was a bit fiddly getting the paths working both locally (Windows) & in CI. Used `weather_sp/test_data` as an example here but happy to change if there are any suggestions!

weather_dl/download_pipeline/pipeline_test.py
```diff
@@ -632,5 +635,19 @@ def test_target_name_date_and_target_directory_additional_partitions_in_path(sel
         self.assertEqual(target_name, "somewhere/expver-1/2017/01/15-pressure-500.nc")
 
 
+class ConfigTest(unittest.TestCase):
```
alxmrs, line 638: Let's move these config tests to a separate test file. To me, they are independent from the pipeline tests.

weather_dl/download_pipeline/pipeline_test.py
alxmrs: Yeah, I agree with you. The example in `weather_sp/test_data` is not the cleanest. I'm ok to merge this from now and address this path string all at once later.

alxmrs: nit: I see no need for data_dir to be private in these tests.

weather_dl/download_pipeline/pipeline_test.py
```diff
@@ -632,5 +635,19 @@ def test_target_name_date_and_target_directory_additional_partitions_in_path(sel
...
+
+    def setUp(self):
+        self._data_dir = f'{next(iter(weather_dl.__path__))}/../configs'
+
+    def test_process_config_files(self):
```
alxmrs, line 643: Ideally, I would like to see a separate test failure message for each config that fails. In other words, it would be nice to generate separate tests methods for each config in the directory. 

The simplest is to use unittest's [sub tests](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests).

weather_dl/download_pipeline/pipeline_test.py
```diff
@@ -632,5 +635,19 @@ def test_target_name_date_and_target_directory_additional_partitions_in_path(sel
...
+        for filename in os.listdir(self._data_dir):
+            config = os.path.join(self._data_dir, filename)
+            try:
+                run(["weather-dl", config, "--dry-run"])
+            except Exception as e:
+                assert False, ("\n" + type(e).__name__ + " " + str(e) + "\nIn " + filename)
```
alxmrs, line 649: This assertion statement is a bit messy. I'd suggest you consider applying one of two alternatives (the former is more conventional): 

- Instead of `assert False`, consider using `self.fail()`. Then, use an `f` string to template the failure statement
- ```
   raise AssertionError(f'Config {filename!r} is incorrect') from e
   ```

configs/yesterdays_surface_example.cfg
```diff
@@ -13,7 +13,10 @@
 # limitations under the License.
 
 [parameters]
-target_path=all.an
+client=fake
+partition_keys=
+    type
+target_path=all-{}.an
```
alxmrs, line 19: Glad we have these tests! It's good to catch these errors.

configs/seasonal_forecast_example_config.cfg
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 [parameters]
+client=fake
```
alxmrs, line 16: I think the client should be `cds`.

configs/yesterdays_surface_example.cfg
```diff
@@ -13,7 +13,10 @@
 # limitations under the License.
 
 [parameters]
-target_path=all.an
+client=fake
```
alxmrs, line 16: I think the client is `mars`.

alxmrs (changes requested):
Thanks @CillianFn for the patch. Your tests generally look good, and I'm really happy they caught problems in the configs.

weather_dl/download_pipeline/pipeline_test.py
CillianFn: Thanks for the tip! 

Note this required installation of [pytest-subtests](https://pypi.org/project/pytest-subtests/)

configs/yesterdays_surface_example.cfg
```diff
@@ -13,7 +13,10 @@
 # limitations under the License.
 
 [parameters]
-target_path=all.an
+client=mars
+partition_keys=
+    type
```
alxmrs, line 18: Looking at this again, isn't it possible to download everything into one chunk? If so, why do we need a partition at all? If this isn't supported by the parser / dry-runs, then I think it's a bug (we can raise an issue for later). 

Further, `type` is a strange thing to partition on. I think `date` or `param` would be better.

weather_dl/download_pipeline/config_test.py
```diff
@@ -0,0 +1,38 @@
...
+        self._data_dir = f'{next(iter(weather_dl.__path__))}/../configs'
+
+    def test_process_config_files(self):
+        for filename in os.listdir(self._data_dir):
+            with self.subTest(filename=filename):
+                config = os.path.join(self._data_dir, filename)
+                try:
+                    run(["weather-dl", config, "--dry-run"])
+                except Exception:
+                    self.fail(f'Config {filename!r} is incorrect')
```
alxmrs, line 34: Please end in a full stop (i.e. `.`).

weather_dl/download_pipeline/config_test.py
```diff
@@ -0,0 +1,38 @@
...
+    def setUp(self):
+        self._data_dir = f'{next(iter(weather_dl.__path__))}/../configs'
+
+    def test_process_config_files(self):
+        for filename in os.listdir(self._data_dir):
+            with self.subTest(filename=filename):
+                config = os.path.join(self._data_dir, filename)
+                try:
+                    run(["weather-dl", config, "--dry-run"])
+                except Exception:
```
alxmrs, line 33: IIRC, `Exception` is not necessary here.

weather_dl/download_pipeline/config_test.py
```diff
@@ -0,0 +1,38 @@
...
+import unittest
+
+from .pipeline import run
+import weather_dl
+
+
+class ConfigTest(unittest.TestCase):
+
+    def setUp(self):
+        self._data_dir = f'{next(iter(weather_dl.__path__))}/../configs'
```
alxmrs, line 25: Do you mind removing the first `_` from this variable name? For tests, this can be public.

configs/yesterdays_surface_example.cfg
CillianFn: Good point! There is a `KeyError` [here ](https://github.com/google/weather-tools/blob/main/weather_dl/download_pipeline/parsers.py#L223) when `partition_keys` is not set

CillianFn: Created #59 to track

weather_dl/download_pipeline/config_test.py
```diff
@@ -0,0 +1,38 @@
...
+    def setUp(self):
+        self.data_dir = f'{next(iter(weather_dl.__path__))}/../configs'
+
+    def test_process_config_files(self):
+        for filename in os.listdir(self.data_dir):
+            with self.subTest(filename=filename):
+                config = os.path.join(self.data_dir, filename)
+                try:
+                    run(["weather-dl", config, "--dry-run"])
+                except Exception:
```
CillianFn, line 33: @alxmrs you were correct that `Exception` is not required here but its removal caused the linter to fail. I typically wouldn't run a linter on test files, is that a convention you guys have? 🙂

alxmrs: I think it's best to lint all python files, including tests. However, I'm happy if we add a comment to exclude this line from whatever linting rule is violated.

alxmrs (approved):
Looks good to me. Let's merge this after fixing the minor linting issue.

weather_dl/download_pipeline/config_test.py
CillianFn: The linter is passing currently with `except Exception:` and failing on `except:`

Would you prefer, 

a) Leave it as is with `except Exception:`

b) Change to `except:` and exclude this line from linting

alxmrs: b) sounds good to me. Thank you!

PR #66, opened 2022-01-10
Updated the standard example configs to use Reanalyses instead of Ensemble Means.

alxmrs:
This change reflects the more common use cases for the Era 5 dataset.

### DIFF
--- a/configs/era5_example_config_preproc.cfg
+++ b/configs/era5_example_config_preproc.cfg
@@ -20,7 +20,8 @@ partition_keys=
     month
     day
 [selection]
-product_type=reanalysis
+# These are the averages of an ensemble of analyses.
+product_type=ensemble_mean
 format=grib
 grid=o1280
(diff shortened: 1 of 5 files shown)


alxmrs:
@lakshmanok How do these configs look to you?

configs/era5_example_config_preproc.cfg
```diff
@@ -20,7 +20,7 @@ partition_keys=
     month
     day
 [selection]
-product_type=ensemble_mean
+product_type=reanalysis
```
uhager, line 23: Could we leave one as ensemble_mean? Or even make one ensemble_spread or ensemble_members? Just to cover more options in the examples.

alxmrs: I'm ok with that. Will update now.

uhager (approved):

PR #68, opened 2022-01-11
Setting initial versions of each weather-tool.

alxmrs:
Marking each tool as v0.1. This should make it easier for tools to receive updates.

### DIFF
--- a/weather_dl/setup.py
+++ b/weather_dl/setup.py
@@ -35,7 +35,7 @@ setup(
     version='0.1.0',
     author='Anthromets',
     author_email='anthromets-ecmwf@google.com',
-    url='https://github.com/google/weather-tools',
+    url='https://weather-tools.readthedocs.io/en/latest/weather_dl/',
     description='A tool to download weather data.',
     install_requires=base_requirements,
 )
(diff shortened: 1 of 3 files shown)


weather_dl/setup.py
```diff
@@ -32,9 +32,10 @@
 setup(
     name='download_pipeline',
     packages=find_packages(),
+    version='0.1.0',
     author='Anthromets',
     author_email='anthromets-ecmwf@google.com',
-    url='https://gitlab.com/google-pso/ais/grid_intelligence_ai/ecmwf',
+    url='https://github.com/google/weather-tools',
```
uhager, line 38: Here and for the other two: Should we add the specific URLs instead? Maybe to the README? We'd have to update that if we change the current structure, but then if we ever change that, there'll be a lot of changes anyway, and in the mean time maybe it's easier for users if they get to the readme for the tool they are currently trying to use?

alxmrs: I like that idea a lot! For the main project URL, we use a readthedocs link. How about we use the URL for each tool's documentation?

uhager (approved):

PR #72, opened 2022-01-11
`weather-dl`: Fix GCS timeout issues the pipelines intermittently experiences.

alxmrs:
- When copying file objects, we now write the full WRITE_CHUNK_SIZE limit for GCS.
- For the FSStore, we've added an Apache Beam retry with backoff, filtering by server errors and timeouts. In experimental runs, this solution appears to fix the timeout.

### DIFF
--- a/weather_dl/download_pipeline/stores.py
+++ b/weather_dl/download_pipeline/stores.py
@@ -20,6 +20,7 @@ import tempfile
 import typing as t
 
 from apache_beam.io.filesystems import FileSystems
+from apache_beam.utils import retry
 
 
 class Store(abc.ABC):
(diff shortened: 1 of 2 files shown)


weather_dl/download_pipeline/stores.py
```diff
@@ -96,6 +97,8 @@ def exists(self, filename: str) -> bool:
 class FSStore(Store):
     """Store data into any store supported by Apache Beam's FileSystems."""
 
+    @retry.with_exponential_backoff(
```
pramodg, line 100: qq, how many times will it retry ?

pramodg (approved):

weather_dl/download_pipeline/stores.py
alxmrs: The default number of retries is 7.

PR #74, opened 2022-01-11
Improve grib file processing speed

pramodg:
Copy the grib data variables to in memory to avoid disk access on subsequent reads.

### DIFF
--- a/weather_mv/loader_pipeline/pipeline.py
+++ b/weather_mv/loader_pipeline/pipeline.py
@@ -57,6 +57,16 @@ def _prod(xs: t.Iterable[int]) -> int:
     return reduce(operator.mul, xs, 1)
 
 
+def _make_grib_dataset_inmem(grib_ds: xr.Dataset) -> xr.Dataset:
+    # Copies all the vars to in-memory to reduce disk seeks everytime a single row is processed.
+    # This also removes the need to keep the backing temp source file around.
+    data_ds = grib_ds.copy(deep=True)
+    for v in grib_ds.variables:
+        if v not in data_ds.coords:
+            data_ds[v].variable.values = grib_ds[v].variable.values
+    return data_ds
+
+
 def __open_dataset_file(filename: str) -> xr.Dataset:
     try:
         return xr.open_dataset(filename)
@@ -66,7 +76,8 @@ def __open_dataset_file(filename: str) -> xr.Dataset:
             raise
     # Trying with explicit engine for cfgrib.
     try:
-        return xr.open_dataset(filename, engine='cfgrib', backend_kwargs={'indexpath': ''})
+        return _make_grib_dataset_inmem(
+            xr.open_dataset(filename, engine='cfgrib', backend_kwargs={'indexpath': ''}))
     except ValueError as e:
         if "multiple values for key 'edition'" not in str(e):
             raise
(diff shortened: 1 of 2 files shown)


weather_mv/loader_pipeline/pipeline.py
```diff
@@ -57,6 +57,16 @@ def _prod(xs: t.Iterable[int]) -> int:
     return reduce(operator.mul, xs, 1)
 
 
+def _make_grib_dataset_inmem(grib_ds: xr.Dataset) -> xr.Dataset:
```
alxmrs, line 60: I don't think this function is necessary. xarray's `open_dataset` provides the `cache` argument, which can be used to open data in memory: https://xarray.pydata.org/en/stable/generated/xarray.open_dataset.html

weather_mv/loader_pipeline/pipeline_test.py
```diff
@@ -301,6 +315,24 @@ def test_multiple_editions(self):
         }
         self.assertRowsEqual(actual, expected)
 
+    @_handle_missing_grib_be
+    def test_timing_profile(self):
```
alxmrs, line 319: I like the profiling test!

pramodg:
As a simple comparison on a local machine, for a single variable read for 1K coordinates the speed up is ~40 times. (Not rigorously tested, but rough ballpark).

weather_mv/loader_pipeline/pipeline.py
alxmrs: The `chunks` argument may also help here.

pramodg: cache=True is lazy caching, while it does improve read times, its much smaller improvement (only ~2x) on the data i tested. 
Unless we have a memory concern, don't think using that path instead is much more helpful.

alxmrs: I see. I'm glad that you tested this pathway. In that case, I'm happy with the solution you provided. I'll take one more look at the CL to see if I can find any issues.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -66,7 +76,8 @@ def __open_dataset_file(filename: str) -> xr.Dataset:
             raise
     # Trying with explicit engine for cfgrib.
     try:
-        return xr.open_dataset(filename, engine='cfgrib', backend_kwargs={'indexpath': ''})
+        return _make_grib_dataset_inmem(
```
alxmrs, line 79: Can you add a flag so the user can control if the dataset should be opened in memory? I can see this tool being used on O(100GB) files, where users would like to turn this off.

weather_mv/loader_pipeline/pipeline_test.py
```diff
@@ -301,6 +315,23 @@ def test_multiple_editions(self):
...
+        self.test_data_path = f'{self.test_data_folder}/test_data_grib_single_timestep'
+        counter = 0
+        i = extract_rows(self.test_data_path)
+        # Read once to avoid counting dataset open times, etc.
+        _ = next(i)
+        with timing('Loop'):
+            for v in i:
+                # Don't do everything, 10K coordinates is enough for testing.
+                if counter > 10000:
```
alxmrs, line 328: nit: `>=`, since it's zero indexed.

weather_mv/loader_pipeline/pipeline.py
alxmrs: I'm open to creating an issue to address this in the future.

alxmrs (approved):
Added a recommendation for usability. Otherwise, LGTM.

weather_mv/loader_pipeline/pipeline.py
pramodg: Agreed, let's open an issue and get this submitted for now.

weather_mv/loader_pipeline/pipeline_test.py
pramodg: done

PR #88, opened 2022-01-17
Groomed documentation

### LINKED ISSUE #46: `weather-dl`: Document the -n argument
In an oversight, a recent feature of `weather-dl` has not been documented. Let's make known to readers of the tool's README that the flag exists and describe what it does. Further, let's add a usage example.

alxmrs:
- Fixed mobile responsiveness issue with the sphinx template. 
- Fixed #46 
- Updated `weather-mv` docs to reflect need for `--temp_location`. 
- Fixed a link at the top of the README by extracting a documentation section.

### DIFF
--- a/weather_mv/README.md
+++ b/weather_mv/README.md
@@ -93,6 +93,10 @@ streaming ingestion, use the `--topic` flag (see above). Objects that don't matc
 ingestion. It's worth noting: when setting up PubSub, **make sure to create a topic for GCS `OBJECT_FINALIZE` events
 only.**
 
+Data is written into BigQuery using streaming inserts. It may
+take [up to 90 minutes](https://cloud.google.com/bigquery/streaming-data-into-bigquery#dataavailability)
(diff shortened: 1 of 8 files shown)


weather_mv/README.md
```diff
@@ -48,6 +45,7 @@ _Usage Examples_:
 ```bash
 weather-mv --uris "gs://your-bucket/*.nc" \
            --output_table $PROJECT.$DATASET_ID.$TABLE_ID \
+           --temp_location "gs://$BUCKET/tmp" \  # Needed for batch writes to BigQuery
```
uhager, line 48: Is temp_location always required? It should be added above to 'Common options' and marked as '(required)' if applicable.

alxmrs: Thanks! This argument is not always required – only for batch uploads (the BigQuery transform needs it). For the streaming use case, this is not necessary.

pramodg (approved):

Issue #59, opened 2022-01-02
Absence of partition_keys in config file should not cause an exception


CillianFn:
As [discussed](https://github.com/google/weather-tools/pull/56#discussion_r776860848) , it should be possible to **not** specify a partition key and download all data as one chunk

Currently the absence of `partition_keys` in the config file causes a `KeyError`

pranay101:
I am willing to work on this issue. please guild me further

alxmrs:
Sure thing, @pranay101. Sometimes, users may want to download data in one chunk instead of smaller shards. This should be possible today, but isn't. Ideally, users should be able to pass a target template with no string templates, and thus no need to write partition_keys at all. The work for this issue involves update weather-dl's parser (around [these lines](https://github.com/google/weather-tools/blob/375bf7874a6df6d8529253295bab118c2ab6301c/weather_dl/download_pipeline/parsers.py#L309)) such that: 
1. users can create a target path with no templates
2. if (1), then users do not need to specify any partition key argument 
3. update error messages to guide the user to the right config, given these changes
3. update documentation related to these parser changes
4. update and add to unit tests to check this behavior

Further, we need to test that the rest of the pipeline works e2e when there is only one shard worth of work. To my knowledge, this should definitely be the case, but it's always worth while to manually test (this can be done with a local run).

CillianFn:
Fixed with #90

PR #90, opened 2022-01-18
Add check for partition_keys when using append_date_dirs

CillianFn:
Related to #59

### DIFF
--- a/weather_dl/download_pipeline/parsers.py
+++ b/weather_dl/download_pipeline/parsers.py
@@ -220,7 +220,8 @@ def _number_of_replacements(s: t.Text):
 
 
 def use_date_as_directory(config: t.Dict):
-    return 'date' in config['parameters']['partition_keys'] \
+    return 'partition_keys' in config['parameters'] \
+           and 'date' in config['parameters']['partition_keys'] \
            and config['parameters'].get('append_date_dirs', 'false') == 'true' \
            and 'target_filename' in config['parameters']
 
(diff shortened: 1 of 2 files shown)


weather_dl/download_pipeline/parsers.py
```diff
@@ -220,7 +220,8 @@ def _number_of_replacements(s: t.Text):
 
 
 def use_date_as_directory(config: t.Dict):
-    return 'date' in config['parameters']['partition_keys'] \
+    return 'partition_keys' in config['parameters'] \
+           and 'date' in config['parameters']['partition_keys'] \
```
alxmrs, line 224: Nice catch!

(It's hard to see in this diff): Is there a test for a successful config with no partition keys? If not, can you add one?

weather_dl/download_pipeline/parsers_test.py
```diff
@@ -565,6 +565,34 @@ def test_accepts_partition_keys_matching_sections(self):
...
+                    1990
+                    2000
+                    2010
+                    2020
+                """
+        ) as f:
+            config = process_config(f)
+            params = config.get('parameters', {})
+            self.assertTrue(bool(config))
+            self.assertIsInstance(params['partition_keys'], list)
```
CillianFn, line 594: @alxmrs Checking that `partition_keys` is an instance of type list even when not specified in the config file seems to be intended logic [here](https://github.com/google/weather-tools/blob/main/weather_dl/download_pipeline/parsers.py#L309) & [here](https://github.com/google/weather-tools/blob/main/weather_dl/download_pipeline/parsers.py#L339-L340)

LMK if this still holds up

alxmrs: I appreciate including this in the tests. However, this is more of an implementation detail and not part of the end interface – so no need to exercise it.

alxmrs (approved):

CillianFn:
@alxmrs I accidentally deleted the branch, my bad.

alxmrs:
Are you able to merge the change?

Issue #84, opened 2022-01-14
Unit tests are running slow / CI takes a long time. 


alxmrs:
I observe that our CI takes a long time to complete. It is early in the lifetime of this project for our unit tests to be slow (certainly, the codebase is small enough). Having fast tests and an expedient CI system increases the pace of development & bugfixes. 

This bug involves investigating why our unit tests takes a long time to run. This may be related to #53.

CillianFn:
@alxmrs This is caused by these [two tests](https://github.com/google/weather-tools/blob/main/weather_dl/download_pipeline/stores_test.py#L47-L59) and the exponential backoff [here](https://github.com/google/weather-tools/blob/main/weather_dl/download_pipeline/stores.py#L100-L101)

![image](https://user-images.githubusercontent.com/18357650/150012460-6d08d8b4-1121-4551-b283-c55bf891c72c.png)

I'm thinking of overriding some of the [defaults ](https://beam.apache.org/releases/pydoc/2.4.0/apache_beam.utils.retry.html#apache_beam.utils.retry.with_exponential_backoff) for the tests only, sound good?

alxmrs:
Great investigation! It makes sense that retries are causing the delay. 

I looked at Apache Beam's [filter definition](https://beam.apache.org/releases/pydoc/2.4.0/_modules/apache_beam/utils/retry.html#retry_on_server_errors_and_timeout_filter), and I don't understand why a value error would cause retries. I wonder if the underlying GCS File IO modules are wrapping the exceptions. Does your fix from the edit save time in practice?

CillianFn:
@alxmrs Seems to be back in business :turtle: :rabbit2:
https://github.com/google/weather-tools/runs/4860959961?check_suite_focus=true

alxmrs:
Thanks Cillian! Marking as closed.

PR #92, opened 2022-01-18
Fix slow test runs 

CillianFn:
More information over on #84 

Unit test run down to ~20 seconds here from ~15-18 mins


alxmrs (approved):
LGTM! Great fix

Issue #70, opened 2022-01-11
Exclude test data from tarball


alxmrs:
Looking at the upload-to-pypi logs, it looks like testdata is included in the tarball (I see .grib and .nc files in the logs). This bug involves updating the MANIFEST.in to exclude testdata in the tars. That way, they will be lightweight / quick-to-install.

alxmrs:
Fixed with #93, closing...

PR #93, opened 2022-01-19
Exclude test data from tarball

CillianFn:
#70 
Verified this works locally using `python setup.py sdist` in both `weather_mv` and `weather_sp`


alxmrs (approved):
LGTM! Thanks for the fix.

CillianFn:
@alxmrs np 😄 you asked on another PR, I don't have merge rights so I can't merge this.

PR #97, opened 2022-01-25
Updating tool versions in prep for release.

alxmrs:
Documented version update and release process.

Updating version of `weather_sp` to reflect new functionality. `weather_dl` updated version to reflect fixes. `weather_mv` has already been incremented for this next release.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -174,7 +174,7 @@ python -m http.server -d docs/_build/html
 ## Versions & Releasing
 
 We aim to represent the version of each tool using [semver.org](https://semver.org/) semantic versions. To that end,
-let's abide by the following pattern:
+we will abide by the following pattern:
 
 - When making a change to a particular tool, please remember to update its semantic version in the `setup.py` file.
 - The version for _all the tools_ should be incremented on update to _any_ tool. If one tool changes, the
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,10 +85,12 @@ setup.py  # Project is pip-installable, project requirements managed here.
 
 ### Windows Developer Instructions
 
-Windows support for each CLI is currently under development (See [#64](https://github.com/google/weather-tools/issues/64)).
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -168,3 +170,17 @@ After the docs are re-generated, you can view them by starting a local file serv
 ```shell
 python -m http.server -d docs/_build/html
 ```
+
+## Versions & Releasing
+
+We aim to represent the version of each tool using [semver.org](https://semver.org/) semantic versions. To that end,
+let's abide by the following pattern:
```
pramodg, line 177: s/let's/we will/ ?

pramodg (approved):
Are there also release notes to capture high level changes per release ?
(Is that a separate CL)

weather_sp/setup.py
```diff
@@ -26,7 +26,7 @@
     packages=find_packages(),
     author='Anthromets',
     author_email='anthromets-ecmwf@google.com',
-    version='0.1.0',
+    version='0.2.0',
```
uhager, line 29: What determines that this is 0.2.0 while the other one is now 0.1.1? From the Versions & Release text it sounds like all tools should have the same version. Can you clarify this a bit more in the explanation? Is it ok if major / minor versions of the tools drift apart?

alxmrs:
> Are there also release notes to capture high level changes per release ?

Instead of keeping the release notes in a CHANGELOG.md file, I think it's simpler to make use of Githubs release system. So, the current description of changes are located here: https://github.com/google/weather-tools/releases

weather_sp/setup.py
alxmrs: Yeah, that's a great question. I'm following the [semver.org](http://semver.org)'s rules of thumb. I'm updating the minor version since `weather-sp` gained a backwards compatible new feature, namely the output path structuring things you recently landed. If that were considered a bugfix, then I would increment the version to `0.1.1`. 

I could probably provide an example in the release process notes – Each specific tool will have it's own version, and we'll increment the versions of the whole `weather-tools` package if any subtool is updated.

PR #101, opened 2022-02-01
Refactored weather-mv to work with pluggable Data Sinks.

alxmrs:
- Created abstraction for Data Sinks (where weather data can be moved to)
- Pulled out existing pipeline into a BigQuery sink
- Added more tests (for example, of the CLI).
- Added `dataclasses` as a project dependency for older python versions
- Updated class documentation


weather_mv/loader_pipeline/bq.py
```diff
@@ -0,0 +1,299 @@
+# Copyright 2022 Google LLC
```
uhager, line 1: 'bq' as a file name seems terse bordering on cryptic. maybe at least bq_sink, or to_big_query? (Apologies if this is still naming-in-progress and you are still working on the structure, I just saw the emails and got curious)

alxmrs: I like the name `bq`, so long as it's obvious that I'm referring to BigQuery. I don't think `_sink` is necessary. According to the Google Python Style Guide, abbreviations are OK so long as they're understood beyond the scope of the project [0]. It seems like "bq" is pretty conventional to mean BigQuery, see [its usage in Pandas](https://pandas.pydata.org/docs/reference/api/pandas.read_gbq.html), for example. 

[0]: https://google.github.io/styleguide/pyguide.html#316-naming

weather_mv/loader_pipeline/bq.py
```diff
@@ -0,0 +1,299 @@
...
+
+    # Add an extra columns for recording import metadata.
+    fields.append(bigquery.SchemaField(DATA_IMPORT_TIME_COLUMN, 'TIMESTAMP', mode='NULLABLE'))
+    fields.append(bigquery.SchemaField(DATA_URI_COLUMN, 'STRING', mode='NULLABLE'))
+    fields.append(bigquery.SchemaField(DATA_FIRST_STEP, 'TIMESTAMP', mode='NULLABLE'))
+
+    return fields
+
+
+def to_json_serializable_type(value: t.Any) -> t.Any:
```
pramodg, line 152: Maybe useful to move functions like these to a util file

weather_mv/loader_pipeline/bq.py
```diff
@@ -0,0 +1,299 @@
...
+    logger.info(f'target-only dataset size: {dropped_ds.nbytes}')
+
+    return dropped_ds
+
+
+def _prod(xs: t.Iterable[int]) -> int:
+    return reduce(operator.mul, xs, 1)
+
+
+def get_coordinates(ds: xr.Dataset, uri: str = '') -> t.Iterator[t.Dict]:
```
pramodg, line 215: this seems more like a xarray util as compared to bq func

pramodg (approved):
In general, I really the approach of splitting the single file up :)

PR #105, opened 2022-02-08
Update weather-sp's templating system to allow users to specify level and shortname.

### LINKED ISSUE #102: Splitter template syntax is not expressive enough for backwards compatability.
I used the weather splitter on a file system before it had templates, and now in its current version. The original file system I have looks like: 
`.../2020/202001_hres_pcp.grb2_surface_d2m.grib`

Using the new templating system in the weather splitter, I cannot format a template string to match this file path. This is because I don't have control over the position of the `level` and `short_name` variables. It seems like the templating system is oriented towards setting variable file paths instead of templating the variable and level placement in the string. The latter templating options are more needed than the former.

alxmrs:
Fixes #102. In addition, file endings are no longer needed -- now, we attempt to read the file to determine which splitter to use.

### DIFF
--- a/weather_sp/splitter_pipeline/file_name_utils.py
+++ b/weather_sp/splitter_pipeline/file_name_utils.py
@@ -53,7 +53,6 @@ def get_output_file_base_name(filename: str,
     """
     split_name, ending = os.path.splitext(filename)
     if ending in GRIB_FILE_ENDINGS or ending in NETCDF_FILE_ENDINGS:
-        ending = ending
         filename = split_name
     else:
         ending = ''
--- a/weather_sp/splitter_pipeline/file_name_utils.py
+++ b/weather_sp/splitter_pipeline/file_name_utils.py
@@ -16,9 +16,6 @@ import logging
 import os
 import typing as t
 
-GRIB_FILE_ENDINGS = ('.grib', '.grb', '.grb2', '.grib2', '.gb')
-NETCDF_FILE_ENDINGS = ('.nc', '.cd')
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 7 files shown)


weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -51,21 +48,21 @@ def get_output_file_base_name(filename: str,
             The output file is then created by replacing this part of the input name
             with the output pattern.
     """
-    file_ending = ''
-    split_name, ending = os.path.splitext(filename)
-    if ending in [*GRIB_FILE_ENDINGS, *NETCDF_FILE_ENDINGS]:
-        file_ending = ending
-        filename = split_name
+    filename, ending = os.path.splitext(filename)
```
uhager, line 51: I'm pretty sure you suggested this in a previous change and I already argued against it :)
Just checking some example files I have for testing, there's a file '20210101_gfs.t00z.pgrb2.0p25.f012'. I'm fairly certain our gfs gribs follow this king of patter ending with '.pgrb2.0p25.f0xy' depending on the step for the forecast time. 'f012' should no be used as ending.

alxmrs: Ah, now these lines make sense, thanks for the context, I'll add them back.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -51,21 +51,26 @@ def get_output_file_base_name(filename: str,
             The output file is then created by replacing this part of the input name
             with the output pattern.
     """
-    file_ending = ''
     split_name, ending = os.path.splitext(filename)
-    if ending in [*GRIB_FILE_ENDINGS, *NETCDF_FILE_ENDINGS]:
-        file_ending = ending
+    if ending in GRIB_FILE_ENDINGS or ending in NETCDF_FILE_ENDINGS:
+        ending = ending
```
uhager, line 56: Is this line needed?

alxmrs: Thanks! 🤦🏽

uhager (approved):

PR #108, opened 2022-02-10
Added skipping logic to weather-sp

alxmrs:
* Added a way to skip splitting files that have already been split.
* Use can turn off skipping with a `-f` or force flag.

### DIFF
--- a/weather_sp/splitter_pipeline/file_splitters.py
+++ b/weather_sp/splitter_pipeline/file_splitters.py
@@ -94,7 +94,7 @@ class GribSplitter(FileSplitter):
         outputs = dict()
 
         if self.should_skip():
-            metrics.Metrics.counter('file_splitters', f'skipped').inc()
+            metrics.Metrics.counter('file_splitters', 'skipped').inc()
             self.logger.info('Skipping %s, file already split.', repr(self.input_path))
             return
 
@@ -125,7 +125,7 @@ class NetCdfSplitter(FileSplitter):
 
     def split_data(self) -> None:
         if self.should_skip():
-            metrics.Metrics.counter('file_splitters', f'skipped').inc()
+            metrics.Metrics.counter('file_splitters', 'skipped').inc()
             self.logger.info('Skipping %s, file already split.', repr(self.input_path))
             return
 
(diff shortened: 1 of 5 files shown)


weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -73,12 +74,30 @@ def _get_output_file_path(self, key: SplitKey) -> str:
...
         return self.output_info.file_name_template.format(**split_keys)
 
+    def should_skip(self):
+        """Skip splitting if the data was already split."""
+        if self.force_split:
+            return False
+
+        for match in FileSystems().match([
+            self._get_output_file_path(SplitKey('', '**')),
+            self._get_output_file_path(SplitKey('**', '**')),
```
uhager, line 84: I'm unsure about this. In many cases, this is probably what is wanted, but what if we split some large files, find missing variables, download those variables in separate files, then split those new files? In a way, that use case is part of what this splitter is for. Of course, in that case we can set force_split to true, but I think having the default be false can make the behaviour unexpected. I think it would be better to have a skip_if_similar_exists or something, but by default, resplit everything.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -73,12 +74,30 @@ def _get_output_file_path(self, key: SplitKey) -> str:
             split_keys['levelType'] = f'{key.levelType}_'
         return self.output_info.file_name_template.format(**split_keys)
 
+    def should_skip(self):
```
uhager, line 77: Tests for this?

weather_sp/splitter_pipeline/file_splitters.py
alxmrs: It's a bit tough to say which way the default should be. I'm glad we're in agreement that either way, the user will be able to configure the tool to do what they want. 

I'm leaning towards defaulting to skip splitting, using a flag to override. I think this will be more consistent with the other weather tools. From my experience running the tool, I'd prefer to skip work already done by default.

weather_sp/splitter_pipeline/file_splitters.py
alxmrs: Thanks, will write right now.

uhager (approved):

PR #110, opened 2022-02-12
Upload to cloud is robust to socket timeout errors.

### LINKED ISSUE #109: line 1099, in read return self._sslobj.read(len, buffer) socket.timeout: The read operation timed out
After running weather-dl for a long time (~25 days), I hit this error:
```
line 1099, in read return self._sslobj.read(len, buffer) socket.timeout: The read operation timed out
```

My suspicion is that reading or writing large files to GCS is timing out. While we have retries for this, we filter out certain errors for retrying, including this one. A fix for this might look like revisiting that retry logic and making sure to include these type of timeout errors.

alxmrs:
Fixes #109.

- Created a Beam retry filter to account for socket timeouts.
- Applied filter to both the file open and file upload steps.

### DIFF
--- a/weather_dl/download_pipeline/util.py
+++ b/weather_dl/download_pipeline/util.py
@@ -1,0 +1,44 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


weather_dl/download_pipeline/util.py
```diff
@@ -0,0 +1,44 @@
...
+        return True
+    return retry.retry_if_valid_input_but_server_error_and_timeout_filter(exception)
+
+
+class _FakeClock:
+    def sleep(self, value):
+        pass
+
+
+def retry_with_exponential_backoff(fun):
```
pramodg, line 33: Would it be useful to make it more generic using @functools.wraps etc.

pramodg (approved):

alxmrs:
Thanks for the review, Pramod!

PR #111, opened 2022-02-24
Fix wrong output file example in weather-sp readme.

uhager:


alxmrs (approved):
LGTM! Thanks for the fix.

PR #112, opened 2022-02-28
Added basic tests

pranay101:


pranay101:
Hey, @alxmrs just wanted to about the wirings of the test files a bit. I have had a hard time figuring out where these test cases are being run. before I commit these changes I manually tested the function and it worked for all the cases.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -42,4 +43,17 @@ def test_no_duplicate_coordinates(self):
 
 class ToJsonSerializableTypeTests(unittest.TestCase):
     # TODO(#106): Write tests...
```
alxmrs, line 45: Please remove.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -42,4 +43,17 @@ def test_no_duplicate_coordinates(self):
...
+    def assert_serializes_json(self, input):
+        try:
+            json.loads(json.dumps(input))
+            print("Successfully passed!")
+        except json.JSONDecodeError:
+            self.fail()
+
+    def test_numpy_float_becomes_float(self):
+        res = to_json_serializable_type(np.float(42))
+        self.assert_serializes_json(res)
```
alxmrs, line 55: This assert is not sufficient. Please test that `res` is of type `float` and equals `42`.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -42,4 +43,17 @@ def test_no_duplicate_coordinates(self):
...
+
+    def test_numpy_array(self):
+        res = to_json_serializable_type(np.array([1, 2, 3, 4, 5777, 7, 4252, 5426]))
```
alxmrs, line 58: Please use an `assertEquals` to set assert against the expected result, in addition to this assert.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -42,4 +43,17 @@ def test_no_duplicate_coordinates(self):
 
 class ToJsonSerializableTypeTests(unittest.TestCase):
     # TODO(#106): Write tests...
-    pass
+    def assert_serializes_json(self, input):
```
alxmrs, line 46: Please add a test case for every expected input type (check out the if statements in the method you're testing to infer which types to test). In general, every conditional block in the method should be exercised.

alxmrs:
> I have had a hard time figuring out where these test cases are being run

As per our developer documentation, you can use `pytest` to run this unit test. If you run `pytest weather_mv`, for example, it will run all the unit tests in `weather_mv`. Check out these pytest docs to learn more about how to run specific tests on the command line: https://docs.pytest.org/en/6.2.x/contents.html

Out of curiosity, which IDE are you using to develop?

pranay101:
> > I have had a hard time figuring out where these test cases are being run
> 
> As per our developer documentation, you can use `pytest` to run this unit test. If you run `pytest weather_mv`, for example, it will run all the unit tests in `weather_mv`. Check out these pytest docs to learn more about how to run specific tests on the command line: https://docs.pytest.org/en/6.2.x/contents.html
> 
> Out of curiosity, which IDE are you using to develop?

yeah @alxmrs, I am using vs code currently.

weather_mv/loader_pipeline/util_test.py
pranay101: yes, was a bit confused about the wiring so didn't make it through...looking forward to test it for all possible cases now

pranay101:
`ToJsonSerializableTypeTests::test_date_and_time Failed: [undefined]AssertionError: '2022-03-10T20:41:23.776660' != '2022-03-10T20:41:23.776660+00:00'
- 2022-03-10T20:41:23.776660
+ 2022-03-10T20:41:23.776660+00:00
?                                                  ++++++
self = <weather_mv.loader_pipeline.util_test.ToJsonSerializableTypeTests testMethod=test_date_and_time>`


hey, @alxmrs. I have been stuck with this error when checking for the DateTime test. the output from the function to_json_serializable_type returns date with time attached in the end(+00:00). how should I check it with the original? and sorry for the commit I think I messed up with some imports so it didn't pass the test. looking for the error.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -41,5 +42,68 @@ def test_no_duplicate_coordinates(self):
...
 class ToJsonSerializableTypeTests(unittest.TestCase):
-    # TODO(#106): Write tests...
-    pass
+
+    def assert_value_and_type(self,input,output):
+        # check if it retains the original value
+        self.assertEqual(input,output)
+
+        # check if it retains the original type
+        self.assertEqual(type(input), type(output))
```
alxmrs, line 51: This doesn't look right to me. I don't think this kind of assert can be reused for all the tests, because serialization often changes the type of the output. For example, `datetime.datetime`s should change to type `str`.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -41,5 +42,68 @@ def test_no_duplicate_coordinates(self):
...
+
+        # check if it retains the original value and type
+        self.assert_value_and_type(input,output)
+       
+
+    def test_numpy_array(self):
+        input = np.array([1, 2, 3, 4, 5777, 7, 4252, 5426])
+        output = to_json_serializable_type(input)
+        self.assert_serializes_json(output)
+        self.assert_value_and_type(input,output)
```
alxmrs, line 79: I disagree with this assert: the new type should be a list of ints, IIRC.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -41,5 +42,68 @@ def test_no_duplicate_coordinates(self):
...
+
+    def test_strings(self):
+        input = "hello there"
+        output = to_json_serializable_type(input)
+        self.assert_serializes_json(output)
+        self.assert_value_and_type(input,output)
+
+    def test_date_and_time(self):
+        input = datetime.datetime.now().isoformat()
```
alxmrs, line 88: It would be a more powerful test if you also exercised serializing a `datetime.datetime`, `np.datetime64`, etc. in addition to an ISO string. As mentioned above, you will likely not be able to use the `assert_value_and_type` method.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -41,5 +42,68 @@ def test_no_duplicate_coordinates(self):
...
+        self.assert_serializes_json(output)
+        self.assert_value_and_type(input,output)
+
+    def test_dictionary(self):
+        input = {"a": 1, "b": 2, "c": 3, "d": 4}
+        output = to_json_serializable_type(input)
+        self.assert_serializes_json(output)
+        self.assert_value_and_type(input,output)
+
+    def test_None(self):
```
alxmrs, line 105: please, use snake case (all lowercase).

weather_mv/loader_pipeline/util_test.py
```diff
@@ -41,5 +42,68 @@ def test_no_duplicate_coordinates(self):
...
+        input = None
```
alxmrs, line 106: Consider exercising other values that will become None, for example, `np.NaN` or float('NaN').

alxmrs:
> returns date with time attached in the end(+00:00). 

Yes, the implementation of the function adds timezone information to the datetime before producing the ISO string. I think there's a better way you could go about writing these tests.

pranay101:
hey, @alxmrs. sorry for the lack of commits. I am currently working for an org from google summer of codes. need a few days until I submit my proposal there. and then I can start working on this issue.

alxmrs:
No problem. Thanks for keeping me posted.

pranay101:
Hey, @alxmrs. Created a few more commits. I wanted to how should I check for input None. I have had a hard time brainstorming it.

pranay101:
Hey @alxmrs are you on it?

alxmrs:
> I wanted to know how should I check for input None.
Sorry for the late reply. I recommend trying values that are Not-a-Number (float('Nan').

alxmrs:
Marking closed, as the changes have gotten a bit stale.

PR #115, opened 2022-03-03
New partitioning strategy for `weather-dl`

### LINKED ISSUE #98: `weather-dl`: Out-of-order license distribution error
In the weather downloader, we underutilize licenses due to the current way they are distributed. Worse, if we use near the max number of queued workers (like we do by default for the MARS downloads), pipelines will fail with a "MaxAPIRequests" error from ECMWF. 

My leading hypothesis about what's causing this is that our license distribution scheme depends on a particular order of future requests. However, we've made recent changes to better work with the dataflow autoscaling algorithm to parallelize download requests, which allocates the `fetch_data` step in a random order. 

## How to reproduce
1. Run `weather-dl` with multiple licenses (CDS is fine).
2. Look up the request queue for the client (the docstrings for the client have an appropriate URL) for each license / account.

*Outcome*: There will be an uneven distribution of download requests. Some may have duplicates while others don't have any work dispatched. This leads to underutilization of licenses and in some cases brittle pipelines. 
*Expected*: Each worker should at least have their request limit maxed out (licenses fully utilized). Requests, in general, should be evenly distributed. Pipelines should never hit the "Too many requests" error. 

## Workarounds
The strategy to get a robust yet high utilization of downloads is to split data requests into multiple configs and run `weather-dl` twice (or multiple times). It especially helps to leverage the `-n` flag to limit the number of data requests to the access limit (for CDS, the defaults are currently fine. For MARS, set `-n 2`). 

## Potential fix
A good strategy would be to restructure the pipeline such that: 
* License creation is separate from config partitioning (i.e. different `beam.Create()` steps). 
* The two streams (licenses and partitions) get joined together (like a `zip` in Python) after a reshuffle so licenses are evenly distributed.

alxmrs:
In this new scheme, we make use of `beam.Partition` to split work by available license (via param subsections). This simplifies a lot of the pipeline logic.

E2E tests will validate if this, indeed, fixes #98.

### DIFF
--- a/weather_dl/download_pipeline/pipeline.py
+++ b/weather_dl/download_pipeline/pipeline.py
@@ -140,7 +140,14 @@ def _create_partition_config(option: t.Tuple, config: t.Dict, index: int = -1) -
 
 
 def partition_by_index(config: t.Dict, num_partitions: int) -> int:
-    """Split config partitions into groups based on their `__index__` value."""
+    """Split config partitions into groups based on their `__index__` value.
+
+    Args:
+       config: A sharded download config with an `__index__` int in the `parameters` section.
+       num_partitions: must be 1 or greater.
+    Returns:
+        The partition index.
+    """
     index = config.get('parameters', {}).get('__index__', -1)
     if index == -1:
         return index
--- a/weather_dl/download_pipeline/pipeline.py
+++ b/weather_dl/download_pipeline/pipeline.py
@@ -57,7 +57,7 @@ def configure_workers(client_name: str,
     """Configure the number of workers and threads for the pipeline, allowing for user control."""
 
     # The number of workers should always be proportional to the number of licenses.
-    num_api_keys = config.get('parameters', {}).get('num_api_keys', 1)
+    num_api_keys = len(get_subsections(config))
 
     # If user doesn't specify a number of requestors, make educated guess based on clients and dataset.
     if num_requesters_per_key == -1:
--- a/weather_dl/download_pipeline/pipeline.py
+++ b/weather_dl/download_pipeline/pipeline.py
@@ -107,7 +107,7 @@ def prepare_target_name(config: t.Dict) -> str:
     return target
 
 
-def _create_partition_config(option: t.Tuple, config: t.Dict) -> t.Dict:
+def _create_partition_config(option: t.Tuple, config: t.Dict, index: int = -1) -> t.Dict:
     """Create a config for a single partition option.
 
     Output a config dictionary, overriding the range of values for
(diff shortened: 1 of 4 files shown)


alxmrs:
The new download Dataflow graph looks like this:

dl-partition

alxmrs:
In preliminary e2e tests, this seems to fix #98 (will take longer, but it's possible to verify this via integration tests). However, it's more noticeable now that we get poor utilization due the to autoscaling algorithm. I have an idea for an approach that should produce higher utilization on even small numbers of requests (as well as compute) via [Splittable Do functions](https://beam.apache.org/documentation/programming-guide/#splittable-dofns). Given all this the next steps for work on the downloader seem to be: 
- Land this change, after a few more iterations (I think it could use some more tests). 
- Write a design doc (maybe, on Github issues?) to refactor the downloader to make use of Splittable Do functions, which improve performance when dealing with IO.

alxmrs:
> However, it's more noticeable now that we get poor utilization due the to autoscaling algorithm. 

The specific ill here is that the autoscaling algorithm will start with 1 worker and stay there for a while. Only after making one full download, will it start to pick up the pace, and make fetches up to the worker limit.

weather_dl/download_pipeline/pipeline.py
```diff
@@ -133,9 +133,20 @@ def _create_partition_config(option: t.Tuple, config: t.Dict) -> t.Dict:
...
+
     return out
 
 
+def partition_by_index(config: t.Dict, num_partitions: int) -> int:
+    """Split config partitions into groups based on their `__index__` value."""
+    index = config.get('parameters', {}).get('__index__', -1)
+    if index == -1:
+        return index
+    return index % num_partitions
```
uhager, line 147: Should there be a check that index is >0? Currently it's created using enumerate, so it has to be, but just in case.

alxmrs: So, as long as `num_partitions` is positive, the resulting mod operation will also be positive: 
> The modulo operator always yields a result with the same sign as its second operand (or zero)
https://docs.python.org/3/reference/expressions.html#binary-arithmetic-operations

However, it's possible that `num_partitions` receives a negative number. How about I update the documentation of the method to indicate that the number of partitions should be at least 1?

uhager (approved):

weather_dl/download_pipeline/pipeline.py
```diff
@@ -150,7 +168,32 @@ def skip_partition(config: t.Dict, store: Store) -> bool:
...
+
+    If the `parameters` section contains subsections (e.g. '[parameters.1]',
+    '[parameters.2]'), collect a list of the subsection key-value
+    pairs. Otherwise, return a list of an empty dictionary (i.e. to address
+    the case where there are no subsections).
+
+    This is useful for specifying multiple API keys for your configuration.
+    For example:
+    ```
+      [parameters.deepmind]
```
pramodg, line 182: Maybe use more generic names than team names ?

alxmrs: I've only copied the old docstring message from one place to another here. Further, I like these names. I think they add  color to how users can think about licenses and who they belong to, using course examples from Alphabet.

pramodg (approved):
Approving this for now. 
It may be useful to consider an alternate strategy in the long term to increase throughput.

alxmrs:
Closing this in favor of #116. While this is logically correct (for lack of a better term), it doesn't work well with the autoscaling algorithm in Dataflow. Thus, in effect, this only fetches on shard at a time. #116 gets better results, since it takes a data-oriented approach (instead of trying to fight the autoscaling algorithm).

PR #117, opened 2022-03-18
Downloader refactor: extracted out partitioning; tested pipeline args.

alxmrs:
This change improves the modularity and testability of the downloader pipeline. Here, we extract out the partition preparation steps into its own composite PTransform. This provides a better location for the partitioning logic, and makes testing how we partition much better.

Next, we've decoupled running the pipeline from processing user arguments, and added unit tests to the user processing part. This ensures we can exercise logic about how to configure the pipeline without actually running it.

### DIFF
--- a/weather_dl/download_pipeline/partition_test.py
+++ b/weather_dl/download_pipeline/partition_test.py
@@ -1,0 +1,262 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import itertools
+import json
+import tempfile
+import typing as t
+import unittest
+from unittest.mock import MagicMock
+
+import apache_beam as beam
+from xarray_beam._src.test_util import EagerPipeline
+
+from .manifest import MockManifest, Location, LocalManifest
+from .parsers import get_subsections
+from .partition import skip_partition, PartitionConfig
+from .stores import InMemoryStore, Store
+
+
+class OddFilesDoNotExistStore(InMemoryStore):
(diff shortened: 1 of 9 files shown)


weather_dl/download_pipeline/partition_test.py
```diff
@@ -0,0 +1,262 @@
...
+                'features': ['pressure', 'temperature', 'wind_speed_U', 'wind_speed_V'],
+                'month': [str(i) for i in range(1, 13)],
+                'year': [str(i) for i in range(2015, 2021)]
+            }
+        }
+
+        actual = self.create_partition_configs(config)
+
+        self.assertListEqual([d['selection'] for d in actual], [
+            {**config['selection'], **{'year': [str(i)]}}
```
fredzyda, line 70: I feel with statements like this, it might soon be that it is time to make the config a real class instead of a dict of dicts of lists of dicts or whatever the actual type is now.

weather_dl/download_pipeline/partition_test.py
```diff
@@ -0,0 +1,262 @@
...
+                'features': ['pressure', 'temperature', 'wind_speed_U', 'wind_speed_V'],
+                'month': [str(i) for i in range(1, 3)],
+                'year': [str(i) for i in range(2015, 2017)]
+            }
+        }
+
+        actual = self.create_partition_configs(config)
+
+        self.assertListEqual([d['selection'] for d in actual], [
+            {**config['selection'], **{'year': ['2015'], 'month': ['1']}},
```
fredzyda, line 90: I know this is a more compact representation of the expected value, but I prefer the more explicit definition below.

fredzyda (approved):
This does look a lot easier to test and simplifies things in a nice way.

I really do think you should make the configuration a real class at some point.
That would let you more easily validate what was in there and prevent typo
bugs that seem really easy to create now.

alxmrs:
> make the configuration a real class at some point.

I totally agree that we should do this. I'll file an issue to track that work, and we can tackle it later. It seems related to another goal we have for the parser (to parse types).

weather_dl/download_pipeline/partition_test.py
alxmrs: Created an issue to track: #118.

weather_dl/download_pipeline/partition_test.py
alxmrs: I think after we convert these to dataclasses, the representation will be much cleaner.

PR #125, opened 2022-03-26
Flexible splits

uhager:
Allow more flexible splitting.

Dimensions to split along are specified using output formatting to avoid mismatch between formatting and splitting.
Possible splits for GRIB files are all metadata fields, like shortName, typeOfLevel, date, time, step, forecastTime, etc.
Possible splits for NetCDF files are 'level', 'time', 'variable'. If the specified dimension (time, level) is not in the file, and exception will the thrown.

Output formatting can be given using an output template that contains formatting marks, or an output directory with an additional formatting string. If an output directory is specified without formatting, file type-specific defaults are used.

### DIFF
--- a/weather_sp/README.md
+++ b/weather_sp/README.md
@@ -78,7 +78,7 @@ On the other hand, to specify a specific pattern use
 --input-pattern 'gs://test-tmp/era5/2017/*/*.nc'
 ```
 
-## Output & Dimensions to split
+## Output & Split Dimensions
 
 The base output file names are specified using the `--output-template` or `--output-dir` flags. These flags are mutually
 exclusive, and one of them is required. \
--- a/weather_sp/README.md
+++ b/weather_sp/README.md
@@ -11,9 +11,24 @@ Splits NetCDF and Grib files into several files by variable (_alpha_).
 ## Usage
 
 ```
-usage: weather-sp [-h] -i INPUT_PATTERN -output-dir OUTPUT_DIR [-d]
+usage: weather-sp [-h] -i INPUT_PATTERN (--output-template OUTPUT_TEMPLATE | --output-dir OUTPUT_DIR) [--formatting FORMATTING] [-d] [-f]
 
-Split weather data file into files by specified dimensions.
+Split weather data file into files by variable.
+
+optional arguments:
+  -h, --help            show this help message and exit
+  -i INPUT_PATTERN, --input-pattern INPUT_PATTERN
+                        Pattern for input weather data.
+  --output-template OUTPUT_TEMPLATE
+                        Template specifying path to output files using python-style formatting substitution of input directory names. For `input_pattern a/b/c/**` and file
+                        `a/b/c/file.grib`, a template with formatting `/somewhere/{1}-{0}.{level}_{shortName}.grib` will give `somewhere/c-file.level_shortName.grib`
+  --output-dir OUTPUT_DIR
+                        Output directory that will replace the common path of the input_pattern. For `input_pattern a/b/c/**` and file `a/b/c/file.nc`, `outputdir /x/y/z` will create
+                        output files like `/x/y/z/c/file_variable.nc`
+  --formatting FORMATTING
+                        Used in combination with `output-dir`: specifies the how to split the data and format the output file. Example: `_{time}_{level}hPa`
+  -d, --dry-run         Test the input file matching and the output file scheme without splitting.
+  -f, --force           Force re-splitting of the pipeline. Turns of skipping of already split data.
 ```
 
 _Common options_:
@@ -89,13 +104,14 @@ For example, adding `{time}` in the output template or formatting will cause the
 How a file can be split depends on the file type.
 
 #### GRIB
-GRIB files can be split along any dimension that is available in the file's metadata. \
-Examples: 'typeOfLevel', 'level', 'step', 'shortName', 'gridType', 'time', 'forecastTime'
+GRIB files can be split along any dimensions that is available in the file's metadata. \
+Examples: 'typeOfLevel', 'level', 'step', 'shortName', 'gridType', 'time', 'forecastTime' \
+Any available dimensions can be combined when splitting.
 
 #### NetCDF
 NetCDF files are already in a hypercube format and can only be split by one of the dimensions and by data variable.
 Since splitting by latitude or longitude would lead to a large number of small files, this is not supported,
-and it is recommended to use the weather-mv tool instead. \
+and it is recommended to use the `weather-mv` tool instead. \
 Supported splits for NetCDF files are thus 'variable', 'time', 'level'.
 
 
--- a/weather_sp/README.md
+++ b/weather_sp/README.md
@@ -12,23 +12,6 @@ Splits NetCDF and Grib files into several files by variable (_alpha_).
 
 ```
 usage: weather-sp [-h] -i INPUT_PATTERN (--output-template OUTPUT_TEMPLATE | --output-dir OUTPUT_DIR) [--formatting FORMATTING] [-d] [-f]
-
-Split weather data file into files by variable.
-
-optional arguments:
-  -h, --help            show this help message and exit
-  -i INPUT_PATTERN, --input-pattern INPUT_PATTERN
-                        Pattern for input weather data.
-  --output-template OUTPUT_TEMPLATE
-                        Template specifying path to output files using python-style formatting substitution of input directory names. For `input_pattern a/b/c/**` and file
-                        `a/b/c/file.grib`, a template with formatting `/somewhere/{1}-{0}.{level}_{shortName}.grib` will give `somewhere/c-file.level_shortName.grib`
-  --output-dir OUTPUT_DIR
-                        Output directory that will replace the common path of the input_pattern. For `input_pattern a/b/c/**` and file `a/b/c/file.nc`, `outputdir /x/y/z` will create
-                        output files like `/x/y/z/c/file_variable.nc`
-  --formatting FORMATTING
-                        Used in combination with `output-dir`: specifies the how to split the data and format the output file. Example: `_{time}_{level}hPa`
-  -d, --dry-run         Test the input file matching and the output file scheme without splitting.
-  -f, --force           Force re-splitting of the pipeline. Turns of skipping of already split data.
 ```
 
 _Common options_:
@@ -50,6 +33,7 @@ _Usage examples_:
 ```bash
 weather-sp --input-pattern 'gs://test-tmp/era5/2017/**' \
            --output-dir 'gs://test-tmp/era5/splits'
+           --formatting '.{typeOfLevel}'
 ```
 
 Preview splits with a dry run:
@@ -57,6 +41,7 @@ Preview splits with a dry run:
 ```bash
 weather-sp --input-pattern 'gs://test-tmp/era5/2017/**' \
            --output-dir 'gs://test-tmp/era5/splits' \
+           --formatting '.{typeOfLevel}'
            --dry-run
 ```
 
@@ -65,6 +50,7 @@ Using DataflowRunner
 ```bash
 weather-sp --input-pattern 'gs://test-tmp/era5/2015/**' \
            --output-dir 'gs://test-tmp/era5/splits'
+           --formatting '.{typeOfLevel}'
            --runner DataflowRunner \
            --project $PROJECT \
            --temp_location gs://$BUCKET/tmp  \
@@ -122,21 +108,19 @@ Based on the output directory path, the directory structure of the input pattern
 To create the directory structure, the common path of the input pattern is removed from the input file path and replaced
 with the output path. \
 The formatting specified by the `--formatting` flag is added between file name and ending and is
-used to determine along which dimensions to split.\
-If `--formatting` is not specified, file type-specific defaults are used:
-* GRIB: '_{typeOfLevel}_{shortName}'
-* NetCFD: '_{variable}'
+used to determine along which dimensions to split.
 
 Example:
 
 ```bash
 --input-pattern 'gs://test-input/era5/2020/**' \
 --output-dir 'gs://test-output/splits'
+--formatting '.{variable}'
 ```
 
 For a file `gs://test-input/era5/2020/02/01.nc` the output file pattern is
-`gs://test-output/splits/2020/02/01_{variable}.nc` and if the temperature is a variable in that data, the output file
-for that split will be `gs://test-output/splits/2020/02/01_t.nc`.
+`gs://test-output/splits/2020/02/01.{variable}.nc` and if the temperature is a variable in that data, the output file
+for that split will be `gs://test-output/splits/2020/02/01.t.nc`.
 
 Example:
 
--- a/weather_sp/README.md
+++ b/weather_sp/README.md
@@ -11,18 +11,20 @@ Splits NetCDF and Grib files into several files by variable (_alpha_).
 ## Usage
 
 ```
-usage: weather-sp [-h] -i INPUT_PATTERN -o OUTPUT_DIR [-d]
+usage: weather-sp [-h] -i INPUT_PATTERN -output-dir OUTPUT_DIR [-d]
 
-Split weather data file into files by variable.
+Split weather data file into files by specified dimensions.
 ```
 
 _Common options_:
 
 * `-i, --input-pattern`: Pattern for input weather data.
-* `--output-template`: Path to template using Python formatting, see [Output section](#output) below. Mutually exclusive
+* `--output-template`: Template of output file path using Python formatting, see [Output section](#output) below. Mutually exclusive
   with `--output-dir`.
 * `--output-dir`: Path to base folder for output files, see [Output section](#output) below. Mutually exclusive
   with `--output-template`
+* `--formatting`: Used only with `--output-dir` to specify splitting and output format
+   using Python formatting, see [Output section](#output) below.
 * `-f, --force`: Force re-splitting of the pipeline. Turns of skipping of already split data.
 * `-d, --dry-run`: Test the input file matching and the output file scheme without splitting.
 
@@ -75,17 +77,39 @@ On the other hand, to specify a specific pattern use
 --input-pattern 'gs://test-tmp/era5/2017/*/*.nc'
 ```
 
-## Output
+## Output & Dimensions to split
 
 The base output file names are specified using the `--output-template` or `--output-dir` flags. These flags are mutually
-exclusive, and one of them is required.
+exclusive, and one of them is required. \
+The output formatting also specifies which dimensions to split by using Python formatting.
+For example, adding `{time}` in the output template or formatting will cause the file to be split along the time dimension.
+
+### Available dimensions to split
+
+How a file can be split depends on the file type.
+
+#### GRIB
+GRIB files can be split along any dimension that is available in the file's metadata. \
+Examples: 'typeOfLevel', 'level', 'step', 'shortName', 'gridType', 'time', 'forecastTime'
+
+#### NetCDF
+NetCDF files are already in a hypercube format and can only be split by one of the dimensions and by data variable.
+Since splitting by latitude or longitude would lead to a large number of small files, this is not supported,
+and it is recommended to use the weather-mv tool instead. \
+Supported splits for NetCDF files are thus 'variable', 'time', 'level'.
+
 
 ### Output directory
 
 Based on the output directory path, the directory structure of the input pattern is replicated.
 
 To create the directory structure, the common path of the input pattern is removed from the input file path and replaced
-with the output path. A '_' is added to separate the split level / variable.
+with the output path. \
+The formatting specified by the `--formatting` flag is added between file name and ending and is
+used to determine along which dimensions to split.\
+If `--formatting` is not specified, file type-specific defaults are used:
+* GRIB: '_{typeOfLevel}_{shortName}'
+* NetCFD: '_{variable}'
 
 Example:
 
@@ -95,26 +119,38 @@ Example:
 ```
 
 For a file `gs://test-input/era5/2020/02/01.nc` the output file pattern is
-`gs://test-output/splits/2020/02/01.{shortname}.nc` and if the temperature is a variable in that data, the output file
-for that split will be `gs://test-output/splits/2020/02/01.t.nc`
+`gs://test-output/splits/2020/02/01_{variable}.nc` and if the temperature is a variable in that data, the output file
+for that split will be `gs://test-output/splits/2020/02/01_t.nc`.
+
+Example:
+
+```bash
+--input-pattern 'gs://test-input/era5/2020/**' \
+--output-dir 'gs://test-output/splits'
+--formatting '_{date}:{time}_{level}hPa'
+```
+For a file `gs://test-input/era5/2020/02/01.grib` the output file pattern is
+`gs://test-output/splits/2020/02/01_{date}:{time}_{level}hPa.grib` and for date = 20200101, time = 1200, level = 400,
+ the output file for that split will be `gs://test-output/splits/2020/02/01_20200101:1200_400hPa.grib`.
+
 
 ### Output template with Python-style formatting
 
 Using Python-style substitution (e.g. `{1}`) allows for more flexibility when creating the output files. The
 substitutions are based on the directory structure of the input file, where each `{<x>}` stands for one directory name,
 counting backwards from the end, i.e. the file name is `{0}`, the immediate directory in which it is located is `{1}`,
-and so on. In addition, you may supply `{shortname}` or `{levelType}` in the output template. These will be filled by values
-found in each file (NetCDF files typically only have one type of level, so this variable is not needed).
+and so on. In addition, you need to supply the split dimensions in the output template. These will be filled by values
+found in each file.
 
 Example:
 
 ```bash
 --input-pattern 'gs://test-input/era5/2020/**' \
---output-template 'gs://test-output/splits/{2}.{0}.{1}T00.{shortname}.nc'
+--output-template 'gs://test-output/splits/{2}.{0}.{1}T00.{variable}.nc'
 ```
 
 For a file `gs://test-input/era5/2020/02/01.nc` the output file pattern is
-`gs://test-output/splits/2020.01.02T00.{shortname}.nc` and if the temperature is a variable in that data, the output
+`gs://test-output/splits/2020.01.02T00.{variable}.nc` and if the temperature is a variable in that data, the output
 file for that split will be `gs://test-output/splits/2020.01.02T00.t.nc`
 
 ## Dry run
--- a/weather_sp/splitter_pipeline/file_name_utils.py
+++ b/weather_sp/splitter_pipeline/file_name_utils.py
@@ -24,7 +24,9 @@ NETCDF_FILE_ENDINGS = ('.nc', '.cd')
 
 class OutFileInfo(t.NamedTuple):
     file_name_template: str
-    ending: str
+    formatting: str = ''
+    ending: str = ''
+    template_folders: t.List[str] = []
     output_dir: bool = False
 
     def __str__(self):
@@ -34,7 +36,8 @@ class OutFileInfo(t.NamedTuple):
 def get_output_file_base_name(filename: str,
                               input_base_dir: str = '',
                               out_pattern: t.Optional[str] = None,
-                              out_dir: t.Optional[str] = None) -> OutFileInfo:
+                              out_dir: t.Optional[str] = None,
+                              formatting: str = '') -> OutFileInfo:
     """Construct the base output file name by applying the out_pattern to the filename.
 
     Example:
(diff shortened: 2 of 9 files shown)


weather_sp/README.md
```diff
@@ -11,18 +11,20 @@ Splits NetCDF and Grib files into several files by variable (_alpha_).
 ## Usage
 
 ```
-usage: weather-sp [-h] -i INPUT_PATTERN -o OUTPUT_DIR [-d]
+usage: weather-sp [-h] -i INPUT_PATTERN -output-dir OUTPUT_DIR [-d]
```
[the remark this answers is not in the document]
alxmrs: ...and if you have done that already... then I'm surprised by the output. :)

weather_sp/README.md
```diff
@@ -75,17 +77,39 @@ On the other hand, to specify a specific pattern use
...
+
+#### GRIB
+GRIB files can be split along any dimension that is available in the file's metadata. \
+Examples: 'typeOfLevel', 'level', 'step', 'shortName', 'gridType', 'time', 'forecastTime'
+
+#### NetCDF
+NetCDF files are already in a hypercube format and can only be split by one of the dimensions and by data variable.
+Since splitting by latitude or longitude would lead to a large number of small files, this is not supported,
+and it is recommended to use the weather-mv tool instead. \
+Supported splits for NetCDF files are thus 'variable', 'time', 'level'.
```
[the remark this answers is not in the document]
alxmrs: These both might be examples of non-index coordinates.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -22,20 +22,47 @@
...
+        """Construct base output file name with formatting marks."""
+        return self.file_name_template + self.formatting + self.ending
+
+    def set_formatting_if_needed(self, formatting: str):
+        """If formatting string is required but not present, set formatting.
+
+        Does not overwrite already present formatting.
+        """
+        if self.requires_formatting and not self.formatting:
+            self.formatting = formatting
```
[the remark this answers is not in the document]
alxmrs: Maybe, there could be two format functions, each pure, one for out_dir and one for out_pattern?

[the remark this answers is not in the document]
alxmrs: Maybe, there could be two format functions, each pure, one for out_dir and one for out_pattern?

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
 
 class GribSplitter(FileSplitter):
 
+    def __init__(self, input_path: str, output_info: OutFileInfo,
+                 force_split: bool = False, logging_level: int = logging.INFO):
+        super().__init__(input_path, output_info,
+                         force_split, logging_level)
+        self.output_info.set_formatting_if_needed('_{typeOfLevel}_{shortName}')
+        self.split_dims = self._get_split_dims()
```
alxmrs, line 93: If this is a property of `output_info` and `set_formatting_if_needed` is moved to a pure function on `output_info`, then there would be no need to overload the constructor. Indeed, the formatting in question could be static data in each Splitter class.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
...
+
     def split_data(self) -> None:
+        if not self.split_dims:
+            raise ValueError('No splitting specified in template.')
         outputs = dict()
 
         if self.should_skip():
             metrics.Metrics.counter('file_splitters', 'skipped').inc()
-            self.logger.info('Skipping %s, file already split.', repr(self.input_path))
+            self.logger.info('Skipping %s, file already split.',
```
alxmrs, line 102: Optional nit for the format string: it could be:
```
self.logger.info('Skipping {self.input_path!r}, file already split.')
```
Though... is there a performance reason why you'd want to keep it the same? I remember discussing this with @fredzyda, and I don't remember the conclusion of this conversation.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
...
+                 force_split: bool = False, logging_level: int = logging.INFO):
+        super().__init__(input_path, output_info,
+                         force_split, logging_level)
+        self.output_info.set_formatting_if_needed('_{typeOfLevel}_{shortName}')
+        self.split_dims = self._get_split_dims()
+
     def split_data(self) -> None:
+        if not self.split_dims:
+            raise ValueError('No splitting specified in template.')
         outputs = dict()
```
alxmrs, line 98: Optional nit: Move this line after the `should_skip` if block, which returns early.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
...
 
         with self._open_grib_locally() as grbs:
             for grb in grbs:
-                key = SplitKey(grb.typeOfLevel, grb.shortName)
+                splits = dict()
+                for dim in self.split_dims:
+                    try:
+                        splits[dim] = getattr(grb, dim)
+                    except RuntimeError:
+                        self.logger.error(
```
alxmrs, line 113: Are you sure we want to log a failure instead of throw the error?

alxmrs: Let me see if there's a test case for this...

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
...
 
     @contextmanager
     def _open_grib_locally(self) -> t.Iterator[t.Iterator[pygrib.gribmessage]]:
         with self._copy_to_local_file() as local_file:
             yield pygrib.open(local_file.name)
 
-    def _open_outfile(self, key: SplitKey):
-        return FileSystems.create(self._get_output_file_path(key))
+    def _open_outfile(self, key: str):
+        return FileSystems.create(key)
```
alxmrs, line 132: Let's replace invocation of this method with the direct `FileSystems.create` call.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
...
 
 
 class NetCdfSplitter(FileSplitter):
 
+    def __init__(self, input_path: str, output_info: OutFileInfo,
+                 force_split: bool = False, logging_level: int = logging.INFO):
+        super().__init__(input_path, output_info,
+                         force_split, logging_level)
+        self.output_info.set_formatting_if_needed('_{variable}')
+        self.split_dims = self._get_split_dims()
```
alxmrs, line 142: See comments above about constructors.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
...
-
-        with self._open_dataset_locally() as nc_data:
-            fields = [var for var in nc_data.variables.keys() if
-                      var not in nc_data.dimensions.keys()]
-            for field in fields:
-                self._create_netcdf_dataset_for_variable(nc_data, field)
-            self.logger.info('split %s into %d files', self.input_path, len(fields))
+        if any(split not in ('time', 'level', 'variable') for split in self.split_dims):
+            raise ValueError(
+                'netcdf split: unknown split dimension, supported are time, level, variable')
```
alxmrs, line 154: If not now, in v2 of this feature, this should be supported.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
...
+            for comb in combinations:
+                selected = comb[0]
+                for da in comb[1:]:
+                    if 'time' in da.coords:
+                        selected = selected.sel(time=da.time)
+                    if 'level' in da.coords:
+                        selected = selected.sel(level=da.level)
+                self._write_dataset(selected)
+            self.logger.info('split %s into %d files',
+                             self.input_path, len(combinations))
```
alxmrs, line 176: It's a bit late in the day, I'll take a closer look at this block tomorrow. I do think that XArray's indexing and selection capabilities should provide enough affordances for arbitrary selection of chunks of datasets.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
...
-                        dest[name][:] = dataset[name][:]
-            temp_file.flush()
-            self._copy_dataset_to_storage(temp_file,
-                                          self._get_output_file_path(
-                                              SplitKey('', variable)))
+            yield xr.open_dataset(local_file.name)
+
+    def _write_dataset(self, dataset: xr.Dataset()) -> None:
+        with FileSystems().create(self._get_output_for_dataset(dataset)) as dest_file:
+            dest_file.write(dataset.to_netcdf())
```
alxmrs, line 185: Nice

alxmrs (commented):
Will look at the core of NetCDF splitting tomorrow.

weather_sp/README.md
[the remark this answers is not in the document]
uhager: Didn't realize this was supposed to be the -h output. In hindsight, it's obvious...

weather_sp/README.md
```diff
@@ -75,17 +77,39 @@ On the other hand, to specify a specific pattern use
...
+exclusive, and one of them is required. \
+The output formatting also specifies which dimensions to split by using Python formatting.
+For example, adding `{time}` in the output template or formatting will cause the file to be split along the time dimension.
+
+### Available dimensions to split
+
+How a file can be split depends on the file type.
+
+#### GRIB
+GRIB files can be split along any dimension that is available in the file's metadata. \
```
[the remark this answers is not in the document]
uhager: Edited to clarify.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -22,20 +22,47 @@
 NETCDF_FILE_ENDINGS = ('.nc', '.cd')
 
 
-class OutFileInfo(t.NamedTuple):
-    file_name_template: str
-    ending: str
-    output_dir: bool = False
+class OutFileInfo:
+    """Holds data required to construct an output file name."""
```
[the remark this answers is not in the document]
uhager: Added to doc.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -22,20 +22,47 @@
...
+
+    def set_formatting_if_needed(self, formatting: str):
+        """If formatting string is required but not present, set formatting.
+
+        Does not overwrite already present formatting.
+        """
+        if self.requires_formatting and not self.formatting:
+            self.formatting = formatting
+
+    def get_formatted_output_file_path(self, splits: t.Dict[str, str]) -> str:
```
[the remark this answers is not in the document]
uhager: Renamed. just format seems a bit short though.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -68,20 +62,21 @@ def _copy_dataset_to_storage(self, src_file: t.IO, target: str):
         with FileSystems().create(target) as dest_file:
             shutil.copyfileobj(src_file, dest_file)
 
-    def _get_output_file_path(self, key: SplitKey) -> str:
-        split_keys = key._asdict()
-        if self.output_info.output_dir and key.levelType:
-            split_keys['levelType'] = f'{key.levelType}_'
-        return self.output_info.file_name_template.format(**split_keys)
+    def _get_output_file_path(self, splits: t.Dict[str, str]) -> str:
+        return self.output_info.get_formatted_output_file_path(splits)
```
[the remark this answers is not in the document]
uhager: Removed.

weather_sp/splitter_pipeline/file_name_utils.py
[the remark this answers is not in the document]
uhager: I don't really want to split those. The reason I made this mutable is that I need to be able to set a default formatting if the user only gives an output directory. So we either create the outFileInfo first and then update the formatting based on the splitter type, or we create the splitter first and then set the splitter's OutFileInfo, or create the OutFileInfo as part of the splitter constructor.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +85,132 @@ def should_skip(self):
 
 class GribSplitter(FileSplitter):
 
+    def __init__(self, input_path: str, output_info: OutFileInfo,
+                 force_split: bool = False, logging_level: int = logging.INFO):
+        super().__init__(input_path, output_info,
+                         force_split, logging_level)
+        self.output_info.set_formatting_if_needed('_{typeOfLevel}_{shortName}')
```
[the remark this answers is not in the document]
uhager: I'm not sure I understand how you want this here. My goal was to set this in a way that does not accidentally overwrite the value the user has specified, and also to avoid repeatedly having to check whether the default is needed.

weather_sp/README.md
```diff
@@ -75,17 +77,39 @@ On the other hand, to specify a specific pattern use
...
+Examples: 'typeOfLevel', 'level', 'step', 'shortName', 'gridType', 'time', 'forecastTime'
+
+#### NetCDF
+NetCDF files are already in a hypercube format and can only be split by one of the dimensions and by data variable.
+Since splitting by latitude or longitude would lead to a large number of small files, this is not supported,
```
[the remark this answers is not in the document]
uhager: Done.

weather_sp/README.md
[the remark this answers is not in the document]
uhager: NetCdf doesn't have step. At least none of the once that I have found. I don't have an ensemble file example, so I don't know if that would be possible.
For now, only the listed splits are supported, making netcdf splitting more general is really tricky. And I don't know if it's possible to cleanly split along a coordinate that is not a dimension.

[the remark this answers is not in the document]
uhager: For now, I would like to leave this and possibly consider adding more dimensions in the future.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -22,20 +22,47 @@
 NETCDF_FILE_ENDINGS = ('.nc', '.cd')
 
 
-class OutFileInfo(t.NamedTuple):
-    file_name_template: str
-    ending: str
-    output_dir: bool = False
+class OutFileInfo:
```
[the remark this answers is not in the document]
uhager: Done.

weather_sp/README.md
```diff
@@ -75,17 +77,39 @@ On the other hand, to specify a specific pattern use
 --input-pattern 'gs://test-tmp/era5/2017/*/*.nc'
 ```
 
-## Output
+## Output & Dimensions to split
```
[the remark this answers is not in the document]
uhager: I initially tried using a separate list for split dimensions and the formatting, but gave up about 2 min after I had a prototype and got the first error because I modified one but not the other...

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -22,20 +22,47 @@
...
+        self.formatting = formatting
+        self.ending = ending
+        self.template_folders = template_folders
+        self.requires_formatting = requires_formatting
 
     def __str__(self):
-        return f'{self.file_name_template}/*/{self.ending}'
+        return self.get_unformatted_output_name()
+
+    def get_unformatted_output_name(self):
```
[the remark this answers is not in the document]
uhager: Renamed. I don't really see the advantage to making it a property?

weather_sp/splitter_pipeline/file_name_utils_test.py
```diff
@@ -14,58 +14,97 @@
...
+                                        out_pattern='gs://my_bucket/splits/{2}-{1}-{0}_old_data.{variable}',
+                                        out_dir=None,
+                                        input_base_dir='ignored')
+        self.assertEqual(out_info.file_name_template,
+                         'gs://my_bucket/splits/{2}-{1}-{0}_old_data.{variable}')
+        self.assertEqual(out_info.template_folders, [
+                         '21', '01', '2020', 'data_to_split', 'my_bucket', 'gs:'])
+        self.assertEqual(out_info.ending, '')
+        self.assertEqual(out_info.formatting, '')
+        self.assertEqual(out_info.requires_formatting, False)
```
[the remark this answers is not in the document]
uhager: Done.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -68,20 +62,21 @@ def _copy_dataset_to_storage(self, src_file: t.IO, target: str):
...
-        if self.output_info.output_dir and key.levelType:
-            split_keys['levelType'] = f'{key.levelType}_'
-        return self.output_info.file_name_template.format(**split_keys)
+    def _get_output_file_path(self, splits: t.Dict[str, str]) -> str:
+        return self.output_info.get_formatted_output_file_path(splits)
+
+    def _get_split_dims(self) -> t.List[str]:
+        all_format = list(filter(None, [field[1] for field in string.Formatter().parse(
+            self.output_info.get_unformatted_output_name())]))
+        return [key for key in all_format if not key.isdigit()]
```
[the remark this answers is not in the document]
uhager: Moved split_dims to OutFileInfo.

weather_sp/splitter_pipeline/file_splitters.py
uhager: Done, removed this method

weather_sp/splitter_pipeline/file_splitters.py
uhager: I'm honestly not sure how to do this, I'll have to think more and do some testing.

weather_sp/splitter_pipeline/file_splitters.py
uhager: This has been moved.

weather_sp/splitter_pipeline/file_splitters.py
uhager: I was planning to look more into this as well to get this more flexible. If you have some example NetCDF files with more dimensions / coordinates that would be helpful for testing. However, for this first version I would like to leave this and tackle NetCDF improvements in a separate PR.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -22,20 +24,56 @@
...
+                    when using --output-dir.
+        ending: file ending.
+        template_folders:
+    """
     file_name_template: str
-    ending: str
-    output_dir: bool = False
+    formatting: str = ''
+    ending: str = ''
+    template_folders: t.List[str] = field(default_factory=lambda: [])
```
alxmrs, line 42: A typical pattern in python is to set mutable data structures to None, and then set it to an empty collection. 

https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

So, in this case, this would be a `t.Optional[t.List[str]]`, and in the `__post_init__` method, you'd check for None and set it to an empty list.

weather_sp/splitter_pipeline/file_splitters.py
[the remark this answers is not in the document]
uhager: Removed this. No longer needed if we don't set defaults.

weather_sp/splitter_pipeline/file_name_utils.py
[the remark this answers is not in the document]
uhager: No longer needed if we don't set defaults.

weather_sp/README.md
```diff
@@ -11,18 +11,35 @@ Splits NetCDF and Grib files into several files by variable (_alpha_).
...
+  --output-template OUTPUT_TEMPLATE
+                        Template specifying path to output files using python-style formatting substitution of input directory names. For `input_pattern a/b/c/**` and file
+                        `a/b/c/file.grib`, a template with formatting `/somewhere/{1}-{0}.{level}_{shortName}.grib` will give `somewhere/c-file.level_shortName.grib`
+  --output-dir OUTPUT_DIR
+                        Output directory that will replace the common path of the input_pattern. For `input_pattern a/b/c/**` and file `a/b/c/file.nc`, `outputdir /x/y/z` will create
+                        output files like `/x/y/z/c/file_variable.nc`
+  --formatting FORMATTING
+                        Used in combination with `output-dir`: specifies the how to split the data and format the output file. Example: `_{time}_{level}hPa`
+  -d, --dry-run         Test the input file matching and the output file scheme without splitting.
+  -f, --force           Force re-splitting of the pipeline. Turns of skipping of already split data.
```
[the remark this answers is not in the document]
uhager: Updated.

weather_sp/splitter_pipeline/file_name_utils.py
uhager: Made all parameters required instead, really makes more sense to set them explicitly.

weather_sp/README.md
```diff
@@ -33,13 +33,15 @@ _Usage examples_:
 ```bash
 weather-sp --input-pattern 'gs://test-tmp/era5/2017/**' \
            --output-dir 'gs://test-tmp/era5/splits'
+           --formatting '.{typeOfLevel}'
```
alxmrs, line 36: I think removing the defaults makes this flag combination really clear.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -22,20 +24,48 @@
...
+    def split_dims(self) -> t.List[str]:
+        all_format = list(filter(None, [field[1] for field in string.Formatter().parse(
+            self.unformatted_output_path())]))
+        return [key for key in all_format if not key.isdigit()]
 
-    def __str__(self):
-        return f'{self.file_name_template}/*/{self.ending}'
+    def formatted_output_path(self, splits: t.Dict[str, str]) -> str:
+        """Construct output file name with formatting applied"""
+        return self.unformatted_output_path().format(*self.template_folders, **splits)
```
alxmrs, line 59: This is much cleaner, I like the latest changes.

weather_sp/splitter_pipeline/file_name_utils_test.py
```diff
@@ -14,58 +14,86 @@
...
 
-    def test_get_output_file_base_name_format(self):
-        out_info = get_output_file_base_name(filename='gs://my_bucket/data_to_split/2020/01/21.nc',
-                                             out_pattern='gs://my_bucket/splits/{2}-{1}-{0}_old_data.',
-                                             out_dir=None,
-                                             input_base_dir='ignored')
-        self.assertEqual(out_info.file_name_template, 'gs://my_bucket/splits/2020-01-21_old_data.')
-        self.assertEqual(out_info.ending, '.nc')
+    def test_get_output_file_info_pattern(self):
+        out_info = get_output_file_info(filename='gs://my_bucket/data_to_split/2020/01/21.nc',
```
alxmrs, line 23: nit (optional): I see the common convention of `actual` vs `expected` pretty often.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -68,20 +61,19 @@ def _copy_dataset_to_storage(self, src_file: t.IO, target: str):
...
 
-    def _get_output_file_path(self, key: SplitKey) -> str:
-        split_keys = key._asdict()
-        if self.output_info.output_dir and key.levelType:
-            split_keys['levelType'] = f'{key.levelType}_'
-        return self.output_info.file_name_template.format(**split_keys)
+    def _get_split_dims(self) -> t.List[str]:
+        all_format = list(filter(None, [field[1] for field in string.Formatter().parse(
+            self.output_info.unformatted_output_path())]))
+        return [key for key in all_format if not key.isdigit()]
```
alxmrs, line 67: Is this still used? How does it compare to the split_dims on the dataclass?

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +82,123 @@ def should_skip(self):
...
-
-        with self._open_dataset_locally() as nc_data:
-            fields = [var for var in nc_data.variables.keys() if
-                      var not in nc_data.dimensions.keys()]
-            for field in fields:
-                self._create_netcdf_dataset_for_variable(nc_data, field)
-            self.logger.info('split %s into %d files', self.input_path, len(fields))
+        if any(split not in ('time', 'level', 'variable') for split in self.output_info.split_dims()):
+            raise ValueError(
+                'netcdf split: unknown split dimension, supported are time, level, variable')
```
alxmrs, line 144: We can land this for now, and generalize it in a future PR.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +82,123 @@ def should_skip(self):
...
+
+        with self._open_dataset_locally() as dataset:
+            if any(split not in dataset.dims and split not in ('variable') for split in self.output_info.split_dims()):
+                raise ValueError(
+                    'netcdf split: requested dimension not in dataset')
+            iterlists = []
+            iterlists.append([dataset[var].to_dataset(
+            ) for var in dataset.data_vars] if 'variable' in self.output_info.split_dims() else [dataset])
```
alxmrs, line 152: I find this a bit tricky to read. Maybe a (nested) for loop would be more readable?

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +82,123 @@ def should_skip(self):
...
+            for dim in ('time', 'level'):
+                if dim in self.output_info.split_dims():
+                    iterlists.append(dataset[dim])
+            combinations = list(itertools.product(*iterlists))
```
alxmrs, line 156: No need to cast this to a list if you're only iterating through the items from beginning to end.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -90,89 +82,123 @@ def should_skip(self):
...
+        with self._open_dataset_locally() as dataset:
+            if any(split not in dataset.dims and split not in ('variable') for split in self.output_info.split_dims()):
+                raise ValueError(
+                    'netcdf split: requested dimension not in dataset')
+            iterlists = []
+            iterlists.append([dataset[var].to_dataset(
+            ) for var in dataset.data_vars] if 'variable' in self.output_info.split_dims() else [dataset])
+            for dim in ('time', 'level'):
+                if dim in self.output_info.split_dims():
+                    iterlists.append(dataset[dim])
```
alxmrs, line 155: These two may be coordinate indexes. Check out this method: 
https://github.com/google/weather-tools/blob/0ee97b4502f63fae35f605511c3e468b3ac6d2fa/weather_mv/loader_pipeline/util.py#L92

Specifically this line: 
https://github.com/google/weather-tools/blob/0ee97b4502f63fae35f605511c3e468b3ac6d2fa/weather_mv/loader_pipeline/util.py#L106

This may be useful to you to get the allowed coordinate dimensions to iterate over.

weather_sp/splitter_pipeline/file_splitters.py
uhager: Oops! Forgot to clean that up! Thanks for spotting this!

weather_sp/splitter_pipeline/file_splitters.py
uhager: I'm pretty sure that I read that for logging %s style is used over {} python style.

weather_sp/splitter_pipeline/file_splitters.py
uhager: Done. I then have to cast it to get the length though for the log output.

weather_sp/splitter_pipeline/file_name_utils_test.py
uhager: Done.

weather_sp/splitter_pipeline/file_splitters.py
uhager: I'll try to make NetCDF splitting more flexible in a follow-up PR if that's ok.

weather_sp/splitter_pipeline/file_splitters.py
uhager: Separated into if-else.

weather_sp/README.md
```diff
@@ -75,46 +78,79 @@ On the other hand, to specify a specific pattern use
 --input-pattern 'gs://test-tmp/era5/2017/*/*.nc'
 ```
 
-## Output
+## Output & Dimensions to split
```
alxmrs, line 81: very optional nit: can we say "Split Dimensions"? Is that clear?

weather_sp/README.md
```diff
@@ -75,46 +78,79 @@ On the other hand, to specify a specific pattern use
...
 
 The base output file names are specified using the `--output-template` or `--output-dir` flags. These flags are mutually
-exclusive, and one of them is required.
+exclusive, and one of them is required. \
+The output formatting also specifies which dimensions to split by using Python formatting.
+For example, adding `{time}` in the output template or formatting will cause the file to be split along the time dimension.
```
alxmrs, line 86: For a future PR: It would be cool if the user could choose the interval of time to split over.

alxmrs (approved):
LGTM!

weather_sp/README.md
uhager: I've been thinking about this, and it's not easy given how netcfd and grib handle these differently. I'll have to look into that, but first I want to improve the splitting for netcdf to make it more flexible with different dimensions.

PR #130, opened 2022-03-29
Fixes issue (127) where append_date_dirs wasn't working properly.

pbattaglia:
* Fixed bug where partition_keys was being modified in-place, and causing problems laters.

* Adjusted how target_path is constructed so that trailing forward slashes in target_path are all removed.

* Added typing.cast to comply with pytype. I'm not sure why this wasn't triggered previously, but I was getting pytype errors from the prepare_target_name() function, which I was editing, so I fixed it in this PR.

### DIFF
--- a/weather_dl/download_pipeline/parsers.py
+++ b/weather_dl/download_pipeline/parsers.py
@@ -328,8 +328,9 @@ def process_config(file: t.IO) -> Config:
     num_partition_keys = len(partition_keys)
     if use_date_as_directory(config):
         num_partition_keys -= 1
-        if str(params.get('target_path')).endswith('/'):
-            params['target_path'] = params.get('target_path')[:-1]
+        target_path = t.cast(str, params.get('target_path', ''))
+        if target_path != '':
+            params['target_path'] = target_path.rstrip('/')
 
     require(num_template_replacements == num_partition_keys,
             """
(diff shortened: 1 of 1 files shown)


weather_dl/download_pipeline/parsers.py
```diff
@@ -328,8 +328,12 @@ def require(condition: bool, message: str) -> None:
     num_partition_keys = len(partition_keys)
     if use_date_as_directory(config):
         num_partition_keys -= 1
-        if str(params.get('target_path')).endswith('/'):
-            params['target_path'] = params.get('target_path')[:-1]
+        try:
+            target_path = t.cast(str, params['target_path'])
+        except KeyError:
+            pass
```
alxmrs, line 334: A more conventional method we use in our codebase is to call `get` with a default value. Would something like: 
```
target_path = t.cast(str, params.get('target_path', ''))
```
work here?

pbattaglia: Yep that should be fine. Updating the PR

alxmrs (approved):
LGTM, thanks!

weather_dl/download_pipeline/parsers.py
```diff
@@ -328,8 +328,9 @@ def require(condition: bool, message: str) -> None:
     num_partition_keys = len(partition_keys)
     if use_date_as_directory(config):
         num_partition_keys -= 1
-        if str(params.get('target_path')).endswith('/'):
-            params['target_path'] = params.get('target_path')[:-1]
+        target_path = t.cast(str, params.get('target_path', ''))
+        if target_path != '':
```
alxmrs, line 332: One last nit: It this if statement necessary? I think `rstrip` on an empty string will produce an empty string.

pbattaglia: Right, but I added it because you don't want to add a key to the dict if it doesn't already have it.

pbattaglia: Though, generally, I'm not actually sure why this is located here, because it seems like it'd be more natural to handle it in prepare_target_name().

On a similar note, I think subtracting 1 from num_partition_keys here, but actually removing the "date" partition key in prepare_target_name(), might be a little risky. Could they be done in the same place?

alxmrs: > Right, but I added it because you don't want to add a key to the dict if it doesn't already have it.
Ah, that makes sense. 

> it seems like it'd be more natural to handle it in prepare_target_name().
I agree – this would be a better fit: It'd be best if the parser had less logic, in general. 

> Could they be done in the same place?
Yes, that totally sounds right to me. 

Last: I think we are no longer using the `num_partition_keys` counter anymore. It would be prudent to pull that out.

PR #133, opened 2022-03-30
Code-changes for #49, added S2_LOCATION column

### LINKED ISSUE #49: `weather-mv`: Add columns to Schema for faster geolocation queries
A suggestion from @lakshmanok:  

> It would be good to if weather-mv were to add a `ST_GeogPoint(longitude, latitude)` column instead of (or in addition to) the longitude, latitude columns. This precomputes the [S2 location](https://s2geometry.io/) and makes querying faster.
>
> Even better if the column was a polygon with the extent of the pixel. That way `ST_INTERSECTS` etc. will also work

_ticket updated: 2022-03-24_

## Acceptance Criteria
* Users of the resulting BigQuery table produced by `weather-mv` can make use of the [Geography capabilities of BigQuery](https://cloud.google.com/bigquery/docs/geospatial-data).
  * Every row of the BigQuery table (and thus, all the available variables of that row) has a [column of type `GEOGRAPHY`](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#geography_type), which uses "point geography" from the raw data's lat/lng data.
* Users can still acquire the underling float values for latitude and longitude. 

## Implementation Notes
* One approach for implementing this feature would be to use BQ's SQL functions to add perform the conversion to `GEOGRAPHY` type. Namely, [`ST_GEOPOINT`](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_geogpoint) is a function that can take lat/lng FLOAT64 columns and create a `GEOGRAPHY` Point value from it. 
* Another simple approach would be to handle the Geography conversion from lat/lng in Python (instead of SQL). The BQ Geospatial docs list at least two pathways to do this: Either by writing [WKT / WKB data](https://cloud.google.com/bigquery/docs/geospatial-data#loading_wkt_or_wkb_data) or by writing in the [GeoJSON spec](https://cloud.google.com/bigquery/docs/geospatial-data#loading_geojson_data).
* Please note that the GEOGRAPHY values must fall within certain lat/lng ranges. This means that data outside of these ranges will need to be converted to those ranges (e.g. [something like this – see the example section](https://xarray.pydata.org/en/stable/generated/xarray.DataArray.assign_coords.html)).
* This will likely require a 2 part code change.
  * First, the BQ schema will need to be updated. https://github.com/google/weather-tools/blob/fd0c5e4f63b2bebea30d743e1f04f9c008ce914f/weather_mv/loader_pipeline/bq.py#L70
  * Second, the processing of the data into Dicts / JSON rows will need to change. https://github.com/google/weather-tools/blob/fd0c5e4f63b2bebea30d743e1f04f9c008ce914f/weather_mv/loader_pipeline/bq.py#L173 

Implementing extra columns to include the polygon of the (grid) area where the values are relevant is a great bonus feature. However, it will not be possible for all types of input data. This experience is only possible if the XArray DataFrame includes extra information, like a coordinate or variable attribute. 

For now, while we implement the first part of this ticket, let's also investigate if such metadata exists on our happy-path data sources. If it does, we'll create a follow up ticket to implement this. 

### Acceptance Criteria
* Investigate the underlying real-time data source in XArray to determine if variables along coordinates are associated with a specific area.
* Demonstrate results in an interactive report (e.g. a Colab notebook). 
* If this is possible with the data, create a design doc to further extend the BQ ingestion to include this feature, taking into account edge cases (the metadata is not available, users may want to override these columns, complexity in types of grid, etc.).

ksic8:
Fixed #49 added S2_LOCATION column

### DIFF
--- a/weather_mv/loader_pipeline/bq.py
+++ b/weather_mv/loader_pipeline/bq.py
@@ -18,6 +18,7 @@ import logging
 import typing as t
 
 import apache_beam as beam
+import geojson
 import numpy as np
 import pandas as pd
 import xarray as xr
@@ -35,9 +36,8 @@ DEFAULT_IMPORT_TIME = datetime.datetime.utcfromtimestamp(0).replace(tzinfo=datet
 DATA_IMPORT_TIME_COLUMN = 'data_import_time'
 DATA_URI_COLUMN = 'data_uri'
 DATA_FIRST_STEP = 'data_first_step'
-S2_LOCATION = 's2_location'  # st_geog_point
-MIN_LATITUDE = -90
-MAX_LATITUDE = 90
+GEO_POINT_COLUMN = 'geo_point'
+LATITUDE_RANGE = (-90, 90)
 
 
 @dataclasses.dataclass
@@ -146,21 +146,18 @@ def to_table_schema(columns: t.List[t.Tuple[str, str]]) -> t.List[bigquery.Schem
     fields.append(bigquery.SchemaField(DATA_IMPORT_TIME_COLUMN, 'TIMESTAMP', mode='NULLABLE'))
     fields.append(bigquery.SchemaField(DATA_URI_COLUMN, 'STRING', mode='NULLABLE'))
     fields.append(bigquery.SchemaField(DATA_FIRST_STEP, 'TIMESTAMP', mode='NULLABLE'))
-    fields.append(bigquery.SchemaField(S2_LOCATION, 'GEOGRAPHY', mode='NULLABLE'))
+    fields.append(bigquery.SchemaField(GEO_POINT_COLUMN, 'GEOGRAPHY', mode='NULLABLE'))
 
     return fields
 
 
-def fetch_s2_location(lat, long):
-    """
-    This function calculates and returns s2_location
-    from the input latitude and longitude values
-    """
-    if lat > MAX_LATITUDE or lat < MIN_LATITUDE:
+def fetch_geo_point(lat: float, long: float) -> str:
+    """Calculates a geography point from an input latitude and longitude."""
+    if lat > LATITUDE_RANGE[1] or lat < LATITUDE_RANGE[0]:
         raise ValueError(f"Invalid latitude value '{lat}'")
     long = ((long + 180) % 360) - 180
-    row = '{"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[%d,%d]}]}' % (long, lat)
-    return row
+    point = geojson.dumps(geojson.Point((long, lat)))
+    return point
 
 
 def extract_rows(uri: str, *,
@@ -209,7 +206,7 @@ def extract_rows(uri: str, *,
             row[DATA_IMPORT_TIME_COLUMN] = import_time
             row[DATA_URI_COLUMN] = uri
             row[DATA_FIRST_STEP] = first_time_step
-            row[S2_LOCATION] = fetch_s2_location(row['latitude'], row['longitude'])
+            row[GEO_POINT_COLUMN] = fetch_geo_point(row['latitude'], row['longitude'])
 
             # 'row' ends up looking like:
             # {'latitude': 88.0, 'longitude': 2.0, 'time': '2015-01-01 06:00:00', 'd': -2.0187, 'cc': 0.007812,
--- a/weather_mv/loader_pipeline/bq.py
+++ b/weather_mv/loader_pipeline/bq.py
@@ -35,6 +35,9 @@ DEFAULT_IMPORT_TIME = datetime.datetime.utcfromtimestamp(0).replace(tzinfo=datet
 DATA_IMPORT_TIME_COLUMN = 'data_import_time'
 DATA_URI_COLUMN = 'data_uri'
 DATA_FIRST_STEP = 'data_first_step'
+S2_LOCATION = 's2_location'  # st_geog_point
+MIN_LATITUDE = -90
+MAX_LATITUDE = 90
 
 
 @dataclasses.dataclass
@@ -143,8 +146,21 @@ def to_table_schema(columns: t.List[t.Tuple[str, str]]) -> t.List[bigquery.Schem
     fields.append(bigquery.SchemaField(DATA_IMPORT_TIME_COLUMN, 'TIMESTAMP', mode='NULLABLE'))
     fields.append(bigquery.SchemaField(DATA_URI_COLUMN, 'STRING', mode='NULLABLE'))
     fields.append(bigquery.SchemaField(DATA_FIRST_STEP, 'TIMESTAMP', mode='NULLABLE'))
+    fields.append(bigquery.SchemaField(S2_LOCATION, 'GEOGRAPHY', mode='NULLABLE'))
 
     return fields
+
+
+def fetch_s2_location(lat, long):
+    """
+    This function calculates and returns s2_location
+    from the input latitude and longitude values
+    """
+    if lat > MAX_LATITUDE or lat < MIN_LATITUDE:
+        raise ValueError(f"Invalid latitude value '{lat}'")
+    long = ((long + 180) % 360) - 180
+    row = '{"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[%d,%d]}]}' % (long, lat)
+    return row
 
 
 def extract_rows(uri: str, *,
@@ -193,6 +209,7 @@ def extract_rows(uri: str, *,
             row[DATA_IMPORT_TIME_COLUMN] = import_time
             row[DATA_URI_COLUMN] = uri
             row[DATA_FIRST_STEP] = first_time_step
+            row[S2_LOCATION] = fetch_s2_location(row['latitude'], row['longitude'])
 
             # 'row' ends up looking like:
             # {'latitude': 88.0, 'longitude': 2.0, 'time': '2015-01-01 06:00:00', 'd': -2.0187, 'cc': 0.007812,
--- a/weather_mv/loader_pipeline/bq_test.py
+++ b/weather_mv/loader_pipeline/bq_test.py
@@ -17,13 +17,14 @@ import typing as t
 import unittest
 from contextlib import contextmanager
 from time import perf_counter
+import geojson
 
 import numpy as np
 import pandas as pd
 import xarray as xr
 from google.cloud.bigquery import SchemaField
 
-from .bq import dataset_to_table_schema, extract_rows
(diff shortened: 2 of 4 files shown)


weather_mv/loader_pipeline/bq.py
```diff
@@ -35,6 +35,9 @@
 DATA_IMPORT_TIME_COLUMN = 'data_import_time'
 DATA_URI_COLUMN = 'data_uri'
 DATA_FIRST_STEP = 'data_first_step'
+S2_LOCATION = 's2_location'  # st_geog_point
+MIN_LATITUDE = -90
+MAX_LATITUDE = 90
```
alxmrs, line 40: Optional nit: it might be appropriate to store this as a tuple, since these two values are address the same concern. WDYT?

weather_mv/loader_pipeline/bq.py
```diff
@@ -35,6 +35,9 @@
 DATA_IMPORT_TIME_COLUMN = 'data_import_time'
 DATA_URI_COLUMN = 'data_uri'
 DATA_FIRST_STEP = 'data_first_step'
+S2_LOCATION = 's2_location'  # st_geog_point
```
alxmrs, line 38: This is misnamed. A Geography point is not the same as an S2 location. 
See, for example, [these docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#s2_cellidfrompoint). From a Point, you can convert to a S2 location (an ID).

weather_mv/loader_pipeline/bq.py
```diff
@@ -143,10 +146,23 @@ def to_table_schema(columns: t.List[t.Tuple[str, str]]) -> t.List[bigquery.Schem
     fields.append(bigquery.SchemaField(DATA_IMPORT_TIME_COLUMN, 'TIMESTAMP', mode='NULLABLE'))
     fields.append(bigquery.SchemaField(DATA_URI_COLUMN, 'STRING', mode='NULLABLE'))
     fields.append(bigquery.SchemaField(DATA_FIRST_STEP, 'TIMESTAMP', mode='NULLABLE'))
+    fields.append(bigquery.SchemaField(S2_LOCATION, 'GEOGRAPHY', mode='NULLABLE'))
 
     return fields
 
 
+def fetch_s2_location(lat, long):
```
alxmrs, line 154: As mentioned above, this is not an s2 location, but a "geopoint".

weather_mv/loader_pipeline/bq.py
```diff
@@ -143,10 +146,23 @@ def to_table_schema(columns: t.List[t.Tuple[str, str]]) -> t.List[bigquery.Schem
...
+    """
```
alxmrs, line 155: As per the [Google Python style guide](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings), the docstring should be formatted a bit differently: 
- The description should be on the same line as the open block quote and end with a period. 
- This looks like it will be a single line docstring
- If this is multiline, please add a short 1-line description on the same line as the open-block quote, and then add a longer description in a paragraph below. 
- No need to say "this function", and "calculates and returns" is a bit redundant.

Specifically, this might look like: 
```
def function_name(args): 
    """Calculates a geography point from an input latitude and longitude.""" 
    # ...
```

weather_mv/loader_pipeline/bq.py
```diff
@@ -143,10 +146,23 @@ def to_table_schema(columns: t.List[t.Tuple[str, str]]) -> t.List[bigquery.Schem
...
+    This function calculates and returns s2_location
+    from the input latitude and longitude values
+    """
+    if lat > MAX_LATITUDE or lat < MIN_LATITUDE:
+        raise ValueError(f"Invalid latitude value '{lat}'")
+    long = ((long + 180) % 360) - 180
+    row = '{"type":"GeometryCollection","geometries":[{"type":"Point","coordinates":[%d,%d]}]}' % (long, lat)
```
alxmrs, line 162: Please use python fstrings

weather_mv/loader_pipeline/bq.py
alxmrs: Please add type annotations to the function.

weather_mv/loader_pipeline/bq.py
alxmrs: Should this be a string, or a dictionary? I thought in the documentation used GeoJSON via Python dictionaries?

alxmrs: I don't think this should be a collection of a single point of coordinates. Rather, it should be a single point. Right?

If so, then the inner dictionary (starting with "type":"Point") should be sufficient.

weather_mv/loader_pipeline/bq.py
```diff
@@ -193,6 +209,7 @@ def to_row(it: t.Dict) -> t.Dict:
             row[DATA_IMPORT_TIME_COLUMN] = import_time
             row[DATA_URI_COLUMN] = uri
             row[DATA_FIRST_STEP] = first_time_step
+            row[S2_LOCATION] = fetch_s2_location(row['latitude'], row['longitude'])
```
alxmrs, line 212: Have you tested this e2e with a BQ database? Does this work? May I see the resulting table with live data?

alxmrs (changes requested):

weather_mv/loader_pipeline/bq.py
ksic8: The column_name is changed to geo_location.

weather_mv/loader_pipeline/bq.py
ksic8: Type annotations added to said function.

weather_mv/loader_pipeline/bq.py
```diff
@@ -143,10 +146,23 @@ def to_table_schema(columns: t.List[t.Tuple[str, str]]) -> t.List[bigquery.Schem
...
 
 
+def fetch_s2_location(lat, long):
+    """
+    This function calculates and returns s2_location
+    from the input latitude and longitude values
+    """
+    if lat > MAX_LATITUDE or lat < MIN_LATITUDE:
+        raise ValueError(f"Invalid latitude value '{lat}'")
+    long = ((long + 180) % 360) - 180
```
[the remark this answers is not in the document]
ksic8: bq_test.py has been modified to cover unit-tests for fetch_geo_point function. 13 different combinations of valid & invalid (latitude longitude) values are checked.

weather_mv/loader_pipeline/bq.py
```diff
@@ -143,10 +146,20 @@ def to_table_schema(columns: t.List[t.Tuple[str, str]]) -> t.List[bigquery.Schem
...
 
     return fields
 
 
+def fetch_geo_point(lat: float, long: float) -> str:
+    """Calculates a geography point from an input latitude and longitude."""
+    if lat > LATITUDE_RANGE[1] or lat < LATITUDE_RANGE[0]:
+        raise ValueError(f"Invalid latitude value '{lat}'")
+    long = ((long + 180) % 360) - 180
+    point = geojson.dumps(geojson.Point((long, lat)))
```
alxmrs, line 159: Nice, this is very clear (and probably less error prone).

weather_mv/loader_pipeline/bq_test.py
```diff
@@ -222,9 +232,34 @@ def test_extract_rows_nan(self):
...
+            '{"type": "Point", "coordinates": [-45, -45]}',
+            '{"type": "Point", "coordinates": [0, 0]}',
+            '{"type": "Point", "coordinates": [45, 45]}',
+            '{"type": "Point", "coordinates": [-180, 45]}',
+            '{"type": "Point", "coordinates": [-1, 90]}',
+            '{"type": "Point", "coordinates": [0, 90]}'
+        ]
+        for x in range(len(valid_lat_long)):
+            expected = fetch_geo_point(valid_lat_long[x][0], valid_lat_long[x][1])
+            self.assertEqual(actual_val[x], expected)
```
alxmrs, line 255: This looks good! Consider making these all subtests.

https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

weather_mv/loader_pipeline/bq_test.py
```diff
@@ -222,9 +232,34 @@ def test_extract_rows_nan(self):
...
+            '{"type": "Point", "coordinates": [1, -90]}',
+            '{"type": "Point", "coordinates": [-180, -45]}',
+            '{"type": "Point", "coordinates": [-45, -45]}',
+            '{"type": "Point", "coordinates": [0, 0]}',
+            '{"type": "Point", "coordinates": [45, 45]}',
+            '{"type": "Point", "coordinates": [-180, 45]}',
+            '{"type": "Point", "coordinates": [-1, 90]}',
+            '{"type": "Point", "coordinates": [0, 90]}'
+        ]
+        for x in range(len(valid_lat_long)):
```
alxmrs, line 253: Why not:
```
for (lat, lng) in valid_lat_long:
     expected = fetch_geo_point(lat, lng)
```
?

weather_mv/loader_pipeline/bq_test.py
ksic8: Required code-changes completed.

weather_mv/loader_pipeline/bq_test.py
ksic8: Required code-changes completed.

weather_mv/loader_pipeline/bq_test.py
```diff
@@ -222,9 +232,36 @@ def test_extract_rows_nan(self):
...
+            '{"type": "Point", "coordinates": [0, 90]}'
+        ]
+        for actual, (lat, long) in zip(actual_val, valid_lat_long):
+            with self.subTest():
+                expected = fetch_geo_point(lat, long)
+                self.assertEqual(actual, expected)
+
+    def test_extract_rows__with_invalid_lat(self):
+        invalid_lat_long = [[-100, -2000], [-100, -500], [100, 500], [100, 2000]]
+        for i, (lat, long) in enumerate(invalid_lat_long):
```
alxmrs, line 260: Nit: It looks like the `i` variable is unused here.

alxmrs, line 260: Nit: It looks like the `i` variable is unused here.

ksic8: Suggested code-changes completed.

ksic8: Suggested code-changes completed.

alxmrs (approved):

Issue #118, opened 2022-03-18
Convert `Config` dict into a typed Dataclass


alxmrs:
`Config` has outgrown a dict of dicts representation. We should convert it to a dataclass and update the parser to output a type dataclass. This will prevent easy-to-make typo errors and make the code cleaner. 

_Originally posted by @fredzyda in https://github.com/google/weather-tools/pull/117#r830317579_

alxmrs:
xref: #5

alxmrs:
To expand on the details of this issue...

https://github.com/google/weather-tools/blob/ecb194c7bc62d9d0ad3982cce24eb745a470be84/weather_dl/download_pipeline/parsers.py#L30

Config should be a Dataclass. Each field of the dictionary (as outlined and enforced in the parser, see below) should be typed and documented accordingly. It's fine if some key-value pairs are dictionaries or of any type (e.g., like, the whole selelction section), but we should type things when we can. 

https://github.com/google/weather-tools/blob/ecb194c7bc62d9d0ad3982cce24eb745a470be84/weather_dl/download_pipeline/parsers.py#L261

fredzyda:
I think if we complete #5 by using protobuf, we might get this one for free. I think #5 makes sense as the first thing to work on, then maybe we should reevaluate this one.

alxmrs:
As discussed today in standup and in https://github.com/google/weather-tools/issues/5#issuecomment-1086191311, it does in fact make sense for us to maintain a parser (and not move to protobufs) for the downloader's config language. 

Given that, I'd like to offer pointers to how a config should be structured. This will prove key to creating the dataclass-based abstraction for configs in this issue. 

A Config is made of two sections of key-value pairs. The Parameters section, [documented here](https://github.com/google/weather-tools/blob/main/Configuration.md#parameters-section) has a finite set of required and optional fields, each with specific types. In addition, the Parameters section should be able to detect arbitrary key-value pairs, and these should be passed all the way to each client (basically, as keyword-arguments).

An additional complexity to these generic key-value pairs of the parameters section: Parameters can take named dictionaries of arbitrary key-value pairs in the form of parameter subsections ([documented here](https://github.com/google/weather-tools/blob/main/Configuration.md#subsections)). 

[...]

PR #142, opened 2022-04-08
Defined config dataclass structure

mahrsee1997:
#118 - Defined the structure of config dataclass.

### DIFF
--- a/weather_dl/download_pipeline/config.py
+++ b/weather_dl/download_pipeline/config.py
@@ -57,7 +57,7 @@ class Config:
     selection: t.Dict[str, Values] = dataclasses.field(default_factory=dict)
 
     @classmethod
-    def from_config(cls, config: dict):
+    def from_dict(cls, config: t.Dict) -> 'Config':
         config_instance = cls()
         for section_key, section_value in config.items():
             if section_key == "parameters":
--- a/weather_dl/download_pipeline/config.py
+++ b/weather_dl/download_pipeline/config.py
@@ -1,0 +1,71 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import typing as t
+import dataclasses
+
+Values = t.Union[t.List['Values'], t.Dict[str, 'Values'], bool, int, float, str]  # pytype: disable=not-supported-yet
+
+
+@dataclasses.dataclass
+class Config:
+    """Contains pipeline parameters.
+
+    Attributes:
+        client:
+            Name of the Weather-API-client. Supported clients are mentioned in the 'CLIENTS' variable.
+        dataset (optional):
+            Name of the target dataset. Allowed options are dictated by the client.
+        partition_keys (optional):
+            Choose the keys from the selection section to partition the data request.
+            This will compute a cartesian cross product of the selected keys
+            and assign each as their own download.
+        target_path:
+            Download artifact filename template. Can make use of Python's standard string formatting.
+            It can contain format symbols to be replaced by partition keys;
+            if this is used, the total number of format symbols must match the number of partition keys.
+        subsection_name:
+            Name of the particular subsection. 'default' if there is no subsection.
+        force_download:
+            Force redownload of partitions that were previously downloaded.
+        user_id:
+            Username from the environment variables.
+        kwargs (optional):
+            For representing subsections or any other parameters.
+        selection:
+            Contains parameters used to select desired data.
+    """
+
+    client: str = ""
+    dataset: t.Optional[str] = ""
+    target_path: str = ""
+    partition_keys: t.Optional[t.List[str]] = dataclasses.field(default_factory=list)
+    subsection_name: str = "default"
+    force_download: bool = False
+    user_id: str = "unknown"
+    kwargs: t.Optional[t.Dict[str, Values]] = dataclasses.field(default_factory=dict)
+    selection: t.Dict[str, Values] = dataclasses.field(default_factory=dict)
+
+    @classmethod
+    def from_config(cls, config: dict):
+        config_instance = cls()
+        for section_key, section_value in config.items():
+            if section_key == "parameters":
+                for key, value in section_value.items():
+                    if hasattr(config_instance, key):
+                        setattr(config_instance, key, value)
+                    else:
+                        config_instance.kwargs[key] = value
+            if section_key == "selection":
+                config_instance.selection = section_value
+        return config_instance
--- a/weather_dl/download_pipeline/fetcher.py
+++ b/weather_dl/download_pipeline/fetcher.py
@@ -75,14 +75,11 @@ class Fetcher(beam.DoFn):
 
         client = CLIENTS[self.client_name](config)
         target = prepare_target_name(config)
-        dataset = config.dataset
-        selection = config.selection
-        user = config.user_id
 
-        with self.manifest.transact(selection, target, user):
+        with self.manifest.transact(config.selection, target, config.user_id):
             with tempfile.NamedTemporaryFile() as temp:
                 logger.info(f'[{worker_name}] Fetching data for {target!r}.')
-                self.retrieve(client, dataset, selection, temp.name)
+                self.retrieve(client, config.dataset, config.selection, temp.name)
 
                 logger.info(f'[{worker_name}] Uploading to store for {target!r}.')
                 self.upload(temp, target)
--- a/weather_dl/download_pipeline/fetcher.py
+++ b/weather_dl/download_pipeline/fetcher.py
@@ -23,6 +23,7 @@ from apache_beam.io.gcp.gcsio import WRITE_CHUNK_SIZE
 from .clients import CLIENTS, Client
 from .manifest import Manifest, NoOpManifest, Location
 from .parsers import prepare_target_name
+from .config import Config
 from .partition import skip_partition
 from .stores import Store, FSStore
 from .util import retry_with_exponential_backoff
@@ -64,7 +65,7 @@ class Fetcher(beam.DoFn):
         """Retrieve from download client, with retries."""
         client.retrieve(dataset, selection, dest)
 
-    def fetch_data(self, config: t.Dict, *, worker_name: str = 'default') -> None:
+    def fetch_data(self, config: Config, *, worker_name: str = 'default') -> None:
         """Download data from a client to a temp file, then upload to Cloud Storage."""
         if not config:
             return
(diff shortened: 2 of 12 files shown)


weather_dl/download_pipeline/parsers.py
```diff
@@ -30,6 +31,40 @@
...
+        other_params (optional): For representing subsections or any other parameters.
+        selection: Contains parameters used to select desired data.
+    """
+    client: str
+    dataset: t.Optional[str]
+    target_path: str
+    partition_keys: t.Optional[t.List[str]]
+    target_filename: t.Optional[str]
+    api_key: t.Optional[str]
+    api_url: t.Optional[str]
```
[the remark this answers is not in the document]
alxmrs: One argument for `other_params` is that the API authentication scheme will depend a lot on the client implementation (for example, key/url vs key/url/email). So, we shouldn't specify what is needed at a top level, and leave it to other_params.

alxmrs (changes requested):
Here's an initial review. I look forward to seeing how the dataclass gets used throughout the pipeline.

weather_dl/download_pipeline/parsers.py
```diff
@@ -30,6 +31,40 @@
...
+    __subsection__: str
+    num_api_keys: t.Optional[int]
```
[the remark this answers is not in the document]
mahrsee1997: Removed the code related to `num_api_keys` and kept `subsection_name` at the top level.

weather_dl/download_pipeline/parsers.py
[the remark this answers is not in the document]
mahrsee1997: Yes, it makes sense ... these should belong to `other_params`.

weather_dl/download_pipeline/parsers.py
```diff
@@ -30,6 +31,40 @@
...
+    force_download: bool
+    user_id: str
+    other_params: t.Optional[t.Dict[str, Values]]
```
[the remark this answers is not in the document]
mahrsee1997: Done.

weather_dl/download_pipeline/parsers.py
```diff
@@ -30,6 +31,38 @@
 Config = t.Dict[str, t.Dict[str, Values]]
 
 
+@dataclasses.dataclass()
```
alxmrs, line 34: Please remove the extra `()`.

weather_dl/download_pipeline/parsers.py
```diff
@@ -30,6 +31,38 @@
...
+class Config:
+    """Contains pipeline parameters
```
alxmrs, line 36: Please end with a full stop [`.`].

weather_dl/download_pipeline/parsers.py
```diff
@@ -30,6 +31,38 @@
...
+
+    Attributes:
+        client: Name of the Weather-API-client. Supported clients are mentioned in the 'CLIENTS' variable.
+        dataset (optional): Name of the target dataset. Allowed options are dictated by the client.
+        partition_keys (optional): Choose the keys from the selection section to partition the data request.
+                                   This will compute a cartesian cross product of the selected keys
+                                   and assign each as their own download.
+        target_path: Download artifact filename template. Can make use of Python's standard string formatting.
+                     It can contain format symbols to be replaced by partition keys;
+                     if this is used, the total number of format symbols must match the number of partition keys.
```
alxmrs, line 46: The copy is good here. However, the indentation is a bit unconventional. The the following style guide for examples on proper indentation: 

https://google.github.io/styleguide/pyguide.html#s3.8-comments-and-docstrings

alxmrs (commented):
Some more notes...

weather_dl/download_pipeline/parsers.py
```diff
@@ -321,8 +314,6 @@ def parse_subsections(config: t.Dict) -> t.Dict:
     for_cleanup = [key for key, _ in copy.items() if '.' in key]
     for target in for_cleanup:
         del copy[target]
-    if num_api_keys:
-        copy['parameters']['num_api_keys'] = num_api_keys
     return copy
```
alxmrs, line 317: Thanks for removing the dead code.

weather_dl/download_pipeline/parsers.py
```diff
@@ -159,11 +158,11 @@ def parse_config(file: t.IO) -> Config:
         config.read_file(file)
         config_by_section = {s: _parse_lists(config, s) for s in config.sections()}
         config_with_nesting = parse_subsections(config_by_section)
-        return config_with_nesting
+        return Config.from_config(config_with_nesting)
     except configparser.ParsingError:
         pass
 
-    return {}
+    return Config.from_config({})
```
alxmrs, line 165: Sorry to suggest this, because it may involve a lot of undoing, but: I think that it may be better to create a config object at the end of `process_config`, once all the validation steps have finished. I can think of two reasons why this makes sense to me; please let me know if you agree: 
1. The `parse_config` function is kind of our "lexing" & "parsing" phase ["Map of the territory", _[Crafting Interpreters]_]. It parses an abstract-syntax tree (in our case, JSON-like Python-Dicts) from raw text. Ideally, we shouldn't mix parsing and analysis. There are some advantages for keeping those steps separate; I see them as separate concerns. (In this analogy, I see the `Config` dataclass as an "Intermediate Representation".)
2. I argue it's a bit more efficient to check for syntax errors before creating the dataclass, which requires iterating through the entire nested object. 

What do you think about this? What are your thoughts about transforming to the dataclass earlier? 

[Crafting Interpreters]: https://craftinginterpreters.com/a-map-of-the-territory.html

weather_dl/download_pipeline/parsers.py
```diff
@@ -409,19 +402,17 @@ def require(condition: bool, message: str, error_type: t.Type[Exception] = Value
...
-    parameters = config['parameters']
-    target_path = t.cast(str, parameters.get('target_path', ''))
-    partition_keys = t.cast(t.List[str],
-                            cp.copy(parameters.get('partition_keys', list())))
+    target_path = config.target_path
+    partition_keys = config.partition_keys
 
-    partition_dict = OrderedDict((key, typecast(key, config['selection'][key][0])) for key in partition_keys)
+    partition_dict = OrderedDict((key, typecast(key, config.selection[key][0])) for key in partition_keys)
     target = target_path.format(*partition_dict.values(), **partition_dict)
```
alxmrs, line 416: Similar to the above point, I think it'd be nice just to access the fields on the config directly: 
```
    partition_dict = OrderedDict((key, typecast(key, config.selection[key][0])) for key in config.partition_keys)
    target = config.target_path.format(*partition_dict.values(), **partition_dict)
```

weather_dl/download_pipeline/config.py
```diff
@@ -0,0 +1,71 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import typing as t
+import dataclasses
+
+Values = t.Union[t.List['Values'], t.Dict[str, 'Values'], bool, int, float, str]  # pytype: disable=not-supported-yet
+
+
+@dataclasses.dataclass
+class Config:
```
alxmrs, line 21: I see the `Config` dataclass to be the same concern as the rest of the parsers. The parser.py file is meant to parse a config; for that reason, I think the `Config` dataclass should live in that file.

weather_dl/download_pipeline/partition.py
```diff
@@ -20,7 +20,8 @@
 import apache_beam as beam
 
 from .manifest import Manifest
-from .parsers import Config, prepare_target_name
+from .parsers import prepare_target_name
+from .config import Config
```
alxmrs, line 24: From looking at how `prepare_target_name` is called, I think it could be a method on the `Config` dataclass. What do you think?

weather_dl/download_pipeline/partition_test.py
```diff
@@ -118,21 +121,26 @@ def test_partition_multi_params_multi_key(self):
...
+                 'selection': {**config['selection'],
+                               **{'year': ['2015'], 'month': ['2']}}}),
+            Config.from_config(
+                {'parameters': dict(config['parameters'], api_key='KKKK3', api_url='UUUU3', subsection_name='deepmind'),
+                 'selection': {**config['selection'],
+                               **{'year': ['2016'], 'month': ['1']}}}),
+            Config.from_config(
+                {'parameters': dict(config['parameters'], api_key='KKKK1', api_url='UUUU1', subsection_name='research'),
+                 'selection': {**config['selection'],
+                               **{'year': ['2016'], 'month': ['2']}}}),
```
alxmrs, line 143: What about using a list comprehension here? Something like: 

```suggestion
            Config.from_config(it) for it in [
                {'parameters': dict(config['parameters'], api_key='KKKK1', api_url='UUUU1', subsection_name='research'),
                 'selection': {**config['selection'],
                               **{'year': ['2015'], 'month': ['1']}}},
                {'parameters': dict(config['parameters'], api_key='KKKK2', api_url='UUUU2', subsection_name='cloud'),
                 'selection': {**config['selection'],
                               **{'year': ['2015'], 'month': ['2']}}},
                {'parameters': dict(config['parameters'], api_key='KKKK3', api_url='UUUU3', subsection_name='deepmind'),
                 'selection': {**config['selection'],
                               **{'year': ['2016'], 'month': ['1']}}},
                {'parameters': dict(config['parameters'], api_key='KKKK1', api_url='UUUU1', subsection_name='research'),
                 'selection': {**config['selection'],
                               **{'year': ['2016'], 'month': ['2']}}},
            ]                   
```

alxmrs (changes requested):
Overall, the PR looks great! My main suggestion is to restructure the parser a bit to be more maintainable. Otherwise, I have a few nits.

weather_dl/download_pipeline/config.py
mahrsee1997: To avoid the problem of circular import, we will keep it in a separate file only.

weather_dl/download_pipeline/parsers.py
mahrsee1997: Yes, it makes sense. We should create dataclass object at the end of `process_config`.

weather_dl/download_pipeline/partition.py
mahrsee1997: It should be kept in `parsers.py` only as moving this in the `Config` dataclass will cause problem of circular import.

weather_dl/download_pipeline/fetcher.py
```diff
@@ -74,9 +75,9 @@ def fetch_data(self, config: t.Dict, *, worker_name: str = 'default') -> None:
 
         client = CLIENTS[self.client_name](config)
         target = prepare_target_name(config)
-        dataset = config['parameters'].get('dataset', '')
-        selection = config['selection']
-        user = config['parameters'].get('user_id', 'unknown')
+        dataset = config.dataset
+        selection = config.selection
+        user = config.user_id
```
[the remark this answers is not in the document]
mahrsee1997: Yes, it makes sense. Necessary code changes have been pushed.

weather_dl/download_pipeline/parsers.py
```diff
@@ -340,7 +330,7 @@ def require(condition: bool, message: str, error_type: t.Type[Exception] = Value
             """
             'parameters' section required in configuration file.
 
-            The 'parameters' section specifies the 'dataset', 'target_path', and
+            The 'parameters' section specifies the 'client', 'dataset', 'target_path', and
```
alxmrs, line 333: Nice catch.

weather_dl/download_pipeline/partition.py
alxmrs: ah, that makes sense. The status quo is fine, in that case.

alxmrs (approved):
LGTM!

PR #144, opened 2022-04-08
Changes to make 'target_path' and 'target_filename' compliant with Python's standard string formatting

### LINKED ISSUE #4: `target_path` templates should include be totally compatible with Python format strings
`target_path` (and `target_filename`) should be totally compliant with Python's standard string formatting. This includes being able to use named arguments (e.g. `'gs://bucket/{year}/{month}/{day}'`) as well as specifying formats for strings (e.g.`'gs://bucket/{year:04d}/{month:02d}/{day:02d}'`).

mahrsee1997:
Fixed #4 and #5.

### DIFF
--- a/Configuration.md
+++ b/Configuration.md
@@ -80,8 +80,6 @@ These describe which data source to download, where the data should live, and ho
     * Like `target_path`, `target_filename` can contain format symbols to be replaced by partition keys; if this is
       used, the total number of format symbols in both fields must match the number of partition keys.
     * This field is required when generating a date-based directory hierarchy (see below).
-* `append_date_dirs`: (optional) A boolean indicating whether a date-based directory hierarchy should be created (see
-  below); defaults to false if not used.
 * `partition_keys`: (optional) This determines how download jobs will be divided.
     * Value can be a single item or a list.
     * Each value must appear as a key in the `selection` section.
@@ -91,29 +89,24 @@ These describe which data source to download, where the data should live, and ho
         * E.g. `['year', 'month']` will lead to a config set like `[(2015, 01), (2015, 02), (2015, 03), ...]`.
     * The list of keys will be used to format the `target_path`.
 
+> **NOTE**: `target_path` and `target_filename` templates are totally compatible with Python's standard string formatting.
+> This includes being able to use named arguments (e.g. 'gs://bucket/{year}/{month}/{day}') as well as specifying formats for strings 
+> (e.g. 'gs://bucket/{year:04d}/{month:02d}/{day:02d}').
+
 ### Creating a date-based directory hierarchy
 
-The configuration can be set up to automatically generate a date-based directory hierarchy for the output files.
-
-To enable this feature, the `append_date_dirs` field has to be set to `true`. In addition, the `target_filename` needs
-to be specified, and `date` has to be a `partition_key`;
-`date` will not be used as a replacement in `target_template` but will instead be used to create a directory structure.
-
-The resulting target path will be `<target_path>/{year}/{month}/{day}<target_filename>`. The number of format symbols in
-this path has to match the number of partition keys excluding `date`.
+The date-based directory hierarchy can be created using Python's standard string formatting.
+Below are some examples of how to use `target_path` and `target_filename` with Python's standard string formatting.
 
 <details>
 <summary><strong>Examples</strong></summary>
-
-Below are more examples of how to use `target_path`, `target_filename`, and `append_date_dirs`.
 
 Note that any parameters that are not relevant to the target path have been omitted.
 
 ```
 [parameters]
 target_filename=.nc
-target_path=gs://ecmwf-output-test/era5/
-append_date_dirs=true
+target_path=gs://ecmwf-output-test/era5/{date:%%Y/%%m/%%d}
 partition_keys=
      date
 [selection]
@@ -126,9 +119,8 @@ will create
 
 ```
 [parameters]
-target_filename=-pressure-{}.nc
-target_path=gs://ecmwf-output-test/era5/
-append_date_dirs=true
+target_filename=-pressure-{pressure_level}.nc
+target_path=gs://ecmwf-output-test/era5/{date:%%Y/%%m/%%d}
 partition_keys=
      date
      pressure_level
@@ -145,8 +137,7 @@ will create
 ```
 [parameters]
 target_filename=.nc
-target_path=gs://ecmwf-output-test/pressure-{}/era5/
-append_date_dirs=true
+target_path=gs://ecmwf-output-test/pressure-{pressure_level}/era5/{date:%%Y/%%m/%%d}
 partition_keys=
      date
      pressure_level
@@ -160,12 +151,9 @@ will create
 `gs://ecmwf-output-test/pressure-500/era5/2017/01/01.nc` and  
 `gs://ecmwf-output-test/pressure-500/era5/2017/01/02.nc`.
 
-The above example also illustrates how to create a directory structure based on partition keys, even without using the
-date-based creation:
-
 ```
 [parameters]
-target_path=gs://ecmwf-output-test/era5/{}/{}/{}-pressure-{}.nc
+target_path=gs://ecmwf-output-test/era5/{year:04d}/{month:02d}/{day:02d}-pressure-{pressure_level}.nc
 partition_keys=
     year
     month
@@ -186,6 +174,14 @@ day=
 will create  
 `gs://ecmwf-output-test/era5/2017/01/01-pressure-500.nc` and  
 `gs://ecmwf-output-test/era5/2017/01/02-pressure-500.nc`.
+
+> **Note**: Replacing the `target_path` of the above example with this `target_path=gs://ecmwf-output-test/era5/{year}/{month}/{day}-pressure-
+>{pressure_level}.nc`
+>
+> will create
+>
+> `gs://ecmwf-output-test/era5/2017/1/1-pressure-500.nc` and  
+> `gs://ecmwf-output-test/era5/2017/1/2-pressure-500.nc`.
 
 </details>
 
--- a/Configuration.md
+++ b/Configuration.md
@@ -76,10 +76,6 @@ These describe which data source to download, where the data should live, and ho
 * `dataset`: (optional) Name of the target dataset. Allowed options are dictated by the client.
 * `target_path`: (required) Download artifact filename template. Can use Python string format symbols. Must have the
   same number of format symbols as the number of partition keys.
-* `target_filename`: (optional) This file name will be appended to `target_path`.
-    * Like `target_path`, `target_filename` can contain format symbols to be replaced by partition keys; if this is
-      used, the total number of format symbols in both fields must match the number of partition keys.
-    * This field is required when generating a date-based directory hierarchy (see below).
 * `partition_keys`: (optional) This determines how download jobs will be divided.
     * Value can be a single item or a list.
     * Each value must appear as a key in the `selection` section.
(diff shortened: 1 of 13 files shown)


weather_dl/download_pipeline/parsers.py
```diff
@@ -69,6 +69,86 @@ def date(candidate: str) -> datetime.date:
...
+        nonlocal converted
+        try:
+            if key == "day":
+                assert 1 <= candidate <= 31
+            if key == "month":
+                assert 1 <= candidate <= 12
+        except AssertionError:
+            converted = None
+
+    if (isinstance(candidate, str) or isinstance(candidate, int)):
```
alxmrs, line 123: No need to wrap the conditions in the if statement in parens.

weather_dl/download_pipeline/parsers.py
```diff
@@ -69,6 +69,86 @@ def date(candidate: str) -> datetime.date:
...
+
+def day_month_year(key: str, candidate: t.Any) -> int:
+    """Converts day, month and year strings into 'int'.
+
+    Also validates the day and month input.
+    """
+    converted = None
+
+    def validate(key: str, candidate: int) -> None:
+        nonlocal converted
```
alxmrs, line 114: I recommend against this side-effect approach.

weather_dl/download_pipeline/parsers.py
```diff
@@ -69,6 +69,86 @@ def date(candidate: str) -> datetime.date:
...
+
+
+def day_month_year(key: str, candidate: t.Any) -> int:
+    """Converts day, month and year strings into 'int'.
+
+    Also validates the day and month input.
+    """
+    converted = None
+
+    def validate(key: str, candidate: int) -> None:
```
alxmrs, line 113: This is a neat approach. However, I think it would be clearer / more succinct if you simply raised a `ValueError` if the candidate falls outside of the range.

weather_dl/download_pipeline/parsers.py
```diff
@@ -69,6 +69,86 @@ def date(candidate: str) -> datetime.date:
...
+        nonlocal converted
+        try:
+            if key == "day":
+                assert 1 <= candidate <= 31
+            if key == "month":
+                assert 1 <= candidate <= 12
```
alxmrs, line 119: Potentially, you could pass specific error message to the user with this assert message. If you want to present it as a top level `ValueError`, consider using the `raise from` syntax. 

https://docs.python.org/3/tutorial/errors.html#exception-chaining

alxmrs: In addition to adding an assert message, you'd need to refactor this function.

weather_dl/download_pipeline/parsers.py
```diff
@@ -69,6 +69,86 @@ def date(candidate: str) -> datetime.date:
...
+
+
+def typecast(key: str, value: t.Any) -> t.Any:
+    """typed the value to its appropriate datatype"""
+    if key == "date":
+        return date(value)
+    if key == "time":
+        return time(value)
+    if key == "day" or key == "month" or key == "year":
+        return day_month_year(key, value)
```
alxmrs, line 145: I see why you recommended using dictionary-based control flow in standup the other day. Now that I see it all line up, I agree that this would make a bit cleaner.

weather_dl/download_pipeline/parsers.py
```diff
@@ -69,6 +69,86 @@ def date(candidate: str) -> datetime.date:
...
+    if converted is None:
+        raise ValueError(
+            f"Not a valid day or month or year: {candidate}. Please use valid value."
+        )
+
+    return converted
+
+
+def typecast(key: str, value: t.Any) -> t.Any:
+    """typed the value to its appropriate datatype"""
```
alxmrs, line 139: "typed" should be capitalized, the line should end with a full stop `.`.

weather_dl/download_pipeline/parsers.py
```diff
@@ -303,12 +376,9 @@ def require(condition: bool, message: str) -> None:
...
             Supported clients are {}
             """.format(str(list(CLIENTS.keys()))))
-    if params.get('append_date_dirs', 'false') == 'true':
-        require(use_date_as_directory(config),
-                """
-                'append_date_dirs' set to true, but creating the date directory hierarchy also
-                requires that 'target_filename' is given and that 'date' is a partition_key.
-                """)
+    if 'append_date_dirs' in params:
+        raise NotImplementedError("We are no longer supporting \'append_date_dirs\'! Please refer to documentation for "
```
alxmrs, line 380: I don't think you need the escape strings `'`.

weather_dl/download_pipeline/parsers.py
```diff
@@ -350,22 +412,15 @@ def require(condition: bool, message: str) -> None:
...
-        date = t.cast(str, config['selection']['date'][0])
-        date_vals = date.split('-')
-        target_path = os.path.join(target_path, *date_vals)
-        partition_keys.remove('date')
-
     target_path += target_filename
 
-    partition_key_values = [config['selection'][key][0] for key in partition_keys]
-    target = target_path.format(*partition_key_values)
+    partition_key_values = OrderedDict((key, typecast(key, config['selection'][key][0])) for key in partition_keys)
```
alxmrs, line 422: Please rename this variable.

Configuration.md
```diff
@@ -91,29 +89,24 @@ These describe which data source to download, where the data should live, and ho
...
 <details>
 <summary><strong>Examples</strong></summary>
 
-Below are more examples of how to use `target_path`, `target_filename`, and `append_date_dirs`.
-
 Note that any parameters that are not relevant to the target path have been omitted.
 
 ```
 [parameters]
 target_filename=.nc
```
alxmrs, line 108: I think we should get rid of `target_filename` too. What purpose does it serve that `target_path` does not?

alxmrs: Also, related to #6, what is the best name for this field? `target_template`? Maybe I shouldn't overthink it and `target_path` is fine.

alxmrs: `partition_template`?

alxmrs: If we change the name, we should issue a deprecation warning to the user for now, and also accept it. 
Though, we shouldn't allow both names to be used at once.

alxmrs (changes requested):
Overall, this looks good! I've added a few suggestions. 

A top level issue I'm concerned with: How will readers of this blog post react when the config changes from what they're used to? What's the best way for us to handle this? 

https://towardsdatascience.com/how-to-use-ecmwf-weather-forecasts-in-machine-learning-models-92fc3730c5cf

weather_dl/download_pipeline/parsers.py
mahrsee1997: Refactored the function, please have a look and provide your feedback.

Configuration.md
mahrsee1997: Yes, we can get rid of `target_filename`. I have removed the support for the same and pushed the code.

Regarding renaming of `target_path`, how about ` target_path_template` ? Also should we change it now ?

mahrsee1997:
The blog should be considered as outdated.

And from the end user perspective we are giving appropriate error messages …

- Current version of weather-tool is no longer supporting 'append_date_dirs'! Please refer to documentation for creating date-based directory hierarchy (https://github.com/google/weather-tools/blob/main/Configuration.md#creating-a-date-based-directory-hierarchy).

- Current version of weather-tool is no longer supporting 'target_filename'! Please refer to documentation (https://github.com/google/weather-tools/blob/main/Configuration.md#parameters-section).

We can also display some kind of warning in the starting and ending of the pipeline informing users about the changes and put reference to our documentation.

weather_dl/download_pipeline/parsers.py
```diff
@@ -303,12 +381,16 @@ def require(condition: bool, message: str) -> None:
...
             Supported clients are {}
             """.format(str(list(CLIENTS.keys()))))
-    if params.get('append_date_dirs', 'false') == 'true':
-        require(use_date_as_directory(config),
-                """
-                'append_date_dirs' set to true, but creating the date directory hierarchy also
-                requires that 'target_filename' is given and that 'date' is a partition_key.
-                """)
+    if 'append_date_dirs' in params:
+        raise NotImplementedError("Current version of weather-tool is no longer supporting 'append_date_dirs'! "
```
alxmrs, line 385: Some copy editing: 
```
The current version of `google-weather-tools` no longer supports 'append_date_dirs'! Please refer to the documentation for creating date-based directory hierarchies (https://weather-tools.readthedocs.io/en/latest/Configuration.html#creating-a-date-based-directory-hierarchy)."
```

weather_dl/download_pipeline/parsers.py
```diff
@@ -303,12 +381,16 @@ def require(condition: bool, message: str) -> None:
...
+                                  "Please refer to documentation for creating date-based directory hierarchy "
+                                  "(https://github.com/google/weather-tools/blob/main/Configuration.md"
+                                  "#creating-a-date-based-directory-hierarchy).")
+    if 'target_filename' in params:
+        raise NotImplementedError("Current version of weather-tool is no longer supporting 'target_filename'! "
```
alxmrs, line 390: Please cross-apply similar copy editing to this error message.

weather_dl/download_pipeline/parsers.py
```diff
@@ -303,12 +381,16 @@ def require(condition: bool, message: str) -> None:
...
+                                  "Please refer to documentation "
+                                  "(https://github.com/google/weather-tools/blob/main/Configuration.md"
+                                  "#parameters-section).")
```
alxmrs, line 393: How about we make use of the `require` method so that we have consistent string formatting? I do think it's a good idea to use `NotImplementedError`s. Let's update `require` to take an optional argument that is the type of error (defaulting to ValueError).

alxmrs (changes requested):
Requesting feedback primarily for improvements to parsing types.

weather_dl/download_pipeline/parsers.py
```diff
@@ -69,6 +69,91 @@ def date(candidate: str) -> datetime.date:
...
+
+    if converted is None:
+        raise ValueError(
+            f"Not a valid {key} value: {candidate}. Please use valid value."
+        )
+
+    return converted
+
+
+def literal_eval(candidate: t.Any, key: t.Optional[str] = None) -> t.Any:
```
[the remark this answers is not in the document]
mahrsee1997: Changed the method name to `parse_literal`.

weather_dl/download_pipeline/parsers.py
```diff
@@ -69,6 +69,91 @@ def date(candidate: str) -> datetime.date:
...
+            converted = int(candidate)
+        except ValueError:
+            pass
+        if converted is not None:
+            validate(key, converted)
+
+    if converted is None:
+        raise ValueError(
+            f"Not a valid {key} value: {candidate}. Please use valid value."
+        )
```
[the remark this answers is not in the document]
mahrsee1997: Refactored it completely. LMKYT.

weather_dl/download_pipeline/parsers.py
```diff
@@ -69,6 +69,81 @@ def date(candidate: str) -> datetime.date:
...
+def day_month_year(candidate: t.Any) -> int:
+    """Converts day, month and year strings into 'int'."""
+    try:
+        if isinstance(candidate, str) or isinstance(candidate, int):
+            return int(candidate)
+        raise
+    except ValueError:
+        raise ValueError(
+            f"Not a valid day or month or year value: {candidate}. Please use valid value."
+        )
```
alxmrs, line 115:
```suggestion
    try:
        if isinstance(candidate, str) or isinstance(candidate, int):
            return int(candidate)
        raise
    except ValueError as e:
        raise ValueError(
            f"Not a valid day, month, or year value: {candidate}. Please use valid value."
        ) from e
```

alxmrs: small nits, but in general this looks good.

weather_dl/download_pipeline/parsers.py
```diff
@@ -262,10 +330,10 @@ def process_config(file: t.IO) -> Config:
     """Read the config file and prompt the user if it is improperly structured."""
     config = parse_config(file)
 
-    def require(condition: bool, message: str) -> None:
-        """A assert-like helper that wraps text and throws a `ValueError`."""
+    def require(condition: bool, message: str, error_type: t.Any = ValueError) -> None:
```
alxmrs, line 333: nit: can we have a more specific type for `error_type`? Will `Error` work here? Maybe you need `t.Type` [[1]].

[1]: https://docs.python.org/3/library/typing.html#typing.Type

weather_dl/download_pipeline/parsers.py
```diff
@@ -303,12 +371,21 @@ def require(condition: bool, message: str) -> None:
...
-                'append_date_dirs' set to true, but creating the date directory hierarchy also
-                requires that 'target_filename' is given and that 'date' is a partition_key.
-                """)
+    require('append_date_dirs' not in params,
+            """
+            The current version of 'google-weather-tools' no longer supports 'append_date_dirs'!
+
+            Please refer to documentation for creating date-based directory hierarchy :
+            https://github.com/google/weather-tools/blob/main/Configuration.md#"""
+            """creating-a-date-based-directory-hierarchy.""",
```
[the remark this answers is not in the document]
alxmrs: https://weather-tools.readthedocs.io/en/latest/Configuration.html#creating-a-date-based-directory-hierarchy

alxmrs (approved):
LGTM. Please fix the last few small nits before merging.

alxmrs:
Before merging, please make sure the tests are fixed. It looks like some tests for the fetcher need to be updated.

mahrsee1997:
Thanks for the review @alxmrs !

alxmrs:
Thanks for the final adjustments! Merging...

PR #147, opened 2022-04-16
Netcdf splits

uhager:
weather-sp: enable more flexible splitting.

Allows splitting by any dimension, though explicitly rejecting latitude and longitude to avoid accidentally creating large numbers of tiny files.

### DIFF
--- a/weather_sp/splitter_pipeline/file_splitters.py
+++ b/weather_sp/splitter_pipeline/file_splitters.py
@@ -166,7 +166,7 @@ class NetCdfSplitter(FileSplitter):
         with self._copy_to_local_file() as local_file:
             yield xr.open_dataset(local_file.name, engine='netcdf4')
 
-    def _write_dataset(self, dataset: xr.Dataset(), split_dims: t.List[str]) -> None:
+    def _write_dataset(self, dataset: xr.Dataset, split_dims: t.List[str]) -> None:
         with FileSystems().create(self._get_output_for_dataset(dataset, split_dims)) as dest_file:
             dest_file.write(dataset.to_netcdf())
 
--- a/weather_sp/splitter_pipeline/file_splitters.py
+++ b/weather_sp/splitter_pipeline/file_splitters.py
@@ -165,7 +165,7 @@ class NetCdfSplitter(FileSplitter):
     @contextmanager
     def _open_dataset_locally(self) -> t.Iterator[xr.Dataset]:
         with self._copy_to_local_file() as local_file:
-            yield xr.open_dataset(local_file.name)
+            yield xr.open_dataset(local_file.name, engine='netcdf4')
 
     def _write_dataset(self, dataset: xr.Dataset(), split_dims: t.List[str]) -> None:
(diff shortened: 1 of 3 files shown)


weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -147,37 +148,35 @@ def split_data(self) -> None:
                                   for var in dataset.data_vars])
             else:
                 iterlists.append([dataset])
-            for dim in ('time', 'level'):
-                if dim in self.output_info.split_dims():
-                    iterlists.append(dataset[dim])
+            filtered_split_dims = [
+                x for x in self.output_info.split_dims() if x not in ('variable', self._UNSUPPORTED_DIMENSIONS)]
+            for dim in filtered_split_dims:
+                iterlists.append(dataset[dim])
```
alxmrs, line 154: I think it would be clearer to have this be a single for loop with an if statement. Something like: 
```suggestion
            filtered_split_dims = [
            for dim in self.output_info.split_dims():
                if dim not in ('variable', self._UNSUPPORTED_DIMENSIONS):
                    iterlists.append(dataset[dim])
```

uhager: I reused the filtered_split_dims to create the output file though. So I'd have to do similar filtering (or at least check for 'variable' at that point again.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -147,37 +148,35 @@ def split_data(self) -> None:
...
                 for da in comb[1:]:
-                    if 'time' in da.coords:
-                        selected = selected.sel(time=da.time)
-                    if 'level' in da.coords:
-                        selected = selected.sel(level=da.level)
-                self._write_dataset(selected)
-            self.logger.info('split %s into %d files',
-                             self.input_path, len(list(combinations)))
+                    for dim in da.coords:
+                        selected = selected.sel({dim: getattr(da, dim)})
```
alxmrs, line 160: Really slick.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -147,37 +148,35 @@ def split_data(self) -> None:
...
 
     @contextmanager
     def _open_dataset_locally(self) -> t.Iterator[xr.Dataset]:
         with self._copy_to_local_file() as local_file:
-            yield xr.open_dataset(local_file.name)
+            yield xr.open_dataset(local_file.name, engine='netcdf4')
 
-    def _write_dataset(self, dataset: xr.Dataset()) -> None:
-        with FileSystems().create(self._get_output_for_dataset(dataset)) as dest_file:
+    def _write_dataset(self, dataset: xr.Dataset(), split_dims: t.List[str]) -> None:
```
alxmrs, line 169: Missed it last time: the type should be `xr.Dataset` I believe. Currently, the type is an instance of a dataset, not the type.

uhager: oh wow! Nice catch! I had to read this three times to even notice what you meant!

alxmrs (approved):
Great work! A nice, clean implementation.

PR #148, opened 2022-04-20
Restructured the fetch stage to be a Composite Beam transform and separated the request & download stage of fetching

mahrsee1997:


weather_dl/download_pipeline/clients.py
```diff
@@ -102,6 +114,12 @@ def __init__(self, config: Config, level: int = logging.INFO) -> None:
     def retrieve(self, dataset: str, selection: t.Dict, target: str) -> None:
         self.c.retrieve(dataset, selection, target)
 
+    def fetch(self, dataset: str, selection: t.Dict) -> None:
+        pass
+
+    def download(self, dataset: str, result: t.Dict, output: str) -> None:
+        pass
```
alxmrs, line 121: These should raise a `NotImplementedError`.

weather_dl/download_pipeline/clients.py
```diff
@@ -152,6 +170,90 @@ def __exit__(self, exc_type, exc_value, traceback):
         self._redirector.__exit__(exc_type, exc_value, traceback)
 
 
+class APIRequestExtended(api.APIRequest):
```
alxmrs, line 173: How about we include "MARS" in the name of this class?

weather_dl/download_pipeline/clients.py
```diff
@@ -152,6 +170,90 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
alxmrs, line 175: You can omit the constructor if it just calls super.

weather_dl/download_pipeline/clients.py
```diff
@@ -152,6 +170,90 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            while size != result["size"] and tries < 10:
+                size = self._transfer(
+                    urljoin(self.url, result["href"]), target, result["size"]
+                )
+                if size != result["size"] and tries < 10:
+                    tries += 1
+                    self.log("Transfer interrupted, resuming in 60s...")
+                    time.sleep(60)
+                else:
+                    break
```
alxmrs, line 219: Ideally, we could update or replace this block with code that is capable of downloading more than 1 MB of data at a time. It'd be nice if we could do something like this, for example:

https://github.com/google/weather-tools/blob/4aa98198ab81d895926d08fb6fd70bc25431d03d/weather_mv/loader_pipeline/sinks.py#L90

weather_dl/download_pipeline/clients.py
```diff
@@ -152,6 +170,90 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    def download(self, result, target=None):
+        if target:
+            if os.path.exists(target):
+                # Empty the target file, if it already exists, otherwise the
+                # transfer below might be fooled into thinking we're resuming
+                # an interrupted download.
+                open(target, "w").close()
+
+            size = -1
+            tries = 0
```
alxmrs, line 209: Following up with the comment below: We could use Beam's retry logic with exponential backoff (via the decorator) instead of re-using ECMWF's implementation. WDYT?

weather_dl/download_pipeline/clients.py
```diff
@@ -152,6 +170,90 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+        self.connection.cleanup()
+
+        return result
+
+
+class ECMWFServiceExtended(ECMWFService):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def fetch(self, req):
```
alxmrs, line 232: If it's not too much trouble: Could you add types to these two function signatures?

weather_dl/download_pipeline/clients.py
```diff
@@ -152,6 +170,90 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+            assert size == result["size"]
+
+        self.connection.cleanup()
+
+        return result
+
+
+class ECMWFServiceExtended(ECMWFService):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
alxmrs, line 230: Again, this can be omitted.

weather_dl/download_pipeline/clients.py
alxmrs: Naming discussion. I kind of like "extended", but is there a better name? What about "Adapter"? Any other ideas?

alxmrs: On second thought, adding "MARS" to the name of the class may not be necessary.

weather_dl/download_pipeline/clients.py
```diff
@@ -152,6 +170,90 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    def fetch(self, req):
+        c = APIRequestExtended(
+            self.url,
+            "services/%s" % (self.service,),
+            email=self.email,
+            key=self.key,
+            log=self.log,
+            verbose=self.verbose,
+            quiet=self.quiet,
+        )
```
alxmrs, line 241: I could see this becoming a class variable and being defined in the constructor, since it's repeated.

weather_dl/download_pipeline/clients.py
alxmrs: May be needed, see comment below.

weather_dl/download_pipeline/clients.py
alxmrs: Consider: creating type aliases for the return value (a named alias to `t.Dict` could be helpful for using types to document the feature).

weather_dl/download_pipeline/fetcher.py
```diff
@@ -44,29 +44,122 @@ class Fetcher(beam.DoFn):
         client_name: The name of the download client to construct per each request.
         manifest: A manifest to keep track of the status of requests
         store: To manage where downloads are persisted.
+        optimise_download: Optimised the download.
```
alxmrs, line 47: Please explain in these docs how the download is optimized.

alxmrs: Related: consider using a more specific name for the flag.

weather_dl/download_pipeline/fetcher.py
```diff
@@ -44,29 +44,122 @@ class Fetcher(beam.DoFn):
...
+class EachPartition(beam.DoFn):
+    def process(self, element) -> t.Iterator[t.Tuple[t.Any, str]]:
+        # element: Tuple[Tuple[str, int], Iterator[Config]]
+        """Execute download requests one-by-one."""
+        (subsection, request_idx), partitions = element
+        worker_name = f'{subsection}.{request_idx}'
+        logger.info(f"[{worker_name}] Starting requests...")
+        for partition in partitions:
+            beam.metrics.Metrics.counter('Fetcher', subsection).inc()
+            yield (partition, worker_name)
```
alxmrs, line 91: Nice breakdown of the pipeline into component parts.

weather_dl/download_pipeline/fetcher.py
```diff
@@ -44,29 +44,122 @@ class Fetcher(beam.DoFn):
...
+                request
+                | 'Fetch+Download' >> beam.ParDo(Retrieve(self.client_name, self.manifest, self.store))
+            )
+        (
+            downloaded_data
+            | 'Upload' >> beam.ParDo(Upload(self.client_name, self.manifest, self.store))
+        )
+
+
+class EachPartition(beam.DoFn):
```
alxmrs, line 82: For each of these DoFns, please add class docstrings (I assume you would anyway at a more advanced draft).

weather_dl/download_pipeline/fetcher.py
```diff
@@ -44,29 +44,122 @@ class Fetcher(beam.DoFn):
...
+        for partition in partitions:
+            beam.metrics.Metrics.counter('Fetcher', subsection).inc()
+            yield (partition, worker_name)
+
+
+@dataclasses.dataclass
+class FetchData(beam.DoFn):
+    client_name: str
+    manifest: Manifest = NoOpManifest(Location('noop://in-memory'))
+    store: t.Optional[Store] = None
```
alxmrs, line 98: Consider: using a post-init to create a default store.

weather_dl/download_pipeline/fetcher.py
```diff
@@ -44,29 +44,122 @@ class Fetcher(beam.DoFn):
...
+        client = CLIENTS[self.client_name](config)
+
+        with self.manifest.transact(config.selection, target, config.user_id):
+            logger.info(f'[{worker_name}] Downloading data for {target!r}.')
+            self.download(client, config.dataset, result, temp_name)
+            yield (config, worker_name, temp_name, target)
+
+
+@dataclasses.dataclass
+class Retrieve(beam.DoFn):
```
alxmrs, line 149: Nit: maybe it should be `RetrieveData` to follow the naming conventions of the other classes?

weather_dl/download_pipeline/fetcher.py
```diff
@@ -78,22 +171,30 @@ def fetch_data(self, config: Config, *, worker_name: str = 'default') -> None:
 
         with self.manifest.transact(config.selection, target, config.user_id):
             with tempfile.NamedTemporaryFile() as temp:
-                logger.info(f'[{worker_name}] Fetching data for {target!r}.')
+                logger.info(f'[{worker_name}] Fetching and Downloading data for {target!r}.')
```
alxmrs, line 174: petty nit: can you use an `&` instead of "and" here :) ?

weather_dl/download_pipeline/fetcher.py
```diff
@@ -44,29 +44,122 @@ class Fetcher(beam.DoFn):
...
+            return
+
+        client = CLIENTS[self.client_name](config)
+        target = prepare_target_name(config)
+
+        with self.manifest.transact(config.selection, target, config.user_id):
+            with tempfile.NamedTemporaryFile() as temp:
+                logger.info(f'[{worker_name}] Fetching data for {target!r}.')
+                result = self.fetch(client, config.dataset, config.selection)
+                yield (result, config, worker_name, temp.name, target)
```
alxmrs, line 122: Consider creating an internal data class for passing along this result. 

Also: Are you sure that using the `temp.name` is safe? It's possible that the temporary file will disappear. 

Furthermore, looking at the code: do you need to create a temporary file here? If you don't need it for the fetch, it's probably safer to move this to the next stage.

weather_dl/download_pipeline/fetcher.py
```diff
@@ -44,29 +44,122 @@ class Fetcher(beam.DoFn):
...
+        if not config:
+            return
+
+        if skip_partition(config, self.store):
+            return
+
+        client = CLIENTS[self.client_name](config)
+        target = prepare_target_name(config)
+
+        with self.manifest.transact(config.selection, target, config.user_id):
```
alxmrs, line 118: A complication that I didn't anticipate until now: It would probably be best if we updated the manifest to distinguish between retrieved and downloaded. WDYT?

weather_dl/download_pipeline/fetcher.py
alxmrs: Maybe we don't need the dataclass since we can probably simplify what's returned. For example, We don't need to pass the target since that can be derived from the config. So, I see the tuple consisting of three parts: `config, worker_name, result`.

alxmrs: (you can choose your favorite order for these).

weather_dl/download_pipeline/fetcher.py
```diff
@@ -78,22 +171,30 @@ def fetch_data(self, config: Config, *, worker_name: str = 'default') -> None:
...
+class Upload(beam.DoFn):
+    client_name: str
+    manifest: Manifest = NoOpManifest(Location('noop://in-memory'))
+    store: t.Optional[Store] = None
 
-                logger.info(f'[{worker_name}] Upload to store complete for {target!r}.')
+    @retry_with_exponential_backoff
+    def upload(self, src: str, dest: str) -> None:
+        """Upload blob to cloud storage, with retries."""
+        with io.FileIO(src, 'rb') as src_:
```
alxmrs, line 188: can this be `with open(src, 'rb')`?

weather_dl/download_pipeline/pipeline.py
```diff
@@ -129,6 +130,8 @@ def run(argv: t.List[str], save_main_session: bool = True) -> PipelineArgs:
                         help='Number of concurrent requests to make per API key. '
                              'Default: make an educated guess per client & config. '
                              'Please see the client documentation for more details.')
+    parser.add_argument('-o', '--optimise-download', action='store_true', default=False,
+                        help="Optimised the downloads.")
```
alxmrs, line 134: Make sure you cross apply the description of what this does here.

alxmrs (commented):
Overall, the pipeline improvements look good! Here's my early feedback for the current draft.

weather_dl/download_pipeline/clients.py
mahrsee1997: Named `SplitMARSRequest`.

weather_dl/download_pipeline/clients.py
mahrsee1997: Yes, necessary code changes have been done.

weather_dl/download_pipeline/clients.py
mahrsee1997: Yes, necessary code changes have been done.

weather_dl/download_pipeline/fetcher.py
mahrsee1997: Changed it to `async_downloads`.

weather_dl/download_pipeline/fetcher.py
mahrsee1997: Done.

weather_dl/download_pipeline/fetcher.py
mahrsee1997: Made necessary changes. Added a new class variable in DownloadStatus named stage which represents the current stage of the request.

PR #155, opened 2022-05-04
Validates non-compatible regions scenarios in weather_mv tool.

### LINKED ISSUE #126: Alert the user earlier that the BigQuery table and cloud bucket need to be in the same region.
As a user, I'd like to know if the remote configuration for using this tool is incorrect immediately after invocation (instead of hours later). 

```
RuntimeError: BigQuery job beam_bq_job_LOAD_AUTOMATIC_JOB_NAME_LOAD_STEP_560_0b1bd5a3b84b07378d10258f5d1c31d8_dcaa42e715c949b8a5b8e762480d0b44 failed. Error Result: <ErrorProto message: 'Cannot read and write in different locations: source: US, destination: us-west1' reason: 'invalid'>
```

Implementation notes: When the user makes use of resources in different regions, they should be presented with a warning or an error message.

mahrsee1997:
Fixes #126 .

### DIFF
--- a/weather_mv/loader_pipeline/pipeline.py
+++ b/weather_mv/loader_pipeline/pipeline.py
@@ -54,8 +54,7 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
 def validate_region(output_table: str, temp_location: t.Optional[str] = None, region: t.Optional[str] = None) -> None:
     """Validates non-compatible regions scenarios by performing sanity check."""
 
-    def __cleanup(bigquery_client: bigquery.Client, storage_client: storage.Client,
-                     canary_output_table: str) -> None:
+    def __cleanup(bigquery_client: bigquery.Client, storage_client: storage.Client, canary_output_table: str) -> None:
         bigquery_client.delete_table(canary_output_table, not_found_ok=True)
         try:
             storage_client.get_bucket(CANARY_BUCKET_NAME).delete(force=True)
--- a/weather_mv/loader_pipeline/pipeline.py
+++ b/weather_mv/loader_pipeline/pipeline.py
@@ -81,7 +81,10 @@ def validate_region(output_table: str, temp_location: t.Optional[str] = None, re
     # Doing cleanup if operation get cut off midway.
     # TODO : Should we handle some other signals ?
     do_cleanup = partial(_cleanup, bigquery_client, storage_client, canary_output_table, canay_bucket_name)
+    original_sigint_handler = signal.getsignal(signal.SIGINT)
+    original_sigtstp_handler = signal.getsignal(signal.SIGTSTP)
     signal.signal(signal.SIGINT, do_cleanup)
+    signal.signal(signal.SIGTSTP, do_cleanup)
 
     bucket_region = region
     table_region = None
@@ -114,6 +117,8 @@ def validate_region(output_table: str, temp_location: t.Optional[str] = None, re
         raise RuntimeError(f'Can\'t migrate from source: {bucket_region} to destination: {table_region}')
     finally:
         _cleanup(bigquery_client, storage_client, canary_output_table, canay_bucket_name)
+        signal.signal(signal.SIGINT, original_sigint_handler)
+        signal.signal(signal.SIGINT, original_sigtstp_handler)
 
 
 def pipeline(known_args: argparse.Namespace, pipeline_args: t.List[str]) -> None:
@@ -126,7 +131,7 @@ def pipeline(known_args: argparse.Namespace, pipeline_args: t.List[str]) -> None
 
     if not (known_args.dry_run or known_args.skip_region_validation):
         # Program execution will terminate on failure of region validation.
-        logger.info('Validating regions for data migration. This might take few seconds.')
+        logger.info('Validating regions for data migration. This might take a few seconds...')
         validate_region(known_args.output_table, temp_location=pipeline_options_dict.get('temp_location'),
                         region=pipeline_options_dict.get('region'))
         logger.info('Region validation completed successfully.')
--- a/weather_mv/loader_pipeline/pipeline.py
+++ b/weather_mv/loader_pipeline/pipeline.py
@@ -19,6 +19,11 @@ import json
 import logging
 import tempfile
 import typing as t
+import signal
+import sys
+import uuid
+import traceback
+from functools import partial
 
 import apache_beam as beam
 from apache_beam.io.filesystems import FileSystems
@@ -51,21 +56,32 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
         yield from [x.path for x in match.metadata_list]
 
 
+def _cleanup(bigquery_client: bigquery.Client, storage_client: storage.Client, canary_output_table: str,
+             canay_bucket_name: str, sig: t.Optional[t.Any] = None, frame: t.Optional[t.Any] = None) -> None:
+    bigquery_client.delete_table(canary_output_table, not_found_ok=True)
+    try:
+        storage_client.get_bucket(canay_bucket_name).delete(force=True)
+    except NotFound:
+        pass
+    if sig:
+        traceback.print_stack(frame)
+        sys.exit(0)
+
+
 def validate_region(output_table: str, temp_location: t.Optional[str] = None, region: t.Optional[str] = None) -> None:
     """Validates non-compatible regions scenarios by performing sanity check."""
-
-    def __cleanup(bigquery_client: bigquery.Client, storage_client: storage.Client, canary_output_table: str) -> None:
-        bigquery_client.delete_table(canary_output_table, not_found_ok=True)
-        try:
-            storage_client.get_bucket(CANARY_BUCKET_NAME).delete(force=True)
-        except NotFound:
-            pass
-
     if not region and not temp_location:
         raise ValueError('Invalid GCS location: None.')
 
     bigquery_client = bigquery.Client()
     storage_client = storage.Client()
+    canary_output_table = output_table + CANARY_OUTPUT_TABLE_SUFFIX + str(uuid.uuid4())
+    canay_bucket_name = CANARY_BUCKET_NAME + str(uuid.uuid4())
+
+    # Doing cleanup if operation get cut off midway.
+    # TODO : Should we handle some other signals ?
+    do_cleanup = partial(_cleanup, bigquery_client, storage_client, canary_output_table, canay_bucket_name)
+    signal.signal(signal.SIGINT, do_cleanup)
 
     bucket_region = region
     table_region = None
@@ -78,10 +94,7 @@ def validate_region(output_table: str, temp_location: t.Optional[str] = None, re
         bucket_region = storage_client.get_bucket(bucket_name).location
 
     try:
-        canary_output_table = output_table + CANARY_OUTPUT_TABLE_SUFFIX
-        __cleanup(bigquery_client, storage_client, canary_output_table)
-
-        bucket = storage_client.create_bucket(CANARY_BUCKET_NAME, location=bucket_region)
+        bucket = storage_client.create_bucket(canay_bucket_name, location=bucket_region)
         with tempfile.NamedTemporaryFile(mode='w+') as temp:
             json.dump(CANARY_RECORD, temp)
             temp.flush()
@@ -93,14 +106,14 @@ def validate_region(output_table: str, temp_location: t.Optional[str] = None, re
         table_region = table.location
 
         load_job = bigquery_client.load_table_from_uri(
-            f'gs://{CANARY_BUCKET_NAME}/{CANARY_RECORD_FILE_NAME}',
+            f'gs://{canay_bucket_name}/{CANARY_RECORD_FILE_NAME}',
             canary_output_table,
         )
         load_job.result()
     except BadRequest:
         raise RuntimeError(f'Can\'t migrate from source: {bucket_region} to destination: {table_region}')
     finally:
-        __cleanup(bigquery_client, storage_client, canary_output_table)
+        _cleanup(bigquery_client, storage_client, canary_output_table, canay_bucket_name)
 
 
 def pipeline(known_args: argparse.Namespace, pipeline_args: t.List[str]) -> None:
--- a/weather_mv/loader_pipeline/pipeline.py
+++ b/weather_mv/loader_pipeline/pipeline.py
@@ -25,9 +25,16 @@ from apache_beam.io.filesystems import FileSystems
 from apache_beam.options.pipeline_options import PipelineOptions, SetupOptions
 from google.cloud import bigquery, storage
 from google.api_core.exceptions import BadRequest, NotFound
+from urllib.parse import urlparse
 
 from .bq import ToBigQuery
 from .streaming import GroupMessagesByFixedWindows, ParsePaths
+
+CANARY_BUCKET_NAME = 'anthromet_dummy_bucket'
+CANARY_RECORD = {'foo': 'bar'}
+CANARY_RECORD_FILE_NAME = 'dummy_record.json'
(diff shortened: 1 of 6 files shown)


weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +44,75 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
+def validate_region(output_table: str, temp_location: t.Optional[str] = None, region: t.Optional[str] = None) -> None:
+    """Validates non-compatible regions scenarios by performing sanity check."""
+
+    def do_cleanup(bigquery_client: bigquery.Client, storage_client: storage.Client, dummy_output_table: str,
+                   dummy_bucket_name: str) -> None:
+        bigquery_client.delete_table(dummy_output_table, not_found_ok=True)
+        try:
+            storage_client.get_bucket(dummy_bucket_name).delete(force=True)
+        except NotFound:
+            pass
```
alxmrs, line 56: I like the pattern you have here. And, I'm glad that you're cleaning up resources. I think that a context manager would be a great form factor to perform this work, instead of a try..except.finally block. This seems to be a better fit if you intend to create a separate function, which I agree is a bit cleaner. 

https://docs.python.org/3/library/contextlib.html

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +44,75 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
+
+    bucket_region = region
+    table_region = None
+
+    if temp_location:
+        bucket = temp_location.split('gs://')
+        if len(bucket) < 2:
+            raise ValueError(f'Invalid GCS location: \'{bucket[0]}\'.')
+        bucket_name = bucket[1].split('/')[0]
+        bucket_region = storage_client.get_bucket(bucket_name).location
```
alxmrs, line 72: I recommend using urlparse: https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse

It's a bit safer to use than re-implementing a parser, since it probably handles corner cases that we can't anticipate.

alxmrs: Further, I expect it to throw better error messages.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +44,75 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
+
+    try:
+        DUMMY_BUCKET_NAME = 'anthromet_dummy_bucket'
+        DUMMY_RECORD = {'foo': 'bar'}
+        DUMMY_OUTPUT_TABLE = output_table + '_dummy_table'
+        DUMMY_TABLE_SCHEMA = [bigquery.SchemaField('name', 'STRING', mode='NULLABLE')]
```
alxmrs, line 78: If these are truly constants, then they should be moved to the top of the file.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +44,75 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
+        dummy_table = bigquery_client.create_table(dummy_table, exists_ok=True)
+        table_region = dummy_table.location
+
+        load_job = bigquery_client.load_table_from_uri(
+            f'gs://{DUMMY_BUCKET_NAME}/dummy_record.json',
+            DUMMY_OUTPUT_TABLE,
+        )
+        load_job.result()
+    except BadRequest:
+        raise RuntimeError(f'Can\'t migrate from source: {bucket_region} to destination: {table_region}')
```
alxmrs, line 99: Please `raise RuntimeError(...) from e`.

alxmrs: It'd be nice to present the whole stack trace to the user, for debugging purposes.

weather_mv/loader_pipeline/pipeline.py
alxmrs: I see now that how this is called. If cleanup is required before the test, then a context manager is probably not a good idea. In this case, I think it'd still be better to make this a private function at the top level.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +44,75 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
+        bucket_name = bucket[1].split('/')[0]
+        bucket_region = storage_client.get_bucket(bucket_name).location
+
+    try:
+        DUMMY_BUCKET_NAME = 'anthromet_dummy_bucket'
+        DUMMY_RECORD = {'foo': 'bar'}
+        DUMMY_OUTPUT_TABLE = output_table + '_dummy_table'
+        DUMMY_TABLE_SCHEMA = [bigquery.SchemaField('name', 'STRING', mode='NULLABLE')]
+
+        do_cleanup(bigquery_client, storage_client, DUMMY_OUTPUT_TABLE, DUMMY_BUCKET_NAME)
```
alxmrs, line 80: Why is cleanup done beforehand?

weather_mv/loader_pipeline/pipeline.py
alxmrs: Also, I don't think "DUMMY" is a good prefix. These usually indicate variables for unit testing. Here, these are actually live resources that we exercise! What do you think a better name would be? "Canary" comes to mind for me, e.g. `CANARY_BUCKET_NAME`.

weather_mv/loader_pipeline/pipeline.py
alxmrs: Is it possible for the test operations to work on an existing table (like, via an append) or via an overwrite? 

I see how if the operation get cut off midway, so cleanup might be desirable.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +44,75 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
     if not all_uris:
         raise FileNotFoundError(f"File prefix '{known_args.uris}' matched no objects")
 
     pipeline_options = PipelineOptions(pipeline_args)
+    pipeline_options_dict = pipeline_options.get_all_options()
+
+    if not known_args.dry_run:
+        # Program execution will terminate on failure of region validation.
+        validate_region(known_args.output_table, temp_location=pipeline_options_dict.get('temp_location'),
+                        region=pipeline_options_dict.get('region'))
```
alxmrs, line 115: How long does this pipeline take to run? Can you add a log statement here so users know that validation is happening?

alxmrs (commented):
Thanks for adding this validation! I have some implementation notes. I just want to sanity check: you've tested this e2e? I assume so – and in that case, it's really great that this works!

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: Necessary code changes have been done.

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: Thanks for pointing this out, moved the constants at the top of the file and also renamed to "Canary".

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: It takes around ~30 sec in case of success and ~15 sec in case of failure. Logs have been added.
Also as discussed, a flag named `skip-region-validation` has been introduced to skip this validation check if the user wants.

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: Not required as we are already getting whole stack trace.

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: Done.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -17,15 +17,25 @@
...
+from urllib.parse import urlparse
 
 from .bq import ToBigQuery
 from .streaming import GroupMessagesByFixedWindows, ParsePaths
 
+CANARY_BUCKET_NAME = 'anthromet_dummy_bucket'
+CANARY_RECORD = {'foo': 'bar'}
+CANARY_RECORD_FILE_NAME = 'dummy_record.json'
+CANARY_OUTPUT_TABLE_SUFFIX = '_anthromet_dummy_table'
+CANARY_TABLE_SCHEMA = [bigquery.SchemaField('name', 'STRING', mode='NULLABLE')]
```
alxmrs, line 37: Please update the term "dummy" in the strings, too.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +51,73 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
+
+    if temp_location:
+        parsed_temp_location = urlparse(temp_location)
+        if parsed_temp_location.scheme != 'gs' or parsed_temp_location.netloc == '':
+            raise ValueError(f'Invalid GCS location: {temp_location!r}.')
+        bucket_name = parsed_temp_location.netloc
+        bucket_region = storage_client.get_bucket(bucket_name).location
+
+    try:
+        CANARY_OUTPUT_TABLE = output_table + CANARY_OUTPUT_TABLE_SUFFIX
```
alxmrs, line 82: `CANARY_OUTPUT_TABLE` is a local and dynamic variable. Thus, it should not be treated as a constant, and should be in snake case.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +51,73 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
         yield from [x.path for x in match.metadata_list]
 
 
+def validate_region(output_table: str, temp_location: t.Optional[str] = None, region: t.Optional[str] = None) -> None:
+    """Validates non-compatible regions scenarios by performing sanity check."""
+
+    def __do_cleanup(bigquery_client: bigquery.Client, storage_client: storage.Client,
```
alxmrs, line 57: nit: can we change this name to `cleanup` or `_cleanup`?

alxmrs (commented):
Just a few nits! I'll TAL at the dependency error now.

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: Done.

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: Done.

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: Changed to `__cleanup`.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +56,80 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
+
+    bigquery_client = bigquery.Client()
+    storage_client = storage.Client()
+    canary_output_table = output_table + CANARY_OUTPUT_TABLE_SUFFIX + str(uuid.uuid4())
+    canay_bucket_name = CANARY_BUCKET_NAME + str(uuid.uuid4())
+
+    # Doing cleanup if operation get cut off midway.
+    # TODO : Should we handle some other signals ?
+    do_cleanup = partial(_cleanup, bigquery_client, storage_client, canary_output_table, canay_bucket_name)
+    signal.signal(signal.SIGINT, do_cleanup)
```
alxmrs, line 84: I'm surprised that a `try..except. KeyboardInterrupt` couldn't handle this case: Have you tried that approach?

alxmrs: Does this implementation have other advantages? For example, is this processed async?

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +56,80 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
+
+    bucket_region = region
+    table_region = None
+
+    if temp_location:
+        parsed_temp_location = urlparse(temp_location)
+        if parsed_temp_location.scheme != 'gs' or parsed_temp_location.netloc == '':
+            raise ValueError(f'Invalid GCS location: {temp_location!r}.')
+        bucket_name = parsed_temp_location.netloc
+        bucket_region = storage_client.get_bucket(bucket_name).location
```
alxmrs, line 94: Nice, this looks good.

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +56,80 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
     all_uris = list(pattern_to_uris(known_args.uris))
     if not all_uris:
         raise FileNotFoundError(f"File prefix '{known_args.uris}' matched no objects")
 
     pipeline_options = PipelineOptions(pipeline_args)
+    pipeline_options_dict = pipeline_options.get_all_options()
+
+    if not (known_args.dry_run or known_args.skip_region_validation):
+        # Program execution will terminate on failure of region validation.
+        logger.info('Validating regions for data migration. This might take few seconds.')
```
alxmrs, line 129:
```suggestion
        logger.info('Validating regions for data migration. This might take a few seconds...')
```

alxmrs (commented):
I'll review this tomorrow, but overall this PR looks good. Just one nit involving signals: I think this approach may be more complicated than catching a KeyboardInterrupt.

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: Done,

mahrsee1997: Done.

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: > When a user (not you, but others...) wants to interrupt a task he generally hits multiple times Ctrl-C. The first one will branch execution in the finally block. If another Ctrl-C occurs in the middle of the finally block because it contains slow operations like closing files, a new KeyboardInterrupt will be raised and nothing guarantees that the whole block will be executed.

SO : https://stackoverflow.com/questions/41280318/will-python-execute-finally-block-after-receiving-ctrlc

weather_mv/loader_pipeline/pipeline.py
```diff
@@ -41,12 +56,80 @@ def pattern_to_uris(match_pattern: str) -> t.Iterable[str]:
...
 
 
+def _cleanup(bigquery_client: bigquery.Client, storage_client: storage.Client, canary_output_table: str,
+             canay_bucket_name: str, sig: t.Optional[t.Any] = None, frame: t.Optional[t.Any] = None) -> None:
+    bigquery_client.delete_table(canary_output_table, not_found_ok=True)
+    try:
+        storage_client.get_bucket(canay_bucket_name).delete(force=True)
+    except NotFound:
+        pass
+    if sig:
```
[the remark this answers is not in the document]
mahrsee1997: `sig` is just the signal number like in case of `KeyboardInterrupt (Ctrl+C)` it will be 2.

weather_mv/loader_pipeline/pipeline.py
mahrsee1997: Should we handle some other signals ? WDYT ? 
Currently handled SIGINT (Ctrl+c) and SIGTSTP (Ctrl+z)

alxmrs: Thanks for the link to the SO post. TIL!

I think these signals are fine for now.

alxmrs (approved):
LGTM!

PR #160, opened 2022-06-06
Support for Eumetsat in weather-dl

mahrsee1997:


weather_dl/download_pipeline/clients.py
```diff
@@ -209,6 +215,101 @@ def num_requests_per_key(self, dataset: str) -> int:
...
+            timeout = 900  # customisation can be slow
+            tic = time.time()
+            while time.time() - tic < timeout:
+                if customisation.status == 'DONE':
+                    break
+            else:
+                raise TimeoutError(f'Customisation took longer than {timeout}s')
+            self.logger.info('Customisation for product %s: output %s', product, customisation.outputs)
+            with customisation.stream_output(customisation.outputs[0]) as stream:
+                with open(output, 'wb') as fdst:
```
uhager, line 266: Have you tested that this works? I initially managed to download some files directly in netcdf using this, but recently it has failed with various errors. I think unless this works reliably, we should just support downloading native.

weather_dl/download_pipeline/clients.py
```diff
@@ -209,6 +215,101 @@ def num_requests_per_key(self, dataset: str) -> int:
...
+        EUMETSAT Data Tailor allows maximum 3 (active and/or queued) customisations per user, as of Jun 06, 2022.
+        See: https://eumetsatspace.atlassian.net/wiki/spaces/DSDT/pages/1589379078
+
+        Above limitation is not applicable in case of native downloads.
+
+        If maximum number of (active and/or queued) customizations exceeds 3, then -
+        EUMETSAT Data Tailor API throws "You are exceeding your maximum number 3 of queued+running customisations."
+        error.
+
+        User's personal workspace is restricted to 20 GB. To resolve workspace size exhaust error, please delete
```
uhager, line 297: From what I remember this only applies to the web interface?

weather_dl/download_pipeline/partition.py
```diff
@@ -137,8 +153,14 @@ def prepare_partitions(config: Config) -> t.Iterator[Config]:
     Returns:
         An iterator of `Config`s.
     """
-    for option in itertools.product(*[config.selection[key] for key in config.partition_keys]):
-        yield _create_partition_config(option, config)
+    if config.client == 'eumetsat':
+        client = CLIENTS[config.client](config)
+        slices = time_slices(config.selection['start_date'], config.selection['end_date'])
+        for option in client.products(slices, config.dataset):
+            yield _create_partition_config((option, ), config)
```
uhager, line 160: Does this run in parallel for each slice? Querying the products can take some time, so this should run in parallel on several workers as part of the pipeline.

mahrsee1997: Necessary code changes done. Thanks.

weather_dl/download_pipeline/pipeline.py
```diff
@@ -166,9 +167,15 @@ def run(argv: t.List[str], save_main_session: bool = True) -> PipelineArgs:
...
-        )
+        if client_name == 'eumetsat':
+            num_requesters_per_key = client.num_requests_per_key(
+                config.dataset,
+                bool(config.selection.get('eumetsat_format_conversion_to'))
+            )
+        else:
+            num_requesters_per_key = client.num_requests_per_key(
+                config.dataset
+            )
```
alxmrs, line 184: This shouldn't be necessary. We already delegate calculating the number of requests per key by the download class. Instead of adding an if statement, let's see if we can add this logic without creating a special case for this client. Refactoring the general case is OK.

weather_dl/download_pipeline/clients.py
```diff
@@ -209,6 +215,105 @@ def num_requests_per_key(self, dataset: str) -> int:
...
+class EumetsatClient(Client):
+    def __init__(self, config: Config, level: int = logging.INFO,
+                 config_subsection: t.Optional[t.Tuple] = None) -> None:
+        super().__init__(config, level)
+        self.key = config.kwargs.get('api_key', os.environ.get("EUMETSATAPI_KEY"))
+        self.secret = config.kwargs.get('api_secret', os.environ.get("EUMETSATAPI_SECRET"))
+        if not self.key and config_subsection:
+            self.key = config_subsection[1].get('api_key')
+        if not self.secret and config_subsection:
+            self.secret = config_subsection[1].get('api_secret')
```
alxmrs, line 227: Why do we need to pass in a config subsection here? Shouldn't this already be done in the partition step?  The standard key is the one from the subsection.

weather_dl/download_pipeline/clients.py
```diff
@@ -209,6 +215,105 @@ def num_requests_per_key(self, dataset: str) -> int:
...
+    def download_native(self, product: eumdac.product.Product, output: str) -> None:
+        """Downloads the entry file from the product."""
+        entry = f'{product}.nat'
+        with product.open(entry=entry) as fsrc:
+            with open(output, 'wb') as fdst:
+                shutil.copyfileobj(fsrc, fdst, DEFAULT_READ_BUFFER_SIZE)
+
+    def download_custom(self, product: eumdac.product.Product, token: eumdac.AccessToken,
+                        chain_config: eumdac.tailor_models.Chain, output: str) -> None:
+        """Downloads the prduct after customisation."""
```
alxmrs, line 257: Typo: `prduct`

weather_dl/download_pipeline/clients.py
```diff
@@ -209,6 +215,105 @@ def num_requests_per_key(self, dataset: str) -> int:
...
+                    break
+            else:
+                raise TimeoutError(f'Customisation took longer than {timeout}s')
+            self.logger.info('Customisation for product %s: output %s', product, customisation.outputs)
+            with customisation.stream_output(customisation.outputs[0]) as stream:
+                with open(output, 'wb') as fdst:
+                    shutil.copyfileobj(stream, fdst, DEFAULT_READ_BUFFER_SIZE)
+
+    def retrieve(self, dataset: str, selection: t.Dict, output: str) -> None:
+        selection_ = optimize_selection_partition(selection)
```
alxmrs, line 274: If we only allow partition by ID, then I think this is unnecessary?

weather_dl/download_pipeline/clients.py
```diff
@@ -209,6 +215,105 @@ def num_requests_per_key(self, dataset: str) -> int:
...
+        return eumdac.AccessToken(credentials)
+
+    def download_native(self, product: eumdac.product.Product, output: str) -> None:
+        """Downloads the entry file from the product."""
+        entry = f'{product}.nat'
+        with product.open(entry=entry) as fsrc:
+            with open(output, 'wb') as fdst:
+                shutil.copyfileobj(fsrc, fdst, DEFAULT_READ_BUFFER_SIZE)
+
+    def download_custom(self, product: eumdac.product.Product, token: eumdac.AccessToken,
```
alxmrs, line 255: Let's use the retry decorator util here, too.

weather_dl/download_pipeline/clients.py
```diff
@@ -209,6 +215,105 @@ def num_requests_per_key(self, dataset: str) -> int:
...
+        datastore = eumdac.DataStore(self._token())
+        collection = datastore.get_collection(dataset)
+        for product in collection.search(dtstart=start, dtend=end):
+            yield product._id
+
+    def _token(self) -> eumdac.AccessToken:
+        credentials = (self.key, self.secret)
+        return eumdac.AccessToken(credentials)
+
+    def download_native(self, product: eumdac.product.Product, output: str) -> None:
```
alxmrs, line 248: Let's use the retry decorator util here.

alxmrs (changes requested):
Hey Rahul. I've gone through another revision and left a few minor notes. However, let's schedule some time for us to meet to discuss a slightly new approach for this change. I think we may be able to maintain our partition strategy moreso that I originally though, after looking at the underlying API. Let's meet to brainstorm the topic.

PR #163, opened 2022-06-13
DL: Add support for MARS syntax in JSON config

### LINKED ISSUE #48: JSON files do not support MARs range syntax.
https://github.com/google/weather-tools/blob/2d23ffb508a310acd604e4cff9be6ab22f507ef4/weather_dl/download_pipeline/parsers.py#L72

deepgabani8:
Fixes #48 .


weather_dl/download_pipeline/parsers.py
```diff
@@ -282,15 +284,15 @@ def date_range(start: datetime.date, end: datetime.date, increment: int = 1) ->
...
-    config = dict(config_parser.items(section))
-
+def _parse_lists(config: dict, section: str = '') -> t.Dict:
+    """Parses multiline blocks in *.cfg and *.json files as lists."""
     for key, val in config.items():
-        if '/' in val and 'parameters' not in section:
-            config[key] = parse_mars_syntax(val)
-        elif '\n' in val:
-            config[key] = _splitlines(val)
+        # Check string type for backward compatibility. i.e. <= v0.1.7 supports "padding": 0 in json config
```
alxmrs, line 290: Is this a TODO?

weather_dl/download_pipeline/parsers_test.py
```diff
@@ -63,15 +63,177 @@ def test_throws_error(self):
...
+
+    def test_json_raises_syntax_error_missing_left(self):
+        with self.assertRaises(SyntaxError) as ctx:
+            with io.StringIO('{"section": {"key": "value", "list": "/to/2020-01-07"}}') as f:
+                parse_config(f)
+
+        self.assertEqual("Improper range syntax in '/to/2020-01-07'.", ctx.exception.args[0])
+
+    def test_json_raises_syntax_error_missing_increment(self):
+        with self.assertRaises(SyntaxError) as ctx:
```
alxmrs, line 174: Since I've provided the other test examples, I've found a better way to write these asserts: 

```
with self.assertRaisesRegex(SyntaxError, "Improper range syntax in '2020-01-07/to/2020-01-11/by/'."):
     # ...
```

weather_dl/download_pipeline/parsers.py
```diff
@@ -149,8 +149,10 @@ def typecast(key: str, value: t.Any) -> t.Any:
 def parse_config(file: t.IO) -> t.Dict:
     """Parses a `*.json` or `*.cfg` file into a configuration dictionary."""
     try:
-        # TODO(b/175429166): JSON files do not support MARs range syntax.
-        return json.load(file)
+        config = json.load(file)
+        config_by_section = {s: _parse_lists(v, s) for s, v in config.items()}
+        config_with_nesting = parse_subsections(config_by_section)
+        return config_with_nesting
```
alxmrs, line 155: This is great! 

Is there some way we can combine the logic of the JSON and Config parsing? It seems like there are repeated steps here.

weather_dl/download_pipeline/parsers_test.py
```diff
@@ -63,15 +63,177 @@ def test_throws_error(self):
...
-        with io.StringIO('{"key": "value", "brokenKey": }') as f:
+        with io.StringIO('{"section": {"key": "value", "brokenKey": }}') as f:
             actual = parse_config(f)
             self.assertDictEqual(actual, {})
 
+    def test_json_produces_lists(self):
+        with io.StringIO('{"section": {"key": "value", "list": [0, 10, 20, 30, 40]}}') as f:
+            actual = parse_config(f)
+            for key, val in actual['section'].items():
+                self.assertNotIn('\n', val)
```
alxmrs, line 79: Since this is repeated quite a bit, let's extract it into an (assert) method?

alxmrs (changes requested):
Looks good so far – I've left a couple of notes.

weather_dl/download_pipeline/parsers.py
deepgabani8: Yes.
Just needed to add one condition for it _if not config:_

weather_dl/download_pipeline/parsers_test.py
deepgabani8: Modified it for all similar test cases in parset_test

weather_dl/download_pipeline/parsers.py
deepgabani8: Naa. It was to explain why we're checking for string type in the below line.
The message could've been better.
Updated one: **# Checks str type for backward compatibility since it also support "padding": 0 in json config**

weather_dl/download_pipeline/parsers.py
```diff
@@ -147,10 +147,9 @@ def typecast(key: str, value: t.Any) -> t.Any:
     return converted
 
 
-def parse_config(file: t.IO) -> t.Dict:
-    """Parses a `*.json` or `*.cfg` file into a configuration dictionary."""
+def _read_config_file(file: t.IO) -> t.Dict:
+    """Reads a `*.json` or `*.cfg` file into a configuration dictionary."""
```
alxmrs, line 151:
```suggestion
    """Reads a `*.json` or `*.cfg` files."""
```

weather_dl/download_pipeline/parsers.py
```diff
@@ -299,15 +303,15 @@ def date_range(start: datetime.date, end: datetime.date, increment: int = 1) ->
...
-
+def _parse_lists(config: dict, section: str = '') -> t.Dict:
+    """Parses multiline blocks in *.cfg and *.json files as lists."""
     for key, val in config.items():
-        if '/' in val and 'parameters' not in section:
-            config[key] = parse_mars_syntax(val)
-        elif '\n' in val:
-            config[key] = _splitlines(val)
+        # Checks str type for backward compatibility since it also support "padding": 0 in json config
+        if isinstance(val, str):
```
alxmrs, line 310: style nit: consider "returning" early (or, `continue`ing early) instead of deeper nesting.

weather_dl/download_pipeline/parsers_test.py
```diff
@@ -62,16 +62,150 @@ def test_throws_error(self):
...
     def test_bad_json(self):
-        with io.StringIO('{"key": "value", "brokenKey": }') as f:
+        with io.StringIO('{"section": {"key": "value", "brokenKey": }}') as f:
             actual = parse_config(f)
             self.assertDictEqual(actual, {})
 
+    def test_json_produces_lists(self):
+        with io.StringIO('{"section": {"key": "value", "list": [0, 10, 20, 30, 40]}}') as f:
+            actual = parse_config(f)
+            self._assert_not_in_dict_values('\n', actual['section'])
```
alxmrs, line 82: Since the arguments are the same for all of these, the helper method can be even more specific: 
```suggestion
            self._assert_no_newlines_in_section(actual)
```

We can make it general when we have a need to.  See the YAGNI principle [1].

[1]: https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

alxmrs (approved):
A few minor nits. Otherwise, this change looks great! Please merge this when the outstanding issues have been addressed.

PR #167, opened 2022-06-21
Changes in weather-mv tool to handle multiple dataset (primarily for grib's multiple hypercubes).

### LINKED ISSUE #73: Skipping variable error
Trying to spin up a streaming dataflow job with the below command:
weather-mv --uris "gs://ecmwf-xxxxx/A1D*.bz2" --output_table xxxx-xx.xxxx.xxxx  --topic "projects/xxxx/topics/ecmwf-xxxxx"  --runner DataflowRunner   --project xxx-xxxx   --region us-central1   --temp_location gs://xxxxx-test --job_name ecmwf-xxxxxx-test2

Below is the error. The dataflow job hasn't been created.

image

mahrsee1997:
Fixes #73 .


weather_mv/loader_pipeline/bq.py
```diff
@@ -149,11 +152,30 @@ def map_dtype_to_sql_type(var_type: np.dtype) -> str:
     raise ValueError(f"Unknown mapping from '{var_type}' to SQL type")
 
 
+def sanitize_field_name(name: str) -> str:
+    """Sanitize BigQuery field name based on https://cloud.google.com/bigquery/docs/schemas#column_names."""
+    sanitized_name = None
```
alxmrs, line 157: Nit: this can probably be omitted.

weather_mv/loader_pipeline/bq.py
```diff
@@ -149,11 +152,30 @@ def map_dtype_to_sql_type(var_type: np.dtype) -> str:
...
+    # The maximum column name length is 300 characters.
+    sanitized_name = name[:300]
```
alxmrs, line 159: will this create an out of bounds error?

weather_mv/loader_pipeline/bq.py
```diff
@@ -197,38 +219,42 @@ def prepare_coordinates(uri: str, *,
...
-            to_keep = variables + indices
-            to_drop = set(df.columns) - set(to_keep)
-            df.drop(to_drop, axis=1, inplace=True)
-
-        num_chunks = math.ceil(len(df) / coordinate_chunk_size)
-        for i in range(num_chunks):
-            chunk = df[i * coordinate_chunk_size:(i + 1) * coordinate_chunk_size]
-            yield uri, import_time, first_time_step, chunk
+    with open_dataset(uri, open_dataset_kwargs, disable_in_memory_copy, tif_metadata_for_datetime) as open_ds_list:
+        for ds in open_ds_list:
```
alxmrs, line 223: Do these dataframes have consistent coordinates?

weather_mv/loader_pipeline/bq_test.py
```diff
@@ -370,50 +370,18 @@ def test_multiple_editions(self):
         self.test_data_path = f'{self.test_data_folder}/test_data_grib_multiple_edition_single_timestep.bz2'
         actual = next(self.extract(self.test_data_path))
         expected = {
```
alxmrs, line 372: Ideally, each row would have all available variables for each lat/lng/time coordinate. Is that possible?

alxmrs (changes requested):
I think this is a good initial iteration. However, the data representation in the database could use some work: Ideally, there is one row of data per lat/lng/time coordinate (or, whatever our index is). For that index, all variables should be available. 

As I see it now, users would have to deal with duplicate rows for every location, where non-overlapping columns are null in each row. We should bear the work to normalize / clean the data on ingestion.

PR #170, opened 2022-06-24
Cap max number of workers in weather-dl.

### LINKED ISSUE #169: weather-dl uses to many workers
While we no longer _have_ to manage individual workers for correctness, with our current approach for the downloader, the autoscale algorithm uses a lot of workers in practice. We should cap the max number of workers to the # of possible simultaneous requests + a fudge factor. This way, we won't exhaust our quotas.

mahrsee1997:
Fixes #169.


alxmrs (approved):
A sweet and simple change. Does this work e2e? I bet you've tested it, but I want to double check. If so, LGTM!

PR #187, opened 2022-07-07
weather-mv: EE ingestion pipeline

### LINKED ISSUE #177: Extend weather-mv to ingest data into Earth Engine (EE)

deepgabani8:
Fixes #177 

The functionality to ingest grib data into the earth engine.

It will convert the grib data into the data format which the earth engine supports at the moment - GeoTiff (.tiff).
It will then upload the tiff image into the earth engine.


weather_mv/README.md
```diff
@@ -16,6 +16,7 @@ like [Google BigQuery](https://cloud.google.com/bigquery) (_alpha_).
 * _(new)_ **Grib Regridding**: `weather-mv regrid` uses [MetView](https://metview.readthedocs.io/en/latest/) to
   interpolate Grib files to a
   [range of grids.](https://metview.readthedocs.io/en/latest/metview/using_metview/regrid_intro.html?highlight=grid#grid)
+* _(new)_ **Earth Eangine Ingestion**: `weather-mv earthengine` converts weather data into GeoTiff and allows for data ingestion into earthengine.
```
fredzyda, line 19: Engine.

weather_mv/loader_pipeline/ee.py
```diff
@@ -0,0 +1,371 @@
...
+            tiff_name = self._get_tiff_name(uri)
+            start_time, end_time = (attrs.get(key) for key in ('start_time', 'end_time'))
+            dtype, crs, transform = (attrs.pop(key) for key in ['dtype', 'crs', 'transform'])
+
+            file_name = f'{tiff_name}.tiff'
+
+            with tempfile.TemporaryDirectory() as temp_dir:
+                local_tiff_path = os.path.join(temp_dir, file_name)
+                with rasterio.open(local_tiff_path, 'w',
+                                   driver='GTiff',
```
fredzyda, line 259: I think this is the wrong driver. I think you should be using the COG driver. Documentation for it is here: https://gdal.org/drivers/raster/cog.html#raster-cog

weather_mv/loader_pipeline/ee.py
```diff
@@ -0,0 +1,371 @@
...
+                                   count=len(data),
+                                   crs=crs,
+                                   transform=transform,
+                                   tiled=True) as f:
+                    for i, da in enumerate(data):
+                        f.write(da, i+1)
+
+                # Copy local tiff to gcs.
+                target_path = self._get_target_path(file_name)
+                with open(local_tiff_path, 'rb') as src:
```
fredzyda, line 272: I'm a little confused by this part. Can't you just have rasterio open up an in-memory buffer and then pass that buffer directly to the shutil.copyfileobj call? This seems like a disk write that you don't have to do...

weather_mv/loader_pipeline/ee.py
```diff
@@ -0,0 +1,371 @@
...
+    use_personal_account: bool
+
+    @retry.with_exponential_backoff()
+    def start_ingestion_task(self, asset_request: t.Dict) -> str:
+        """Sends a manifest task request to EE for asset ingestion. Returns the task id."""
+        task_id = ee.data.newTaskId(1)[0]
+        _ = ee.data.startIngestion(task_id, asset_request)
+        return task_id
+
+    def manifest_output(self,
```
fredzyda, line 308: Oh, you're doing ingestion here? I'm not sure this is what we wanted. If you write out a COG, you can then write a COG-backed asset directly to Earth Engine and avoid having to wait for an ingestion task to finish. This is way faster and generally what we use. We should have a conversation about wether we want COG-backed ingestion or geotiff to native ingestion here. Perhaps both, but I think COG-backed assets are what we'll want to use most of the time.

weather_mv/loader_pipeline/ee.py
```diff
@@ -0,0 +1,371 @@
...
+    """
+
+    ee_asset: str
+    use_personal_account: bool
+
+    @retry.with_exponential_backoff()
+    def start_ingestion_task(self, asset_request: t.Dict) -> str:
+        """Sends a manifest task request to EE for asset ingestion. Returns the task id."""
+        task_id = ee.data.newTaskId(1)[0]
+        _ = ee.data.startIngestion(task_id, asset_request)
```
fredzyda, line 305: For a COG-backed asset, the call you want here is ee.data.createAsset. That will return immediately and not start a task. See the documentation here: https://developers.google.com/earth-engine/apidocs/ee-data-createasset

fredzyda (changes requested):
I think this is a fine implementation of EE ingestion to internal EE assets, but I think the COG backed assets are more useful for this tool given how much faster the ingestion is. We should talk about which ingestion technique we want to support or if we should just go ahead and support both with a flag. I will also mention this to @alxmrs to see what his opinion is.

weather_mv/loader_pipeline/util.py
```diff
@@ -18,14 +18,32 @@
...
 from xarray.core.utils import ensure_us_time_resolution
 
+from google.cloud import bigquery, storage
+from google.api_core.exceptions import NotFound
+from google.api_core.exceptions import BadRequest
+
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+CANARY_BUCKET_NAME = 'anthromet_canary_bucket'
```
alxmrs, line 41: Thanks for moving this to utils.

weather_mv/loader_pipeline/ee.py
deepgabani8: Agreed. Updated it.

weather_mv/loader_pipeline/ee.py
```diff
@@ -0,0 +1,371 @@
...
+        return tiff_name
+
+    def _get_target_path(self, file_name: str) -> str:
+        """Creates a GCS target path from the file name."""
+        return os.path.join(self.tiff_location, file_name)
+
+    def process(self, uri: str) -> t.Iterator[TiffData]:
+        """Opens grib files and yields TiffData."""
+
+        logger.info(f'Converting {uri!r} to COGs ...')
```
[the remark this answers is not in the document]
deepgabani8: Oh I added the dry run check in ToEarthEngine's expand. It will just log these path uris and won't go inside any transforms. WDYT? Should we let it go inside each transform in case of dry run?

```
def expand(self, paths):
        """Converts gribs from paths into tiff and uploads them into the earth engine."""
        if not self.dry_run:
            (
                paths
                | 'ConvertToCog' >> beam.ParDo(
                    ConvertToCog(
                        tiff_location=self.tiff_location,
                        open_dataset_kwargs=self.xarray_open_dataset_kwargs,
                        disable_in_memory_copy=self.disable_in_memory_copy,
                        disable_grib_schema_normalization=self.disable_grib_schema_normalization))
                | 'IngestIntoEE' >> beam.ParDo(
                    IngestIntoEE(
                        ee_asset=self.ee_asset,
                        use_personal_account=self.use_personal_account))
            )
        else:
            (
                paths
                | 'Log Grib Files' >> beam.Map(logger.debug)
            )
```

weather_mv/loader_pipeline/ee.py
```diff
@@ -0,0 +1,412 @@
...
+        else:
+            (
+                paths
+                | 'Log Grib Files' >> beam.Map(logger.debug)
+            )
+
+
+@dataclasses.dataclass
+class FilterFiles(beam.DoFn):
+    """Filters out paths for which the assets that are already in the earth engine.
```
alxmrs, line 265: Nice, this is a useful class.

weather_mv/loader_pipeline/util.py
```diff
@@ -162,3 +184,184 @@ def get_coordinates(ds: xr.Dataset, uri: str = '') -> t.Iterator[t.Dict]:
...
+            _cleanup_bigquery(bigquery_client, canary_output_table)
+        signal.signal(signal.SIGINT, original_sigint_handler)
+        signal.signal(signal.SIGINT, original_sigtstp_handler)
+
+
+def _shard(elem, num_shards: int):
+    return (np.random.randint(0, num_shards), elem)
+
+
+class RateLimit(beam.PTransform, abc.ABC):
```
alxmrs, line 283: Excellent RateLimit utility!

alxmrs (approved):
Added some notes, but no major changes requested. LGTM, Given @fredzyda gives it the 👍🏽 too, this is ready for merge (after requested changes are addressed).

weather_mv/loader_pipeline/ee.py
```diff
@@ -0,0 +1,470 @@
...
+        logger=logger.warning,
+        initial_delay_secs=INITIAL_DELAY,
+        max_delay_secs=MAX_DELAY
+    )
+    def start_ingestion(self, asset_request: t.Dict) -> str:
+        """Creates COG-backed asset in earth engine. Returns the asset id."""
+        try:
+            result = ee.data.createAsset(asset_request)
+        except ee.EEException as err:
+            raise err
```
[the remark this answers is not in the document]
fredzyda: I agree, I think the only reason to catch and then re-raise this exception would be to ensure logging happened. Given that the retry is in place, it may be that the first exceptions may not be logged as it retries.

fredzyda (commented):
Generally looks good to me. I'm happy to merge this when @alxmrs is satisfied.

alxmrs:
Our CI system was broken due to downstream dependency changes. I've just pushed a fix in #193. Please merge from the main branch to get our tests to work before the final merge.

weather_mv/loader_pipeline/ee_test.py
```diff
@@ -0,0 +1,87 @@
...
+        self.test_tiff_dir = f'{self.test_data_folder}'  # For resultant tiff.
+        self.convert_to_cog = ConvertToCog(
+            tiff_location=self.test_tiff_dir
+        )
+        self.test_tiff_paths = []
+
+    def tearDown(self):
+        # Cleans up generated tiff files.
+        for test_tiff_path in self.test_tiff_paths:
+            os.remove(test_tiff_path)
```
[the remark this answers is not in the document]
deepgabani8: Yes, thank you.

weather_mv/loader_pipeline/ee_test.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def test_convert_to_cog(self):
+        test_data_path = f'{self.test_data_folder}/test_data_grib_single_timestep'
+        test_tiff_data = self.convert_to_cog.process(test_data_path)
+        expected_tiff_path = os.path.join(self.test_tiff_dir, f'{next(test_tiff_data).name}.tiff')
+        # The size of tiff is expected to be more than grib.
+        self.assertTrue(os.path.getsize(expected_tiff_path) > os.path.getsize(test_data_path))
```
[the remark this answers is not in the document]
deepgabani8: `Caveat: I don't like how the comment the AAA section names, that's unnecessary` would you elaborate this please?

weather_mv/loader_pipeline/ee.py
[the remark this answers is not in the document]
deepgabani8: Yes, logging the error.

weather_mv/loader_pipeline/ee.py
```diff
@@ -0,0 +1,470 @@
...
+
+    Args:
+        use_personal_account: A flag to use personal account for ee authentication.
+        service_account: Service account address when using a private key for earth engine authentication.
+        private_key: A private key path to authenticate earth engine using private key.
+
+    Returns:
+        cred: Credentials object.
+    """
+    # TODO(deepgabani8): Test private key authentication.
```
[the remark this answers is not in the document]
deepgabani8: Right, created the issue #197.

weather_mv/loader_pipeline/ee_test.py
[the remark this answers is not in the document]
alxmrs: Sure thing: I think that including comments like `# Arrange`, `# Act`, and `# Assert` are unnecessary. However, using whitespace to distinguish between each section is great.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -115,6 +115,12 @@ def _get_band_data(i):
     return ds
 
 
+def _to_utc_timestring(np_time: np.datetime64) -> str:
+    """Turn a numpy datetime64 into floating point seconds."""
+    timestamp = float((np_time - np.datetime64(0, 's')) / np.timedelta64(1, 's'))
+    return datetime.datetime.utcfromtimestamp(timestamp).strftime('%Y-%m-%dT%H:%M:%SZ')
```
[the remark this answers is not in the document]
alxmrs: Ah, thanks for checking. Your original formatting LGTM in that case.

fredzyda (approved):

PR #188, opened 2022-07-07
Enforce using temporary directories in regridding.

### LINKED ISSUE #181: Regrid jobs run out of disk space
On realistic test datasets, we've experienced `OSError` and Dataflow errors (like this: https://stackoverflow.com/a/67731781) that indicate that the system is running out of Disk space. 

It seems like there's a disk leak in the transformation.

alxmrs:
Fixes #181. Tested e2e with realistic data by myself and @pbattaglia.


weather_mv/README.md
```diff
@@ -194,6 +194,11 @@ To use this capability of the weather mover, please use the `[regrid]` extra whe
 pip install google-weather-tools[regrid]
 ```
 
+> **Warning**: MetView requires a decent amount of disk space in order to perform any regrid operation! Intermediary 
+> regridding steps will write temporary grib data to disk. Thus, please make use of the `--disk_size_gb` Dataflow 
+> option. A good rule of thumb would be to consume `30 + 3.5x` GBs of disk, where `x` is the size of each source data
```
fredzyda, line 199: That's huge!

weather_mv/loader_pipeline/regrid.py
```diff
@@ -96,6 +98,11 @@ def apply(self, uri: str):
 
             src.flush()
 
+            # Clear the metview temporary directory.
```
fredzyda, line 101: Why not use tempdir's context manager and just make a directory that goes away from hen you exit the context?

fredzyda (approved):
Seems alright, but why not use tempfile to make a tempdir that goes away on its own?

weather_mv/loader_pipeline/regrid.py
alxmrs: We've tried digging deep into MetView's sources, and they don't offer an interface for users to control how their tempfiles get cleaned up (or, where the tempfile is). Instead, they've more-or-less "hard-coded" a pattern where they'll clean up caches once the process exits. That doesn't suit our needs, since the dataflow instance will stay on for a long time.

weather_mv/README.md
alxmrs: This is more or less the recommendation from @pbattaglia, or at least, my interpretation of it. Pete, from your experiments, do you have any idea on how we could cut down this disk usage?

PR #198, opened 2022-07-14
Refactored weather-mv code to fix memory issue.

mahrsee1997:
In the current version of code (without this fix), we try to convert xArray.Dataset into pandas.DataFrame (basically this will attempt to load complete dataset in-memory). So as a result OOM killer is invoked in case of working with a large file (eg: 22G). 

`e2e` bench-marking results:
- Single 260K records file - 
Current Version (without the fix) : ~10.43 min
Fixed Code: ~11 min
- Ten 260K records file -
Current Version (without the fix) : ~16.5 min
Fixed Code: ~18.11 min

### DIFF
--- a/weather_mv/loader_pipeline/bq.py
+++ b/weather_mv/loader_pipeline/bq.py
@@ -16,7 +16,6 @@ import dataclasses
 import datetime
 import json
 import logging
-import math
 import os
 import signal
 import sys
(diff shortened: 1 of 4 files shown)


weather_mv/loader_pipeline/bq.py
```diff
@@ -300,59 +305,59 @@ def prepare_coordinates(
...
-def _convert_time(val) -> t.Any:
-    """Converts pandas Timestamp values to ISO format."""
-    if isinstance(val, pd.Timestamp):
-        return val.replace(tzinfo=datetime.timezone.utc).isoformat()
-    elif isinstance(val, pd.Timedelta):
-        return val.total_seconds()
-    else:
-        return val
+        for chunk in mitertools.ichunked(get_coordinates(data_ds, uri), coordinate_chunk_size):
+            yield uri, list(chunk)
```
alxmrs, line 309: nice, looks good.

alxmrs (approved):
LGTM! Thanks for the quick turn around + benchmarks.

alxmrs:
Any idea why CI is failing here?

mahrsee1997:
> Any idea why CI is failing here?

It's because `more-itertools` dependency was not added in setup.py.

PR #200, opened 2022-07-20
Mv fix regrid leak

pbattaglia:
Fixed bug where the glob module needs to be imported inside the apply() to work properly with beam (for reasons I don't totally understand).


pbattaglia:
Any idea why this lint check is failing? It seems to be seeing the old code, not the new code I've pushed.

PR #204, opened 2022-07-25
Added check for ee setup on retries.

deepgabani8:
When the pipeline tries to ingest into EE, if it fails due to rate-limit then on the next try (on different worker I believe), the ee is may not be initialized. This happens only when rate-limits cause a failure in ingestion and pipeline retries.
It was introduced after adding the rate-limit functionality.

The fix here is to maintain a has_setup flag.


alxmrs (approved):
LGTM; thanks for the fix.

deep-gabani-google:
Tested it and it is working as expected. Merging it.

Issue #210, opened 2022-08-08
Dataflow runner jobs timeout due to long pip resolution times


blackvvine:
## Summary
The installation of the `apache-beam[gcp]` package takes too long due to [BEAM-22218](https://github.com/apache/beam/issues/22218), causing the Dataflow runner jobs to timeout and fail after an hour without any progress.

## Reproducing Steps
```bash
# set up .cdsapirc
git clone https://github.com/google/weather-tools.git
python3 -m venv env
. env/bin/activate
cd weather-tools
pip install . 
weather-dl configs/era5_example_config_local_run.cfg --runner DataflowRunner --project <PROJECT> --temp_location <SOME_GCS_PATH> --job_name morbius --region us-central1
```

## Expectation
Dataflow pipeline runs successfully

## Outcome
Dataflow worker is created but job never starts, fails after an hour with the error "The Dataflow job seems to be stuck".

## Symptoms
- Worker VM CPU usage at 100%
- Fetch job step never starts running, other steps never produce an output
- Infinite loop of INFO logs like:
```
INFO: pip is looking at multiple versions of google-cloud-bigquery to determine which version is compatible with other requirements. This could take a while.
```
- Final error:
```
The Dataflow job appears to be stuck because no worker activity has been seen in the last 1h. Please check the worker logs in Stackdriver Logging. You can also get help with Cloud Dataflow at https://cloud.google.com/dataflow/support.
```

## Diagnosis
Build fails after 2h when using `--prebuild_sdk_container_engine=cloud_build`
Pip tries over and over again to find the right version of `google-cloud-*` packages
According to [BEAM-22218](https://github.com/apache/beam/issues/22218), the issue seems to be resolved in newer pip versions but it's hard to apply their workaround without either using a custom image or downgrading Beam version.

## Workaround
Fixing GCP package versions and moving beam to the end of the list:

```python
base_requirements = [
    "cdsapi",
    "ecmwf-api-client",
    "numpy>=1.19.1",
    "pandas",
    "xarray",
    "requests>=2.24.0",
    "firebase-admin>=5.0.0",
    "google-cloud-firestore",
    "urllib3==1.26.5",
    "google-cloud-bigquery==2.34.4",
    "google-cloud-bigquery-storage==2.14.1",
    "google-cloud-bigtable==1.7.2",
    "google-cloud-core==1.7.3",
    "google-cloud-datastore==1.15.5",
    "google-cloud-dlp==3.8.0",
[...]

PR #212, opened 2022-08-11
Pip install workaround from #210

alxmrs:
Changing weather-dl's base requirement order in order to prevent pip resolution timeouts. Thanks, @blackvvine!


alxmrs:
Closing in favor of #213.

Issue #60, opened 2022-01-03
`weather-dl`: Implement BigQuery manifest


alxmrs:
Similar to #13, except the manifest should be a BigQuery table. This makes sense if users choose to use `weather-mv` and want their data queryable from one place.

blackvvine:
Since we're taking the time to fix this, wouldn't Cloud Logging be a better substitute for BQ, GCS, and Firebase?

alxmrs:
I'm less familiar with Cloud Logger. How would it help suit our needs here, which are primarily to have a central record of our download status?

blackvvine:
[Cloud Logging](https://cloud.google.com/logging) is Stackdriver's new name, where Dataflow pipeline logs show up too. Since it's very unlikely that the users are interested in any complex analytics on the manifest logs, BQ is not a great option. Stackdriver is cheaper, easier to use, and the permissions are less complex to manage. 

I've created and tested a tiny Stackdriver manifest implementation [here](https://github.com/blackvvine/weather-tools/tree/stackdriver-manifest). The logs will show up in Cloud Console like this:

<img width="1241" alt="Screen Shot 2022-09-18 at 6 12 07 PM" src="https://user-images.githubusercontent.com/9886871/190930035-787c5b5a-b896-42f2-8c38-9e1b59d53c87.png">

alxmrs:
What every the service / implementation, I ideally want to be able to answer the following questions quickly (say, with a local script or even a dashboard): 
- What data is currently in progress? What's queued for download? 
- What data have we already ingested? 
- Approximately, how long will it take to finish this job? 

Do you think that Cloud Logging would help us answer these questions? The `Manifest` is not really a logging system, but rather a database that we intend to query to answer these questions.

alxmrs:
To put this feature in terms of a problem: Right now, our default implementation for the manifest is `Firestore`. We have a CLI script we include with `weather-dl` that checks what the current status is: 
https://github.com/google/weather-tools/blob/main/weather_dl/download-status

Unfortunately, due to our database choice and current number of records, this script is not performant + hits API limits right away. This inspiration for this issue was: 
- it would make much more sense if we used a traditional RDBMS instead of a key-value store
- we'd prefer to use BigQuery rather than other DB implementations to start, since our project is pretty wed to BQ.

jayendra13:
[...]

PR #214, opened 2022-08-15
Fix gcs manifest

### LINKED ISSUE #208: "Invalid file open mode" error when using the GCS manifest in weather-dl
Can't run `weather-dl` with the GCS manifest type:

**Reproducing**

```bash
weather-dl configs/era5_example_config_local_run.cfg \
      -m "gs://path/to/manifest.txt" \
      --runner DataflowRunner \
      --project <PROJECT_ID>  \
      --temp_location "gs://path/to/temp_dir" \
      --job_name <JOB_NAME> \
      --region us-central1 \
```

**Error**
```
ValueError: Invalid file open mode: a.
```

The error is thrown [here](https://github.com/google/weather-tools/blob/de4fa216e428666d183a31155862db6f4420e424/weather_dl/download_pipeline/manifest.py#L202) which seems to be [expected](https://beam.apache.org/releases/pydoc/2.10.0/_modules/apache_beam/io/gcp/gcsio.html#:~:text=Invalid%20file%20open%20mode).

blackvvine:
Fixes #208

Appending is not possible in an object storage like GCS. If having the full update history of the manifest is a requirement, a loss-less version is also available [here](https://github.com/blackvvine/weather-tools/blob/10dc5dd2a886c73580748efb94d04eb268caf61a/weather_dl/download_pipeline/manifest.py#L197) but it seems unnecessary and overly complex as it relies on writing to different objects each time and merging them in the end, and Firestore manifest is also overwriting.


setup.py
```diff
@@ -14,10 +14,26 @@
 
 from setuptools import find_packages, setup
 
+beam_gcp_requirements = [
```
alxmrs, line 17: Before the merge of this PR, would you mind cleaning up the history?

weather_dl/download_pipeline/manifest.py
```diff
@@ -188,8 +188,8 @@ def _update(self, download_status: DownloadStatus) -> None:
 class GCSManifest(Manifest):
     """Writes a JSON representation of the manifest to GCS.
 
-    This is an append-only implementation, the latest value in the manifest
-    represents the current state of a download.
+    This is an overriding implementation, the latest value in the manifest
+    file represents the current state of a download.
```
alxmrs, line 192: It would be better if we didn't delete the previous download status data in the manifest.

weather_dl/download_pipeline/manifest_test.py
```diff
@@ -90,3 +97,40 @@ def is_valid_json(self, file: t.IO) -> None:
...
+                for blob in tmp_bucket.list_blobs():
+                    blob.delete()
+                # delete the bucket
+                tmp_bucket.delete()
+                logging.debug(f"Deleted temporary bucket {bucket_name}")
+            except Exception as e:
+                logging.error(e)
+                logging.error(f"Error deleting temporary bucket {bucket_name},"
+                              f" to avoid unnecessary Cloud Storage charges "
+                              f"make sure to delete it manually.")
```
alxmrs, line 136: How about we include this in a test case `tearDown` method?

alxmrs: Similarly, the bucket setup could be moved to a `setUp`.

alxmrs (changes requested):
Thanks for the fixes! I've added a few notes.

weather_dl/download_pipeline/manifest_test.py
blackvvine: You're right. Sorry I wrote these at 2 AM lol, I'll update the CL

weather_dl/download_pipeline/manifest.py
blackvvine: GCS does not support append write mode, so the alternatives would be either:

- Keeping the GCS client open for writes, opening the door for timeouts, connection failures and unexpected behaviour if the updates span a long interval (which they typically do). I'm not sure if flushing works like regular POSIX files in GCS, so the latest manifest might not be available either while the downloads are in progress.
- Writing to multiple files and merging them in the end. As mentioned above I already have implemented this solution [here](https://github.com/blackvvine/weather-tools/blob/10dc5dd2a886c73580748efb94d04eb268caf61a/weather_dl/download_pipeline/manifest.py#L197) but it's a bit overly complex. (*)
- Some kind of hybrid between the two where we pool the output files but rotate them every X minutes (kinda like apache's `logrotate`). I couldn't find an out-of-the-box implementation for this so we'll have to build it from scratch, which might not be worth the effort.

(*) It can be simplified a *little* bit by not using `compose` for the merge, but that would probably make it less time-efficient.

setup.py
blackvvine: Done! cherry-picked the relevant commits from main

weather_dl/download_pipeline/manifest.py
alxmrs: Here's an idea for another approach which contains tradeoffs: Assuming the lock works (this probably is not the case in a Dataflow environment), we could just open the manifest in read mode, copy it to local disk, append another row, and then write the data back to GCS. 

Given that this is a fragile area of code, a low performance option, and that we have plans for a better replacement (#60), I'm also happy to have the "fix" for this feature to simply be to remove the functionality. :)

blackvvine:
I update the change to remove GCS Manifest, we can continue the discussion in #60.

mahrsee1997:
Closing this PR as it has been addressed by the changes made in PR #295 which has been merged.

PR #218, opened 2022-08-15
Verify gcs target

### LINKED ISSUE #209: Validate weather-dl target location before starting the job
When the *weather-dl* job is submitted with an invalid path, e.g. a local path instead of GCS when using the Dataflow runner, the job never throws an error. Instead, it gets stuck in the exponential backoff loop of the `Fetcher.upload()` [method](https://github.com/google/weather-tools/blob/de4fa216e428666d183a31155862db6f4420e424/weather_dl/download_pipeline/fetcher.py#L57).

**Symptoms**

GCS write exponential backoff WARN message:

```
Retry with exponential backoff: waiting for 295.9412769663183 seconds before retrying open because we caught exception: FileNotFoundError: [Errno 2] No such file or directory: ''
```

blackvvine:
Fixes #209

### DIFF
--- a/weather_dl/download_pipeline/config.py
+++ b/weather_dl/download_pipeline/config.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import calendar
-import copy
 import copy as cp
 import dataclasses
 import itertools
@@ -80,7 +79,7 @@ def optimize_selection_partition(selection: t.Dict) -> t.Dict:
 
     Used to support custom syntax and optimizations, such as 'all'.
     """
-    selection_ = copy.deepcopy(selection)
+    selection_ = cp.deepcopy(selection)
 
     if 'day' in selection_.keys() and selection_['day'] == 'all':
         year, month = selection_['year'], selection_['month']


weather_dl/download_pipeline/config.py
```diff
@@ -109,3 +111,46 @@ def optimize_selection_partition(selection: t.Dict) -> t.Dict:
         del selection_['year']
 
     return selection_
+
+
+def prepare_partitions(config: Config) -> t.Iterator[Config]:
```
alxmrs, line 116: I'm curious: why did you move these functions to this file? It seems like `partition.py` is a good place for them.

weather_dl/download_pipeline/parsers.py
```diff
@@ -18,17 +18,27 @@
...
 from urllib.parse import urlparse
 
+from apache_beam.io.gcp import gcsio
+from google.api_core.exceptions import GoogleAPIError
+from google.cloud import storage
+from google.resumable_media import InvalidResponse
+
 from .clients import CLIENTS
 from .config import Config
+from .config import prepare_partitions
```
alxmrs, line 37: We do allow multiple imports per line in this codebase. Feel free to make this a 1-liner.

weather_dl/download_pipeline/parsers.py
```diff
@@ -326,6 +336,36 @@ def _number_of_replacements(s: t.Text):
     return len(set(format_names)) + num_empty_names
 
 
+def _validate_target_path(config: Config):
+    """
+    Checks accessibility of target locations before execution start
+    """
```
alxmrs, line 342: Please make this a 1-line docstring.

weather_dl/download_pipeline/parsers.py
```diff
@@ -326,6 +336,36 @@ def _number_of_replacements(s: t.Text):
...
+    client = storage.Client()
+    for partition_conf in prepare_partitions(config):
+        target = prepare_target_name(partition_conf)
+        parsed = urlparse(target)
+        if parsed.scheme == 'gs':
```
alxmrs, line 347: Is there a way this this validation could be written to make the check portable (to other clouds)? For example, via beam's FileSystem's API?

alxmrs (commented):
I've had to stop this review short; however, I have a few notes for things to work on so far. 

One general idea: Another approach to solve this would be to allow the pipeline to fail in the fetch step when this error is encountered (or, errors where failure is preferable).

weather_dl/download_pipeline/config.py
blackvvine: This was to avoid a cyclic dependency issue. In `partitions.py` we got:
```python
from .parsers import prepare_target_name
```

blackvvine: I just reverted that commit and confirmed it'd cause an `ImportError`

weather_dl/download_pipeline/parsers.py
blackvvine: That's a great point. I figured the "path writeable" check needs to distinguish between file systems so stuck to GCS. Nevertheless `FileSystem` does have an FS-agnostic way for checking whether the directory exists. I've updated the CL to support this.

weather_dl/download_pipeline/parsers.py
```diff
@@ -441,13 +471,19 @@ def require(condition: bool, message: str, error_type: t.Type[Exception] = Value
...
 
 
-def prepare_target_name(config: Config) -> str:
+def prepare_target_name(partition_config: Config) -> str:
     """Returns name of target location."""
-    partition_dict = OrderedDict((key, typecast(key, config.selection[key][0])) for key in config.partition_keys)
-    target = config.target_path.format(*partition_dict.values(), **partition_dict)
+    partition_dict = OrderedDict(
+        (key, typecast(key, partition_config.selection[key][0]))
+        for key in partition_config.partition_keys)
```
alxmrs, line 485: I'm ok with making this multiline, but then let's go all the way :)
```suggestion
    partition_dict = OrderedDict(
        (key, typecast(key, partition_config.selection[key][0]))
        for key in partition_config.partition_keys
    )
```

alxmrs (commented):
I have a few opened ended questions that generally ask how this validation will interact with other features in weather-dl. Also, some nits :) .

weather_dl/download_pipeline/parsers.py
```diff
@@ -326,6 +333,29 @@ def _number_of_replacements(s: t.Text):
...
+            elif parsed.scheme == "s3":
+                S3IO().open(target, 'w').close()
+            elif parsed.scheme == "" or parsed.scheme == "fs":
+                open(target, 'w').close()
+            else:
+                logger.warning(f"Unable to verify path {target} writeable."
+                               f"Persistent IO errors when writing to the target"
+                               f"path during execution may cause deadlocks.")
+        except Exception:
+            raise ValueError(f"Unable to write to {target}")
```
alxmrs, line 355: One more idea: let's chain the errors:
```suggestion
        except Exception as e:
            raise ValueError(f"Unable to write to {target}") from e
```

alxmrs:
I just re-read #209 -- a general thought: Should the fix even be in the config parser? It seems like the source of the error has to do with the pipeline args / environment. 

Furthermore, another possible fix is to change what kinds of errors we choose to retry and others we let bring down the pipeline. These are generally set here: 
https://github.com/google/weather-tools/blob/65b484ebdbc6af3c9311dbd097bf14712d96df4d/weather_dl/download_pipeline/util.py#L20

blackvvine:
> I just re-read #209 -- a general thought: Should the fix even be in the config parser? It seems like the source of the error has to do with the pipeline args / environment.
> 
> Furthermore, another possible fix is to change what kinds of errors we choose to retry and others we let bring down the pipeline. These are generally set here:
> 
> https://github.com/google/weather-tools/blob/65b484ebdbc6af3c9311dbd097bf14712d96df4d/weather_dl/download_pipeline/util.py#L20

I see point that logically it might fit better in the error handler, but IMO it makes more sense for the check to be done in the arg parser since it's fail-fast.  Even if we fix the error handler to tear down the job in case target location is unreachable, it does needlessly *incur costs* for the user to run the job up to the point of failure, and it takes somewhere between 5-20 minutes for it to reach there.

alxmrs:
Ok, the validation is very reasonable to me. I like the principle of failing fast. 

My one nit is: this should at least occur in the `run` function and not in the parser, since it's testing resource choices rather than parsing (e.g. it's validating, not parsing).

PR #219, opened 2022-08-17
Added band names in COG file.

deepgabani8:


alxmrs (approved):
LGTM -- thanks for the fix!

PR #220, opened 2022-08-22
Removed unnecessary properties - band names in metadata.

deepgabani8:
Earlier, the band names were added in COG metadata but now that we are setting the band names, it is unnecessary.


alxmrs (approved):
LGTM

PR #230, opened 2022-09-02
No 'more-itertools' dependency.

alxmrs:
Since we only needed one function from 'more-itertools', I decided just to implement it myself.

### DIFF
--- a/weather_mv/loader_pipeline/util.py
+++ b/weather_mv/loader_pipeline/util.py
@@ -159,7 +159,7 @@ def ichunked(lst: t.Iterable, n: int) -> t.Iterator[t.Iterable]:
     i = iter(lst)
     try:
         while True:
-            it = itertools.islice(lst, n)
+            it = itertools.islice(i, n)
             first = next(it)
             yield itertools.chain([first], it)
     except StopIteration:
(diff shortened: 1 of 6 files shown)


weather_mv/loader_pipeline/util.py
```diff
@@ -154,6 +154,19 @@ def _prod(xs: t.Iterable[int]) -> int:
     return reduce(operator.mul, xs, 1)
 
 
+def ichunked(lst: t.Iterable, n: int) -> t.Iterator[t.Iterable]:
```
mahrsee1997, line 157:
```suggestion
def ichunked(lst: t.Iterable, n: int) -> t.Iterator[t.Iterable]:
    """Yield evenly-sized chunks from an iterable."""
    i = iter(lst)
    try:
        while True:
            it = itertools.islice(lst, n)
            first = next(it)
            yield itertools.chain([first], it)
    except StopIteration:
        pass
```

weather_mv/loader_pipeline/util_test.py
```diff
@@ -40,6 +40,35 @@ def test_no_duplicate_coordinates(self):
         self.assertTrue(all((c == 1 for c in counts.values())))
 
 
+class IChunksTests(unittest.TestCase):
```
mahrsee1997, line 43: Lets include above mention `generator object` case here.

mahrsee1997 (approved):
LGTM ! Please resolve the last unresolved comment regarding the inclusion of `generator object` case in our test suite before merging.

PR #231, opened 2022-09-08
Zarr-to-Zarr regridding with MetView.

alxmrs:
This change allows us to regrid Zarrs with MetView.

This PR will aid in accomplishing Phase 2 of https://github.com/google-research/ARCO-ERA5. There are three ways this routine could be used. First, we can create a version of our datasets that has been regridded to lat/lng with no other interpolation applied. Second, we can subclass the `Regrid` pipeline and override `apply_metview` to add any other fieldset operations we want to the chunks. This latter option allows us to regrid and interpolate data in one pass. Last, I offer a way to use process XArray datasets as MetView fieldsets via a PTransform. I also provide a PTransform for regridding with MetView.

CC: @wundersooner

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -30,6 +30,7 @@ dependencies:
   - pandas=1.5.1
   - pip=22.3
   - pygrib=2.1.4
+  - gcsfs=2022.11.0
   - pip:
     - earthengine-api==0.1.329
     - .[test]
--- a/environment.yml
+++ b/environment.yml
@@ -31,6 +31,7 @@ dependencies:
   - pip=22.3
   - pygrib=2.1.4
   - gcsfs=2022.11.0
+  - xarray-beam=0.3.1
   - pip:
     - earthengine-api==0.1.329
     - .[test]
--- a/weather_mv/loader_pipeline/ee.py
+++ b/weather_mv/loader_pipeline/ee.py
@@ -261,7 +261,8 @@ class ToEarthEngine(ToDataSink):
         pipeline_options = PipelineOptions(pipeline_args)
         pipeline_options_dict = pipeline_options.get_all_options()
 
-        assert not known_args.zarr, 'Reading Zarr is not (yet) supported.'
+        if known_args.zarr:
+            raise RuntimeError('Reading Zarr is not (yet) supported.')
 
         # Check that ee_asset is in correct format.
         if not re.match("^projects/.+/assets.*", known_args.ee_asset):
--- a/weather_mv/loader_pipeline/ee.py
+++ b/weather_mv/loader_pipeline/ee.py
@@ -261,6 +261,8 @@ class ToEarthEngine(ToDataSink):
         pipeline_options = PipelineOptions(pipeline_args)
         pipeline_options_dict = pipeline_options.get_all_options()
 
+        assert not known_args.zarr, 'Reading Zarr is not (yet) supported.'
+
         # Check that ee_asset is in correct format.
         if not re.match("^projects/.+/assets.*", known_args.ee_asset):
             raise RuntimeError("'--ee_asset' is required to be in format: projects/+/assets/*.")
--- a/weather_mv/loader_pipeline/regrid.py
+++ b/weather_mv/loader_pipeline/regrid.py
@@ -177,7 +177,6 @@ class Regrid(ToDataSink):
         with _metview_op():
             # mv.dataset_to_fieldset will error on input where there is only 1 value in
             # a dimension. This has to do with ECMWF's cfgrib being in a alpha version.
-            # Thus, we take an alternative route to writing an XArray dataset to grib...
             try:
                 fs = mv.dataset_to_fieldset(ds)
                 regridded = mv.regrid(data=fs, **self.regrid_kwargs)
--- a/weather_mv/loader_pipeline/regrid.py
+++ b/weather_mv/loader_pipeline/regrid.py
@@ -82,7 +82,6 @@ class Regrid(ToDataSink):
     to_netcdf: bool = False
     zarr_input_chunks: t.Optional[t.Dict] = None
     zarr_output_chunks: t.Optional[t.Dict] = None
-    pipeline: t.Optional[beam.Pipeline] = None
 
     def __post_init__(self):
         if not self.zarr_input_chunks:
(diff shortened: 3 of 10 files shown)


environment.yml
```diff
@@ -30,6 +30,7 @@ dependencies:
   - pandas=1.5.1
   - pip=22.3
   - pygrib=2.1.4
+  - gcsfs=2022.11.0
```
fredzyda, line 33: Is this supposed to match the setup.py? In other words, do you need xarray-beam here?

weather_mv/loader_pipeline/ee.py
```diff
@@ -261,6 +261,8 @@ def validate_arguments(cls, known_args: argparse.Namespace, pipeline_args: t.Lis
         pipeline_options = PipelineOptions(pipeline_args)
         pipeline_options_dict = pipeline_options.get_all_options()
 
+        assert not known_args.zarr, 'Reading Zarr is not (yet) supported.'
```
fredzyda, line 264: It seems weird to assert here where you raise more specific errors in other files...

weather_mv/loader_pipeline/regrid.py
```diff
@@ -32,16 +36,148 @@
...
+        zarr_input_chunks: (Optional) When regridding Zarr data, how the input
+            dataset should be chunked upon open.
+    """
+    regrid_kwargs: t.Dict
+    zarr_input_chunks: t.Optional[t.Dict] = None
+
+    def template(self, source_ds: xr.Dataset) -> xr.Dataset:
+        """Calculate the output Zarr template by regridding a tiny slice of the input dataset."""
+        # Silence Dask warning...
+        with dask.config.set(**{'array.slicing.split_large_chunks': False}):
```
fredzyda, line 124: This seems like a weird way to pass the arguments. Why not use regular keyword args?

weather_mv/loader_pipeline/regrid.py
```diff
@@ -32,16 +36,148 @@
...
+            zeros = source_ds.chunk().pipe(xr.zeros_like)
+
+            # Get a single timeslice of the zeros Dataset (or equivalent chunkable dimension).
+            # We don't know for sure that 'time' is in the Zarr dataset, so here we make our
+            # best attempt to find a good slice.
+            t0 = None
+            for dim in ['time', *(self.zarr_input_chunks or {}).keys()]:
```
fredzyda, line 131: Does this work? I feel like you'll get an empty list in here if zarr_input_chunks is None. This seems like an awkward way of doing this.

weather_mv/loader_pipeline/regrid.py
```diff
@@ -32,16 +36,148 @@
...
+        zarr_output_chunks: (Optional, recommended) When regridding Zarr data, how the output Zarr dataset should be
+            divided into chunks.
+    """
     output_path: str
     regrid_kwargs: t.Dict
     to_netcdf: bool = False
+    zarr_input_chunks: t.Optional[t.Dict] = None
+    zarr_output_chunks: t.Optional[t.Dict] = None
+
+    def __post_init__(self):
```
fredzyda, line 175: The fact that this is here makes me wish this class wasn't a dataclass anymore. I feel like anytime you're actually trying to manipulate the things stored in the class attributes as they're getting set, you should probably just write the init method.

weather_mv/loader_pipeline/regrid.py
```diff
@@ -52,16 +188,28 @@ def add_parser_arguments(cls, subparser: argparse.ArgumentParser) -> None:
...
+            raise ValueError('only Zarr-to-Zarr regridding is allowed!')
+
+        if not known_args.zarr and (known_args.zarr_input_chunks or known_args.zarr_output_chunks):
+            raise ValueError('chunks can only be set when input URI is a Zarr.')
 
     def target_from(self, uri: str) -> str:
         """Create the target path from the input URI.
 
-        Will change the extension if the target is a NetCDF.
+        In the case of Zarr, the output will be treated like a valid path.
```
fredzyda, line 207: This seems like strange and possibly unexpected behavior.

weather_mv/loader_pipeline/regrid.py
```diff
@@ -78,34 +226,49 @@ def apply(self, uri: str):
...
+                with FileSystems().create(self.target_from(uri)) as dst:
+                    shutil.copyfileobj(src, dst, WRITE_CHUNK_SIZE)
 
     def expand(self, paths):
-        paths | beam.Map(self.apply)
+        if not self.zarr:
+            paths | beam.Map(self.apply)
+            return
+
+        source_ds = xr.open_zarr(self.first_uri, **self.zarr_kwargs, chunks=None)
```
fredzyda, line 258: You're opening a file in expand? That seems less than ideal...

weather_mv/loader_pipeline/regrid_test.py
```diff
@@ -32,20 +33,22 @@
 
 def make_skin_temperature_dataset() -> xr.Dataset:
     ds = xr.DataArray(
-        np.zeros((5, 6)) + 300.,
+        np.zeros((4, 5, 6)) + 300.,
```
fredzyda, line 36: Consider np.full here. This seems a bit awkward.

fredzyda (commented):
Generally pretty good, but I have some questions.

weather_mv/loader_pipeline/regrid.py
alxmrs: Will this key work that way if it is `.` delimited?

environment.yml
alxmrs: I didn't know xarray-beam was on `conda`! Thanks.

weather_mv/loader_pipeline/regrid.py
```diff
@@ -32,16 +36,148 @@
...
     import metview as mv
+    Fieldset = mv.bindings.Fieldset
 except (ModuleNotFoundError, ImportError, FileNotFoundError):
     logger.error('Metview could not be imported.')
     mv = None  # noqa
+    Fieldset = t.Any
+
+
+def _clear_metview():
+    """Clear the metview temporary directory."""
```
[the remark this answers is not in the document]
alxmrs: Answered this in the discussion here: https://github.com/google/weather-tools/pull/188#discussion_r918443630

Will add a comment for clarification.

weather_mv/loader_pipeline/regrid.py
alxmrs: You're right -- this will return None in this case. It does throw an exception when this occurs and prompt the user to change `zarr_input_chunks` in this case. So far, this does work. 

My goal is to shrink the input dataset so I can regrid the coordinates without exploding the amount of space. Maybe there is a better way I could do this? 

Deferring to `zarr_input_chunks` makes sense to me since the user will probably have to specify a non-spatial dimension to iterate through during the rest of the pipeline. By default, we guess that the dataset has a 'time' dimension, since this is really common.

alxmrs: I've added a bit of code that will do something simpler if the dataset is small enough :)

weather_mv/loader_pipeline/regrid.py
alxmrs: Changed some default so that this post-init is no longer necessary :)

weather_mv/loader_pipeline/regrid.py
alxmrs: The output Zarr path just needs to be a directory, otherwise it could take any shape. I agree that there is a footgun here. I've added a user warning to the validation step to nudge users to user more standard zarr formatting.

weather_mv/loader_pipeline/regrid.py
alxmrs: Sort of. Mostly, I'm getting the Zarr metadata, which is necessary to open the dataset to chunks. 

https://docs.xarray.dev/en/stable/generated/xarray.open_zarr.html#:~:text=If%20chunks%3DNone%2C%20zarr%20array%20data%20will%20lazily%20convert%20to%20numpy%20arrays%20upon%20access

I'll add a comment here to clarify what's going on.

weather_mv/loader_pipeline/regrid_test.py
alxmrs: Thanks!

weather_mv/loader_pipeline/regrid.py
fredzyda: Oh, I didn't catch that bit of weirdness. You're probably right that it has to be this way. Ugh, definitely not a fan of that.

fredzyda (approved):
Thanks for the changes.

PR #235, opened 2022-09-19
Faster data transfers from MARS.

alxmrs:
I'm taking a leaf from @mahrsee1997's PR #148 so that we can copy data from the MARS server faster (using a larger buffer size).

Thanks for the primary contribution here, Rahul.


alxmrs:
That's sounds very reasonable to me. Good insight.

alxmrs:
This LGTM -- I cannot approve this CL since I am also an author.

fredzyda (approved):
I can approve it if Alex can't.

alxmrs:
Thanks Fred!

PR #236, opened 2022-09-21
Parallelizing fan-out step during `PreparePartitions` in `weather-dl`. 

mahrsee1997:
In this PR, we change the prepare partitions steps to a) use indexes instead of generating most of the configs and b) to split partitions (i.e. cartesian cross-products) into lists of lists. We aim to divide all partitions into a 50 lists (when the number is big), and  offer users a `-p`, `--partition-chunks` argument to let users influence the number of groups.


weather_dl/download_pipeline/partition.py
```diff
@@ -141,14 +142,14 @@ def prepare_partitions(config: Config) -> t.Iterator[Config]:
         yield _create_partition_config(option, config)
 
 
-def new_downloads_only(candidate: Config, store: t.Optional[Store] = None) -> bool:
+def new_downloads_only(candidate: Config, store: t.Optional[Store] = None) -> t.Iterator[Config]:
     """Predicate function to skip already downloaded partitions."""
     if store is None:
         store = FSStore()
     should_skip = skip_partition(candidate, store)
```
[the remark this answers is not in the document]
alxmrs: Thinking about this some more, I don't think my idea is a good approach. This is because we use this function later on to prevent double download. Since the state changes at that time, listing and caching will produce the wrong behavior.

alxmrs:
Is this CL ready for review? Are we waiting on EC to come back online before testing this? 

One idea for testing it now is to download an existing dataset, s.t. it skips all new downloads (and seeing how fast this runs).

weather_dl/download_pipeline/partition.py
```diff
@@ -149,7 +149,8 @@ def new_downloads_only(candidate: Config, store: t.Optional[Store] = None) -> t.
...
     if not should_skip:
         yield candidate
-    beam.metrics.Metrics.counter('Prepare', 'skipped').inc()
+    else:
+        beam.metrics.Metrics.counter('Prepare', 'skipped').inc()
```
alxmrs, line 153: Nice catch.

weather_dl/download_pipeline/partition.py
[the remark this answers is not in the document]
alxmrs: Since the thread approach may not be as fruitful as we originally expected, what if we try the following: We create two methods to check for skipping – `skip_partition` and `bulk_skip`. The latter function can take the file listing approach I mention above, and will only be used for the partition stage of the pipeline. Then, we can keep the existing method for its use downstream in the pipeline (double checking work during fetches).

alxmrs:
Given these docs (found from Dataflow's recommendations page on our pipeline), it looks like a reshuffle after the FlatMap may suffice to address the issue: https://cloud.google.com/dataflow/docs/guides/using-dataflow-insights?&_ga=2.108320322.-1550564198.1664912525#high-fan-out

weather_dl/download_pipeline/util.py
```diff
@@ -42,3 +44,17 @@ def retry_with_exponential_backoff(fun):
         retry_filter=_retry_if_valid_input_but_server_or_socket_error_and_timeout_filter,
         clock=clock,
     )(fun)
+
+
+# TODO(#245): Group with common utilities (duplicated)
+def ichunked(iterable: t.Iterable, n: int) -> t.Iterator[t.Iterable]:
```
shoyer, line 50: Where is this from? If you implemented it yourself, please add a basic test. If you copied or adapted it from another library, it needs to go in a `third_party` sub-directory and preserve the original license.

weather_dl/download_pipeline/partition.py
```diff
@@ -125,19 +133,45 @@ def skip_partition(config: Config, store: Store) -> bool:
...
     """
-    for option in itertools.product(*[config.selection[key] for key in config.partition_keys]):
+    dims = [range(len(config.selection[key])) for key in config.partition_keys]
+    n_partitions = math.prod([len(d) for d in dims])
+    logger.info(f'Creating {n_partitions} partitions.')
+
+    if n_chunks is None:
+        n_chunks = 50
+
+    chunk_size = n_partitions // n_chunks + 1
```
shoyer, line 159: From a usability perspective, I would consider reversing this logic: figure out the number of partitions from the desired chunk sizes. This would probably require a small refactor, but I'm pretty sure you always know how exactly how many configs are to be processed when creating a job?

For example, if each worker processes ~10 configs/second, and you don't want this stage to need to take more than a small number of minutes, you might pick a chunk size in the range of 1000 to 10000. Smaller is probably fine/better (given Beam's autoscaling), as long as overhead is minimized.

weather_dl/download_pipeline/util.py
alxmrs: This came from here: https://github.com/google/weather-tools/blob/06cbc22a931c44ef9e82f18082ed29a7fca1e31c/weather_mv/loader_pipeline/util.py#L153

It was developed in house. Tests live next to where the routine was originally developed. My reasoning for not copying the tests here also war primarily that I wanted to minimize duplicate code.

I wanted to remove the duplication and instead extract out a common module (#245), but that would complicate this CL. 

How about this for a resolution: I copy with tests over with the same TODO, and clean it up when we get to the issue?

weather_dl/download_pipeline/partition.py
alxmrs: I appreciate this feedback -- I was on the fence about which way this should go. 

As far as smaller is better: at a certain point, too small of chunks will be slower due to the reshuffle overhead. 1000 to 10_000 chunks seem to work well in practice.

weather_dl/download_pipeline/util.py
shoyer: Ah, gotcha. I missed this because I didn't read the linked GitHub issue!

You could also just add a brief comment like "duplicate from file XYZ, which has tests", but this is fine, too!

shoyer (approved):

PR #238, opened 2022-09-23
weather_dl tests pass locally w/o a CDS config in home folder.

alxmrs:
Updated the pipeline tests to use a fake key, so users can run them without setting up a CDS license.


mahrsee1997 (approved):
LGTM.

PR #239, opened 2022-09-23
Follow standard logging conventions for weather-sp.

### LINKED ISSUE #232: `weather-sp` produces no log output when run
The other weather tools log Beam traces when they are run. This is useful to know that the tool is working correctly. When run on Dataflow, for example, the log trace contains the URL for the job dashboard. 

To fix this issue, `weather-sp` should log the default Beam traces at the INFO level, like the other tools.

alxmrs:
Added INFO level logging for the beam pipeline in weather-sp.

Fixes #232.

### DIFF
--- a/weather_sp/splitter_pipeline/pipeline.py
+++ b/weather_sp/splitter_pipeline/pipeline.py
@@ -18,8 +18,8 @@ import os
 import typing as t
 
 import apache_beam as beam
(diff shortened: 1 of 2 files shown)


uhager (approved):

alxmrs:
Thanks for the late-in-the-day review, @uhager!

PR #240, opened 2022-09-23
Fixed potential memory leak during splitting.

alxmrs:
When we open a grib locally, we currently do not use a context manager to open it. This means that `close()` is never called on the open grib data, which would thus leak memory.

In addition to this fix, I've refactored a bit of pipeline code to use less variables.

### DIFF
--- a/weather_sp/splitter_pipeline/file_splitters.py
+++ b/weather_sp/splitter_pipeline/file_splitters.py
@@ -115,7 +115,8 @@ class GribSplitter(FileSplitter):
     @contextmanager
     def _open_grib_locally(self) -> t.Iterator[t.Iterator[pygrib.gribmessage]]:
         with self._copy_to_local_file() as local_file:
-            yield pygrib.open(local_file.name)
+            with pygrib.open(local_file.name) as gb:
+                yield gb
 
 
 class NetCdfSplitter(FileSplitter):
--- a/weather_sp/splitter_pipeline/file_splitters.py
+++ b/weather_sp/splitter_pipeline/file_splitters.py
@@ -93,22 +93,23 @@ class GribSplitter(FileSplitter):
             return
 
         with self._open_grib_locally() as grbs:
-            for grb in grbs:
-                splits = dict()
-                for dim in self.output_info.split_dims():
-                    try:
-                        splits[dim] = getattr(grb, dim)
-                    except RuntimeError:
-                        self.logger.error(
-                            'Variable not found in grib: %s', dim)
-                key = self.output_info.formatted_output_path(splits)
-                if key not in outputs:
-                    outputs[key] = FileSystems.create(key)
-                outputs[key].write(grb.tostring())
-                outputs[key].flush()
-
-            for out in outputs.values():
-                out.close()
+            try:
+                for grb in grbs:
+                    splits = dict()
+                    for dim in self.output_info.split_dims():
+                        try:
+                            splits[dim] = getattr(grb, dim)
+                        except RuntimeError:
+                            self.logger.error(
+                                'Variable not found in grib: %s', dim)
+                    key = self.output_info.formatted_output_path(splits)
+                    if key not in outputs:
+                        outputs[key] = FileSystems.create(key)
+                    outputs[key].write(grb.tostring())
+                    outputs[key].flush()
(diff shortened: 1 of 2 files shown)


alxmrs:
CC: @mahrsee1997 @deepgabani8 Hey! Check out [7ea07dc](https://github.com/google/weather-tools/pull/240/commits/7ea07dcbaec8a6d71d1d9bd4d77f60418194a098) – I think this pattern of closing an XArray dataset in a context manager could be useful to you two (from what we discussed about this morning).

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -93,29 +93,31 @@ def split_data(self) -> None:
...
-                            'Variable not found in grib: %s', dim)
-                key = self.output_info.formatted_output_path(splits)
-                if key not in outputs:
-                    outputs[key] = FileSystems.create(key)
-                outputs[key].write(grb.tostring())
-                outputs[key].flush()
-
-            for out in outputs.values():
-                out.close()
+            try:
```
uhager, line 96: why is this needed? is  there an indication that is raises an exception here and the output files are left dangling? Or is this just in case? I'd rather not add just-in-case try-blocks if they may not ever be needed.

alxmrs: This offers a guarantee that the `close()` operations are called during any possible error. I don't have evidence that this is a problem now, but it's prudent to include. If I didn't do this, I would refactor this section of code to use a `with` expression (when I tried this, I found that it caused changes of data). 

I wouldn't say that this is never needed: the `finally` block is called every time this method is run -- including when there are exceptions.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -93,29 +93,31 @@ def split_data(self) -> None:
...
+                    out.close()
             self.logger.info('split %s into %d files',
                              self.input_path, len(outputs))
 
     @contextmanager
     def _open_grib_locally(self) -> t.Iterator[t.Iterator[pygrib.gribmessage]]:
         with self._copy_to_local_file() as local_file:
-            yield pygrib.open(local_file.name)
+            with pygrib.open(local_file.name) as gb:
+                yield gb
```
alxmrs, line 120: This is the main grib memory leak fix.

weather_sp/splitter_pipeline/file_splitters.py
fredzyda: Using with here seems preferable, but I guess this works too.

weather_sp/splitter_pipeline/pipeline.py
```diff
@@ -117,35 +117,31 @@ def run(argv: t.List[str], save_main_session: bool = True):
 
     pipeline_options = PipelineOptions(pipeline_args)
     pipeline_options.view_as(SetupOptions).save_main_session = save_main_session
-    input_pattern = known_args.input_pattern
-    input_base_dir = _get_base_input_directory(input_pattern)
-    output_template = known_args.output_template
-    output_dir = known_args.output_dir
-    formatting = known_args.formatting
+    input_base_dir = _get_base_input_directory(known_args.input_pattern)
```
fredzyda, line 120: These changes seem somewhat unrelated to the memory leak...

fredzyda (approved):
Generally looks fine. I had a couple of nits, but nothing to block an important fix on.

weather_sp/splitter_pipeline/file_splitters.py
alxmrs: I just stepped into this code through the debugger. I now have a better understanding of what's going on. This pattern now makes sense to me: What I was missing before is that the `write` grib message step happens multiple times per output file. So, we would either have to a) keep a in-memory buffer of all the grib messages that will eventually get flused, then use a `with` expression at the end to write it once, or b) do what we're doing now. 

I'll leave a comment here that tries to explain this, for future reference.

weather_sp/splitter_pipeline/pipeline.py
alxmrs: Yes, that is right. I'm happy to revert them / move them to a different CL.

PR #242, opened 2022-09-26
GribSplitterV2: Splitting using `grib_copy`.

alxmrs:
Adding support for splitting grib files by multiple dimensions at once. This PR does this by delegating the heavy lifting to ecCodes' `grib_copy` utility.

TODO
- [ ] Let's ensure that the `grib_copy` utility is installed on the runtime image (the recent Docker changes should fix this).

### DIFF
--- a/weather_sp/README.md
+++ b/weather_sp/README.md
@@ -61,13 +61,13 @@ weather-sp --input-pattern 'gs://test-tmp/era5/2015/**' \
            --job_name $JOB_NAME
 ```
 
-Using ecCodes-powered grib splitting on Dataflow:
+Using ecCodes-powered grib splitting on Dataflow (this is often more robust, especially when splitting multiple 
+dimensions at once):
 
 ```bash
 weather-sp --input-pattern 'gs://test-tmp/era5/2017/**' \
            --output-dir 'gs://test-tmp/era5/splits' \
            --formatting '.{typeOfLevel}' \
-           --use-version v2 \
            --runner DataflowRunner \
            --project $PROJECT \
            --temp_location gs://$BUCKET/tmp  \
--- a/weather_sp/README.md
+++ b/weather_sp/README.md
@@ -12,7 +12,8 @@ Splits NetCDF and Grib files into several files by variable or other dimension. 
 ## Usage
 
 ```
-usage: weather-sp [-h] -i INPUT_PATTERN (--output-template OUTPUT_TEMPLATE | --output-dir OUTPUT_DIR) [--formatting FORMATTING] [-d] [-f]
+usage: weather-sp [-h] -i INPUT_PATTERN (--output-template OUTPUT_TEMPLATE | --output-dir OUTPUT_DIR)
+                  [--formatting FORMATTING] [-d] [-f] [--use-version USE_VERSION]
 ```
 
 _Common options_:
@@ -26,6 +27,8 @@ _Common options_:
    using Python formatting, see [Output section](#output) below.
 * `-f, --force`: Force re-splitting of the pipeline. Turns of skipping of already split data.
 * `-d, --dry-run`: Test the input file matching and the output file scheme without splitting.
+* `--use-version`: Choose splitting implementation version (which depends on the data type)
+  Default: v1.
 
 Invoke with `-h` or `--help` to see the full range of options.
 
@@ -33,7 +36,7 @@ _Usage examples_:
 
 ```bash
 weather-sp --input-pattern 'gs://test-tmp/era5/2017/**' \
-           --output-dir 'gs://test-tmp/era5/splits'
+           --output-dir 'gs://test-tmp/era5/splits' \
            --formatting '.{typeOfLevel}'
 ```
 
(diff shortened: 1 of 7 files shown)


weather_sp/README.md
```diff
@@ -26,14 +27,16 @@ _Common options_:
    using Python formatting, see [Output section](#output) below.
 * `-f, --force`: Force re-splitting of the pipeline. Turns of skipping of already split data.
 * `-d, --dry-run`: Test the input file matching and the output file scheme without splitting.
+* `--use-version`: Choose splitting implementation version (which depends on the data type)
```
uhager, line 30: This only applies to grib though? I think that should be somewhere in the name, e.g. --grib-splitter-version seems clearer to me.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -133,6 +135,40 @@ def _open_grib_locally(self) -> t.Iterator[t.Iterator[pygrib.gribmessage]]:
...
+
+        output_template = self.output_info.unformatted_output_path().replace('{', '[').replace('}', ']')
+        root, output_template = os.path.split(output_template)
+        with self._copy_to_local_file() as local_file:
+            with tempfile.TemporaryDirectory() as tmpdir:
+                dest = os.path.join(tmpdir, output_template)
+
+                cmd = shutil.which('grib_copy')
+                if not cmd:
+                    raise EnvironmentError('binary `grib_copy` is not available in the current environment!')
```
uhager, line 162: I think this check should already be performed earlier if possible.

weather_sp/README.md
```diff
@@ -51,12 +54,28 @@ Using DataflowRunner
...
            --output-dir 'gs://test-tmp/era5/splits'
-           --formatting '.{typeOfLevel}'
+           --formatting '.{typeOfLevel}' \
+           --runner DataflowRunner \
+           --project $PROJECT \
+           --temp_location gs://$BUCKET/tmp  \
+           --job_name $JOB_NAME
+```
+
+Using ecCodes-powered grib splitting on Dataflow:
```
uhager, line 64: There should be some comment here why this might be preferred.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -230,7 +267,13 @@ def get_splitter(file_path: str, output_info: OutFileInfo, dry_run: bool, force_
 
     if b'GRIB' in header:
         metrics.Metrics.counter('get_splitter', 'grib').inc()
-        return GribSplitter(file_path, output_info, force_split)
+
+        if '1' in use_version:
+            return GribSplitter(file_path, output_info, force_split)
+        elif '2' in use_version:
+            return GribSplitterV2(file_path, output_info, force_split)
```
uhager, line 274: Have you checked how much faster v2 ends up being? If it is always faster, instead of adding the --version flag, should we instead check here if grib_copy can be found and if so always prefer that? I.e. remove the `--use-version` flag, and replace the '2' check here with with the `shutil.which(grib_copy)` check.

alxmrs: I like this idea! Great shout.

alxmrs: It may be faster, but the more important issue is that it should provide a more robust way to split multiple dimensions of grib files at once.

weather_sp/README.md
alxmrs: Removing this flag based on your other suggestion.

weather_sp/splitter_pipeline/file_splitters.py
alxmrs: Great idea.

weather_sp/README.md
uhager: Should be removed from the README here as well? Or am I seeing an old version?

weather_sp/README.md
```diff
@@ -51,12 +54,28 @@ Using DataflowRunner
...
            --output-dir 'gs://test-tmp/era5/splits'
-           --formatting '.{typeOfLevel}'
+           --formatting '.{typeOfLevel}' \
+           --runner DataflowRunner \
+           --project $PROJECT \
+           --temp_location gs://$BUCKET/tmp  \
+           --job_name $JOB_NAME
+```
+
+Using ecCodes-powered grib splitting on Dataflow (this is often more robust, especially when splitting multiple
```
uhager, line 64: Can you add an explanation of what the `sdk_contrainer_image` needs to contain, e.g. eccodes, don't know if anything else.

alxmrs: Makes sense. Added.

weather_sp/README.md
uhager: Remove this.

weather_sp/README.md
```diff
@@ -51,12 +54,29 @@ Using DataflowRunner
...
+           --output-dir 'gs://test-tmp/era5/splits' \
+           --formatting '.{typeOfLevel}' \
            --runner DataflowRunner \
            --project $PROJECT \
            --temp_location gs://$BUCKET/tmp  \
+           --experiment=use_runner_v2 \
+           --sdk_container_image="gcr.io/$PROJECT/$REPO:latest"  \
            --job_name $JOB_NAME
 ```
+_Consult [this documentation](../docs/Runtime-Container.md) for steps on how to create an sufficient image._
```
uhager, line 76: nit: create _a_ sufficient image.

weather_sp/README.md
alxmrs: Thanks! I missed this.

uhager (approved):

weather_sp/README.md
alxmrs: Thanks!

PR #250, opened 2022-10-13
Skip "file prefix" validation while executing streaming job

mahrsee1997:
minor fix -- no need to validate file-prefix at this stage of pipeline for streaming jobs.


mahrsee1997:
Not mature enough solution -- this works in case of EE jobs but not for BQ. Need to give deep thought before landing any change.

alxmrs:
Ah, great catching this. I look forward to looking over the next iteration of the solution.

PR #254, opened 2022-10-14
Improvements to file copying in `weather-sp`.

alxmrs:
Setting file copy buffer sizes to use GCS-optimal constants. In addition, I've deleted a dead method.

### DIFF
--- a/weather_sp/splitter_pipeline/file_splitters.py
+++ b/weather_sp/splitter_pipeline/file_splitters.py
@@ -25,6 +25,7 @@ import numpy as np
 import pygrib
 import xarray as xr
 from apache_beam.io.filesystems import FileSystems
+from apache_beam.io.gcp.gcsio import DEFAULT_READ_BUFFER_SIZE, WRITE_CHUNK_SIZE
 
 from .file_name_utils import OutFileInfo
 
@@ -52,13 +53,9 @@ class FileSplitter(abc.ABC):
     def _copy_to_local_file(self) -> t.Iterator[t.IO]:
         with FileSystems().open(self.input_path) as source_file:
             with tempfile.NamedTemporaryFile() as dest_file:
-                shutil.copyfileobj(source_file, dest_file)
+                shutil.copyfileobj(source_file, dest_file, DEFAULT_READ_BUFFER_SIZE)
                 dest_file.flush()
                 yield dest_file
-
-    def _copy_dataset_to_storage(self, src_file: t.IO, target: str):
-        with FileSystems().create(target) as dest_file:
-            shutil.copyfileobj(src_file, dest_file)
 
     def should_skip(self):
         """Skip splitting if the data was already split."""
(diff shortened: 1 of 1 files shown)


weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -25,6 +25,7 @@
 import pygrib
 import xarray as xr
 from apache_beam.io.filesystems import FileSystems
+from apache_beam.io.gcp.gcsio import DEFAULT_READ_BUFFER_SIZE, WRITE_CHUNK_SIZE
```
uhager, line 28: I remember at some point we specifically changed all direct references to gcsio to use FileSystems instead to be completely filesystem-agnostic. So I'm a bit hesitant to bring this back in.
For the READ_BUFFER, can we use this instead: https://github.com/apache/beam/blob/45cc0854be553b5138753cf649fa863a80ff2b04/sdks/python/apache_beam/io/filesystem.py#L48 
which is the same size, just in a different file. For write_chunk, would this change depending on the filesystem?

alxmrs: You raise a really interesting point. This was an impetus for me to dig in deeper into what each of these constants mean. 

So, off the bat, there is not default WRITE_CHUNK for all filesystems. So, I inspected all the filesystem implementations constants to see how they varied (at least, a few major clouds): 
- https://beam.apache.org/releases/pydoc/2.19.0/_modules/apache_beam/io/aws/s3io.html
- https://beam.apache.org/releases/pydoc/2.29.0/_modules/apache_beam/io/azure/blobstorageio.html
- https://beam.apache.org/releases/pydoc/2.11.0/_modules/apache_beam/io/gcp/gcsio.html

Most telling, to me, was this comment explaining GCPs default numbers: https://github.com/apache/beam/blob/b77413309ef3032a6422428cc430eca67582ff61/sdks/python/apache_beam/io/gcp/gcsio.py#L73

It seems like these values are more a matter of opinion (or tradeoffs) than I thought! I have more confidence that this is the case reading the s3 sources, see these write range constants: https://github.com/apache/beam/blob/b77413309ef3032a6422428cc430eca67582ff61/sdks/python/apache_beam/io/aws/s3io.py#L614

Given all this, I think the best plan going forward is to choose our project-specific constants, given that we have a better idea of the typical file sizes being read and written.

uhager: I agree. I think for the reading default, we can go with the io.filesystem default, for the writing, we should likely pick something on the larger side, since our files tend to be large.

alxmrs: I seemed to have gotten into a rabbit hole. For now, I will use a set of a few good defaults and revisit the topic of optimal file transfer methods later.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -24,7 +24,9 @@
 import numpy as np
 import pygrib
 import xarray as xr
+from apache_beam.io.filesystem import DEFAULT_READ_BUFFER_SIZE
 from apache_beam.io.filesystems import FileSystems
+from apache_beam.io.gcp.gcsio import WRITE_CHUNK_SIZE
```
uhager, line 29: Can you add a comment or TODO or something that we don't want to bring in more gcs-specific dependencies and instead should clean this up?

alxmrs: ack

uhager (approved):

PR #263, opened 2022-11-24
Update docs and custom commands to prepare for Conda as the main release channel

blackvvine:


Dockerfile
```diff
@@ -0,0 +1,42 @@
+
```
alxmrs, line 1: nit: remove this empty line.

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+COPY environment.yml /tmp/environment.yml
+RUN conda env create -f /tmp/environment.yml
+
+# Activate the conda env and update the PATH
+ARG CONDA_ENV_NAME=weather-tools
+RUN echo "source activate ${CONDA_ENV_NAME}" >> ~/.bashrc
+ENV PATH /opt/conda/envs/${CONDA_ENV_NAME}/bin:$PATH
+
+# Install SDK.
+RUN pip install --no-cache-dir apache-beam[gcp]==2.40.0
```
alxmrs, line 33: I've recently found out that we should use `pip` managed from `conda` here, else we could get weird link issues. 

Can we `conda` install `apache-beam` instead?

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Verify that the image does not have conflicting dependencies.
+RUN pip check
```
alxmrs, line 36: similarly, we may want to `conda` check (whatever the equivalent cmd is).

README.md
```diff
@@ -33,15 +33,22 @@ their component variables.
...
+sources as follows:
 
   ```shell
-  pip install google-weather-tools
+conda env create --name weather-tools --file=environment.yml
+conda activate weather-tools
+pip install .
+pip install weather_dl/
+pip install weather_mv/
+pip install weather_sp/
```
alxmrs, line 46: Similar to the command above: the `pip install` command should be set in the `pip` section of the conda env.

README.md
```diff
@@ -33,15 +33,22 @@ their component variables.
...
   ```
 
-> Note: If the execution of `pip install` command takes a lot of time, try upgrading your pip version
-> (`pip install --upgrade pip`) or using the legacy resolver (add flag: `--use-deprecated=legacy-resolver`).
+> Note: Due to its use of 3rd-party binary dependencies such as GDAL and MetView, `weather-tools`
+> is transitioning from PyPi to Conda for its main release channel. The instructions above
+> are a temporary workaround before our Conda-forge release.
```
alxmrs, line 51: Perfect!

README.md
```diff
@@ -51,60 +58,77 @@ From here, you can use the `weather-*` tools from your python environment. Curre
...
+4. Create an empty BigQuery Dataset. This can be done using
+   the [Google Cloud Console](https://cloud.google.com/bigquery/docs/quickstarts/quickstart-cloud-console#create_a_dataset)
+   or via the [`bq` CLI tool](https://cloud.google.com/bigquery/docs/quickstarts/quickstart-command-line). 
+   For example:
    ```shell
-   bq mk --project_id=$PROJECT $DATASET_ID
+   bq mk --project_id=$PROJECT_ID $DATASET_ID
    ```
+6. Follow [these steps](https://cloud.google.com/storage/docs/creating-buckets) 
+   to create a bucket for staging temporary files in [Google Cloud Storage](https://cloud.google.com/storage).
```
alxmrs, line 87: Copy looks good :)

alxmrs (changes requested):
New copy looks really great! Thank you for doing that. 

I have some concerns around pip + conda usage (I've recently learned about). Let's resolve these in this PR.

Dockerfile
blackvvine: done!

Dockerfile
blackvvine: I haven't tried that yet but I don't suppose there would be any issues, I'll run the e2e tests today

Dockerfile
blackvvine: Does conda actually have an equivalent of `pip check`? I did a quick search but couldn't find anything. One alternative would be `/path/to/condas/pip check`

blackvvine: IMO this wouldn't even be necessary when the deps are resolved by Conda, is that correct?

alxmrs: You're totally right. We can just get rid of this line (conda will use its resolver).

README.md
blackvvine: done

README.md
```diff
@@ -33,15 +33,18 @@ their component variables.
...
-It's recommended that you create a local python environment (with
-[Anaconda](https://www.anaconda.com/products/individual)). Otherwise, these tools can be installed with pip:
+It is currently recommended that you create a local python environment (with
+[Anaconda](https://www.anaconda.com/products/individual)) and install the
+sources as follows:
 
   ```shell
-  pip install google-weather-tools
+conda env create --name weather-tools --file=environment.yml
+conda activate weather-tools
```
alxmrs, line 42: Nice!

environment.yml
```diff
@@ -33,3 +33,7 @@ dependencies:
   - pip:
     - earthengine-api==0.1.329
     - .[test]
+    - .
+    - ./weather_dl
+    - ./weather_mv
+    - ./weather_sp
```
alxmrs, line 39: Excellent :)

weather_mv/setup.py
```diff
@@ -135,8 +126,7 @@ def RunCustomCommand(self, command_list):
                 'Command %s failed: exit code: %s' % (command_list, p.returncode))
 
     def run(self):
-        for command in CUSTOM_COMMANDS:
-            self.RunCustomCommand(command)
+        pass
```
alxmrs, line 129: No need to do this now, but I think we could delete the whole `CustomCommands` bit of code.

Dockerfile
blackvvine: done

weather_mv/setup.py
blackvvine: on it

Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+# Update miniconda
+RUN conda update conda -y
+
+# Create conda env using environment.yml
+ARG weather_tools_git_rev=main
+RUN git clone https://github.com/google/weather-tools.git /weather
+WORKDIR /weather
+RUN git checkout "${weather_tools_git_rev}"
+RUN conda env create -f environment.yml
```
alxmrs, line 28: I'm curious why we should clone the repo in the image vs. mapping the repo from the host machine into the Docker image. What's the advantage to doing it this way? If we did the latter, could this be simpler?

weather_mv/setup.py
blackvvine: done

Dockerfile
blackvvine: That was how I originally wrote it, but the idea here is that `COPY . /weather` would cause any changes to the directory to undo all caching (i.e. the directory hash will change so it'll have to do the `conda env create` which is the slowest step from scratch)

blackvvine: This is also more consistent with how the Conda recipe is built, as it will read from a specific branch/tag

blackvvine: One alternative is just copying `weather_*` directories and `setup.py`, but it's not super elegant and makes it easier to break with changes in code structure

alxmrs: Thanks for the explanation. This makes sense, and am happy with the existing implementation.

alxmrs (approved):
Looks great! Thank you so much for these changes!!

Issue #256, opened 2022-10-17
Create a performant, cloud-agnostic way to download & upload files to cloud buckets. 


alxmrs:
See discussion here: https://github.com/google/weather-tools/pull/254#discussion_r996220532

To investigate: 
- Can we do better than `shutil.copyfileobj`?
- What are the optimal chunk sizes?
- Can we copy data in parallel?
- Are there optimizations we can do for large files? 

One idea that @bahmandar has explored is calling gsutils in a subprocess (the CLI is really efficient at file transfer).

bahmandar:
Here is the location for using gcloud for gcs files:
https://github.com/bahmandar/weather-tools/blob/mv-faster/weather_mv/loader_pipeline/sinks.py#L416

Fall back is shutil for downloading and fall back for remote is using apache beam file systems.

I also have a different shutil optimized for gcs:
https://github.com/bahmandar/weather-tools/blob/mv-faster/weather_mv/loader_pipeline/sinks.py#L402
One thing to note it seems like it is beneficial to change the buffer for gcs io and shutil together than just one of them.

alxmrs:
Here are some advantages of just using `gsutil` vs a hand-rolled python solution: 
- over a size threshold, gsutil will automatically parallelize file transfer
- gsutil uses checksums to verify the integrity of data transferred, and will automatically retry on corrupted data
- the default dataflow image already has gcloud installed, so in theory it's easy to manage this dependency
- we get to make use of heavily invested code (maybe some magic constants found from trial & error) from GCP + boto devs
- we get all these features with a slick one-liner: `subprocess.run(f'gsutil cp {src!r} {dst!r}', shell=True, check=True)`

alxmrs:
Thanks @mahrsee1997 for pointing this out! https://cloud.google.com/blog/products/storage-data-transfer/new-gcloud-storage-enables-super-fast-data-transfers/

alxmrs:
@mahrsee1997 did some benchmarking of different cloud utilities to see what would be the fastest. Our results show that `gsutil` is the best fit for us. However! – It possible that `gcloud alpha storage` would be faster if we upgraded to the latest version of `gcloud`. Using this version of the SDK requires that we update the versions of all our GCP dependencies. This is something that we'll tackle, but in a future PR. #265 is still a great win. 

> I did the bench-marking on a file of size ~18.42 GiB & it appears that "gsutil" is the most efficient approach here. It's ~77% reduction in time than our original approach of shutil.
> 
[...]

PR #265, opened 2022-12-02
weather-dl now uses `gsutil cp` for file upload.

alxmrs:
A partial implementation of #256. My intention here is to see if this can speed up weather-dl requests.

### DIFF
--- a/weather_dl/download_pipeline/util.py
+++ b/weather_dl/download_pipeline/util.py
@@ -12,12 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import itertools
+import logging
 import socket
 import subprocess
 import sys
 import typing as t
 
 from apache_beam.utils import retry
+
+logger = logging.getLogger(__name__)
 
 
 def _retry_if_valid_input_but_server_or_socket_error_and_timeout_filter(exception) -> bool:
--- a/weather_dl/download_pipeline/util.py
+++ b/weather_dl/download_pipeline/util.py
@@ -67,4 +67,8 @@ def copy(src: str, dst: str) -> None:
     Here, we take advantage of GCP's faster storage transfer:
     https://cloud.google.com/blog/products/storage-data-transfer/new-gcloud-storage-enables-super-fast-data-transfers/
     """
-    subprocess.run(f'gcloud alpha storage cp {src!r} {dst!r}', shell=True, check=True)
+    try:
+        subprocess.run(f'gcloud alpha storage cp {src!r} {dst!r}', shell=True, check=True, capture_output=True)
+    except subprocess.CalledProcessError as e:
+        logger.error(f'Failed to copy file {src!r} to {dst!r} due to {e.stderr.decode("utf-8")}')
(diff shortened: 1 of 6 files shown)


weather_dl/download_pipeline/util.py
```diff
@@ -58,3 +59,12 @@ def ichunked(iterable: t.Iterable, n: int) -> t.Iterator[t.Iterable]:
...
         pass
+
+
+def copy(src: str, dst: str) -> None:
+    """Copy data via `gcloud storage cp`.
+
+    Here, we take advantage of GCP's faster storage transfer:
+    https://cloud.google.com/blog/products/storage-data-transfer/new-gcloud-storage-enables-super-fast-data-transfers/
+    """
+    subprocess.run(f'gcloud alpha storage cp {src!r} {dst!r}', shell=True, check=True)
```
mahrsee1997, line 70: I suggest adding some error handling here.
```suggestion
    try:
        subprocess.run(f'gcloud alpha storage cp {src!r} {dst!r}', shell=True, check=True, capture_output=True)
    except subprocess.CalledProcessError as e:
        logger.error(f'Failed to copy file {src!r} to {dst!r} due to {e.stderr.decode("utf-8")}')
        raise
```

mahrsee1997: Also should we also add retrying logic here or should we fail the complete pipeline if we are not able to upload a file ? WDYT ?

mahrsee1997 (approved):
Overall PR LGTM. I'm happy to get it merged after resolving the above comments & fixing the tests.

Kudos !! This is a noteworthy change.

weather_dl/download_pipeline/util.py
alxmrs: Great thought! From the docs, it looks like `gcloud`/`gsutil` internally handles retries: https://cloud.google.com/storage/docs/gsutil/commands/cp#retry-handling

In my mind, when things fail here, they do so for a good reason. 

I like your log suggestions, I will adopt that :)

weather_dl/download_pipeline/util.py
```diff
@@ -58,3 +62,16 @@ def ichunked(iterable: t.Iterable, n: int) -> t.Iterator[t.Iterable]:
...
+
+
+def copy(src: str, dst: str) -> None:
+    """Copy data via `gcloud storage cp`.
+
+    Here, we take advantage of GCP's faster storage transfer:
+    https://cloud.google.com/blog/products/storage-data-transfer/new-gcloud-storage-enables-super-fast-data-transfers/
+    """
+    try:
+        subprocess.run(f'gcloud alpha storage cp {src!r} {dst!r}', shell=True, check=True, capture_output=True)
```
shoyer, line 74: Generally it's best to avoid `shell=True` for [security considerations](https://docs.python.org/3/library/subprocess.html#security-considerations):
```suggestion
        subprocess.run(['gcloud', 'alpha', 'storage', 'cp', src, dest], check=True, capture_output=True)
```

PR #269, opened 2022-12-09
Updating docs + the conda env for a better runtime container.

alxmrs:
This CL primarily fixes a version conflict issue in our environment file dependencies. This env file will now be oriented towards the runtime docker container (a prod env) rather than for local development. 

In addition, I fix a few documentation issues.


environment.yml
```diff
@@ -4,37 +4,28 @@ channels:
 dependencies:
   - python=3.8.13
   - apache-beam=2.40.0
-  - cdsapi=0.5.1
-  - pytest=7.2.0
-  - pytest-subtests=0.8.0
-  - cfgrib=0.9.10.2
-  - dask=2022.10.0
```
deepgabani8, line 11: We are using `dask` in weather-mv/regrid. Are we sure to remove it from here?

deepgabani8: Checked this. `xarray-beam==0.3.1` has `dask` in it's requirements and it will install `dask==2022.12.0`.
It is safe to remove this from here. 👍

alxmrs: Thanks for checking, Deep! :)

PR #273, opened 2023-01-03
weather-mv: Do convert to asset in a sub-process.

deepgabani8:


weather_mv/loader_pipeline/ee.py
```diff
@@ -458,6 +458,17 @@ def process(self, uri: str) -> t.Iterator[AssetData]:
                 properties=attrs
             )
 
+            queue.put_nowait(asset_data)
+
+    def process(self, uri: str) -> t.Iterator[AssetData]:
+        """Opens grib files and yields AssetData."""
+        queue = Queue(maxsize=2)
```
alxmrs, line 465: Why is the maxsize 2 here? Can you leave a comment with the explanation?

alxmrs: I imagine the comment might explain how this is used to perform the work in a subprocess.

alxmrs (approved):
LGTM, though I wouldn't mind having @fredzyda TAL.

weather_mv/loader_pipeline/ee.py
```diff
@@ -467,7 +467,7 @@ def process(self, uri: str) -> t.Iterator[AssetData]:
         process.start()
         process.join()
 
-        if not queue.empty():
+        while not queue.empty():
```
alxmrs, line 470: +1

fredzyda (approved):
This looks totally reasonable to me.

PR #275, opened 2023-01-05
Preserve dataset level attributes & maintain units for each data variable in dataset level attributes.

mahrsee1997:


weather_mv/loader_pipeline/ee.py
```diff
@@ -405,6 +405,8 @@ def process(self, uri: str) -> t.Iterator[AssetData]:
                                                    ('start_time', 'end_time', 'is_normalized'))
             dtype, crs, transform = (attrs.pop(key) for key in ['dtype', 'crs', 'transform'])
             attrs.update({'is_normalized': str(is_normalized)})  # EE properties does not support bool.
+            # Make attrs EE ingestable.
+            attrs = comb_attributes(attrs)
```
alxmrs, line 409: optional: could we call this `parse_attributes`?

alxmrs: do we use `attrs`? Below, it looks like we use the attributes on `da`.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -177,14 +179,28 @@ def __normalize_grib_dataset(filename: str) -> xr.Dataset:
...
                 copied_da = copied_da.drop_vars(level)
                 _data_array_list.append(copied_da)
 
-    return xr.merge(_data_array_list)
+    # Stick the forecast hour, start_time, end_time, data variables units
+    # in the ds attrs as well, that's useful.
+    ds_attrs['forecast_hour'] = _data_array_list[0].attrs['forecast_hour']
+    ds_attrs['start_time'] = _data_array_list[0].attrs['start_time']
+    ds_attrs['end_time'] = _data_array_list[0].attrs['end_time']
+    ds_attrs.update(**dv_units_dict)
```
alxmrs, line 198: Nice, +1

weather_mv/loader_pipeline/ee.py
mahrsee1997: Yes, we use it refer to [Line: 432](https://github.com/google/weather-tools/pull/275/files#diff-5ddebac011830eb7d97f9a4badf76f3a8f2990918c89b70150b585fe3df0a527R432).

Basically `ds.attrs` are used for setting tiff/dataset level metadata were as `da.attrs` are used for setting band level metadata.

mahrsee1997: umm should we rename it to `make_attrs_ee_compatible` -- this makes more sense as per what method is actually doing. WDYT ?

deepgabani8 (approved):

PR #278, opened 2023-01-06
Adding the mamba solver to our Dockerimage installation with Conda

alxmrs:
This change does two things: 
- We pin all the `pip` dependency versions
- We add the `mamba` solver as the default solver to `conda` during our Docker installation process. 

Together (especially the latter change), we avoid an issue where cloud build fails due to a `conda` timeout during dependency resolution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,10 @@ FROM continuumio/miniconda3:latest
 # Update miniconda
 RUN conda update conda -y
 
+# Add the mamba solver for faster builds
+RUN conda install -n base conda-libmamba-solver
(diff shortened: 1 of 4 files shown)


blackvvine (commented):
I tested the build and it runs successfully, however pinning the package versions seems to have a <5% impact on build time overall. Are there any other reasons to do so?

alxmrs:
I think pinning the pip versions will also help improve the stability of our project (we won't experience breakages related to downstream deps).

blackvvine (approved):
That makes sense, the image can now be rolled out from with fixed dep versions.

PR #282, opened 2023-01-12
Fix issues while building docker image 

mahrsee1997:


alxmrs (approved):
Thanks -- LGTM

I seem to have confused by Conda vs Pip syntax for versions when I pinned them. Thanks for fixing this, Rahul.

PR #285, opened 2023-01-26
weather-sp fixes: better splitting by dates + improvements in skipping. 

alxmrs:


weather_sp/splitter_pipeline/pipeline.py
```diff
@@ -113,7 +115,7 @@ def run(argv: t.List[str], save_main_session: bool = True):
                         help='Force re-splitting of the pipeline. Turns of skipping of already split data.')
     known_args, pipeline_args = parser.parse_known_args(argv[1:])
 
-    configure_logger(2)  # 0 = error, 1 = warn, 2 = info, 3 = debug
+    configure_logger(3)  # 0 = error, 1 = warn, 2 = info, 3 = debug
```
fredzyda, line 118: Do you want to leave logging in debug by default? This seems like an error...

fredzyda (approved):
One nit about debug output being left on, but otherwise fine.

weather_sp/splitter_pipeline/pipeline.py
alxmrs: Thanks, I'll revert this.

PR #290, opened 2023-02-02
Prevent file upload in weather-dl dry-run mode.

### LINKED ISSUE #289: Running weather-dl in dry run mode might corrupt the actual data.
## Context: 
Running a weather-dl job with the actual config in dry run mode. And some files are already there in the target path.

#### Stage A (Fetch->Download Stage):
- Actual fetching & downloading of partition chunks will not take place as in dry run mode it uses FakeClient which just logs the message "Fetching ...." & "Downloading..." .

#### Stage B (Upload Stage):
- There is no check in the upload stage for dry run mode hence it will corrupt the actual data by uploading an empty temp file to the targeted path.

Furthermore, dry run mode uses TempFileStore() instead of FSStore() hence skipping logic will not work as well.

mahrsee1997:
Fixes #289.


ksic8:
We discovered this after some files got deleted from GCP bucket.

PR #292, opened 2023-02-08
While passing --check_skip_in_dry_run, avoid setting force_download to True in dry-run mode.

mahrsee1997:


alxmrs (approved):
This LGTM. However, given the meeting we had earlier today, I think we need to get more alignment on what the --dry-run feature is supposed to do.

mahrsee1997:
Totally agreed.

Issue #246, opened 2022-10-07
Implement default (console) manifest


blackvvine:
BigQuery, GCS, Cloud Monitoring and Firebase all have downsides and set up overhead and are unfit to be the default option for the manifest. 

A manifest that periodically dumps the download progress and status to console (which would in turn show up in Cloud Logging for Dataflow jobs) has the benefit of zero set up overhead (IAM or configurations), and should be the default option.

PR #295, opened 2023-02-09
New manifest structure and added support for BQ manifest.

### LINKED ISSUE #149: Update manifest to track pipeline stages.
In order to assist with solving #139, manifests need to be restructured to update download state. This is helpful when work is split up across a few transforms. Currently `DownloadStatus` only records the state of the whole download. 

https://github.com/google/weather-tools/blob/c574e5435a6f463d8ec47ad4a84fa8cc3f181b6e/weather_dl/download_pipeline/manifest.py#L54

After discussing with @mahrsee1997, we propose adding a `stage` flag to the `DownloadStatus` object. This will keep track of the Transform that the operation is currently in. Each stage can have `in-progress`, `success` or `failure` statuses. 

This will eventually help the user better understand the fine-grained state of the pipeline. Further, it can better illustrate what went wrong in the pipeline and were. 

## Implementation Notes

In addition to updating the `DownloadStatus` schema, the abstract `Manifest` also needs to be updated. A concise way to allow manifest users to express the stage would be to add an argument to the `transact` method:

https://github.com/google/weather-tools/blob/c574e5435a6f463d8ec47ad4a84fa8cc3f181b6e/weather_dl/download_pipeline/manifest.py#L178

After the change, the API should allow the manifest user to specify what stage the pipeline was on. Then, the manifest should update the `status`es and the `stage`s in the normal method, via a context manager (i.e. a `with` statement). 

_Advanced note_: Since the Beam stage name comes from a predicable place, it might be worth investigating if the manifest object can acquire the name of the Transform without having it be passed in as a string (is there a `__name__` that can be grabbed from some exposed object)? This is optional, of course. We can omit this if the approach is too complex.

mahrsee1997:
Fixes #149 and #60.

### DIFF
--- a/weather_dl/README.md
+++ b/weather_dl/README.md
@@ -44,8 +44,11 @@ _Common options_:
   and `--direct_num_workers`
 * `-m, --manifest-location MANIFEST_LOCATION`: Location of the manifest. By default, it will use Cloud Logging
   (stdout for direct runner). You can set the name of the manifest as the hostname of a URL with the 'cli' protocol.
-  For example, 'cli://manifest' will prefix all the manifest logs as '[manifest]'. In addition, users can specify a GCS
-  bucket URI, or 'noop://<name>' for an in-memory location.
+  For example, `cli://manifest` will prefix all the manifest logs as '[manifest]'. In addition, users can specify 
+  either a BigQuery table (`bq://<project-id>.<dataset-name>.<table-name>`) or GCS bucket URI, or `noop://<name>` 
+  for an in-memory location.
+> Note: Tool will create the BQ table itself, if not already present. 
+  Or it will use the existing table but can report errors in case of schema mismatch.
 * `-n, --num-requests-per-key`: Number of concurrent requests to make per API key. Default: make an educated guess per
   client & config. Please see the client documentation for more details.
 * `-p, --partition-chunks`: Group shards into chunks of this size when computing the partitions. Specifically, this 
--- a/weather_dl/README.md
+++ b/weather_dl/README.md
@@ -47,8 +47,6 @@ _Common options_:
   For example, `cli://manifest` will prefix all the manifest logs as '[manifest]'. In addition, users can specify 
   either a BigQuery table (`bq://<project-id>.<dataset-name>.<table-name>`) or GCS bucket URI, or `noop://<name>` 
   for an in-memory location.
-> Note: Tool will create the BQ table itself, if not already present. 
-  Or it will use the existing table but can report errors in case of schema mismatch.
 * `-n, --num-requests-per-key`: Number of concurrent requests to make per API key. Default: make an educated guess per
   client & config. Please see the client documentation for more details.
 * `-p, --partition-chunks`: Group shards into chunks of this size when computing the partitions. Specifically, this 
@@ -58,6 +56,9 @@ _Common options_:
   that partitions will be processed in sequential order of each config; 'fair' means that partitions from each config 
   will be interspersed evenly. Note: When using 'fair' scheduling, we recommend you set the '--partition-chunks' to a 
   much smaller number. Default: 'in-order'.
+
+> Note: In case of BigQuery manifest tool will create the BQ table itself, if not already present. 
+  Or it will use the existing table but can report errors in case of schema mismatch.
 
 Invoke with `-h` or `--help` to see the full range of options.
 
--- a/weather_dl/README.md
+++ b/weather_dl/README.md
@@ -45,7 +45,7 @@ _Common options_:
 * `-m, --manifest-location MANIFEST_LOCATION`: Location of the manifest. By default, it will use Cloud Logging
   (stdout for direct runner). You can set the name of the manifest as the hostname of a URL with the 'cli' protocol.
   For example, `cli://manifest` will prefix all the manifest logs as '[manifest]'. In addition, users can specify 
-  either a BigQuery table (`bq://<project-id>.<dataset-name>.<table-name>`) or GCS bucket URI, or `noop://<name>` 
+  either a BigQuery table (`bq://<project-id>.<dataset-name>.<table-name>`), or `noop://<name>` 
   for an in-memory location.
 * `-n, --num-requests-per-key`: Number of concurrent requests to make per API key. Default: make an educated guess per
   client & config. Please see the client documentation for more details.
--- a/weather_dl/download_pipeline/clients.py
+++ b/weather_dl/download_pipeline/clients.py
@@ -328,7 +328,7 @@ class MarsClient(Client):
 
 class ECMWFPublicClient(Client):
     """A client for ECMWF's public datasets, like TIGGE."""
-    def retrieve(self, dataset: str, selection: t.Dict, output: str) -> None:
+    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest) -> None:
         c = PublicECMWFServerExtended(
             url=self.config.kwargs.get('api_url', os.environ.get("MARSAPI_URL")),
             key=self.config.kwargs.get('api_key', os.environ.get("MARSAPI_KEY")),
@@ -339,7 +339,21 @@ class ECMWFPublicClient(Client):
         )
         selection_ = optimize_selection_partition(selection)
         with StdoutLogger(self.logger, level=logging.DEBUG):
+            manifest.set_stage(Stage.FETCH)
+            precise_fetch_start_time = (
+                datetime.datetime.utcnow()
+                .replace(tzinfo=datetime.timezone.utc)
+                .isoformat(timespec='seconds')
+            )
+            manifest.prev_stage_precise_start_time = precise_fetch_start_time
             result = c.fetch(req=selection_)
+            manifest.set_stage(Stage.DOWNLOAD)
+            precise_download_start_time = (
+                datetime.datetime.utcnow()
+                .replace(tzinfo=datetime.timezone.utc)
+                .isoformat(timespec='seconds')
+            )
+            manifest.prev_stage_precise_start_time = precise_download_start_time
             c.download(result, target=output)
 
     @classmethod
--- a/weather_dl/download_pipeline/clients.py
+++ b/weather_dl/download_pipeline/clients.py
@@ -107,7 +107,8 @@ class CdsClient(Client):
 
     def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest, target: str) -> None:
         selection_ = optimize_selection_partition(selection)
-        with manifest.transact(self.config.selection, target, self.config.user_id, 'RETRIEVE'):
+        with manifest.transact(self.config.kwargs.get('config_name'), self.config.selection,
+                               target, self.config.user_id, 'RETRIEVE'):
             self.c.retrieve(dataset, selection_, output)
 
     @property
@@ -269,9 +270,11 @@ class MarsClient(Client):
         )
         selection_ = optimize_selection_partition(selection)
         with StdoutLogger(self.logger, level=logging.DEBUG):
-            with manifest.transact(self.config.selection, target, self.config.user_id, 'FETCH'):
+            with manifest.transact(self.config.kwargs.get('config_name'), self.config.selection,
+                                   target, self.config.user_id, 'FETCH'):
                 result = self.c.fetch(req=selection_)
-            with manifest.transact(self.config.selection, target, self.config.user_id, 'DOWNLOAD'):
+            with manifest.transact(self.config.kwargs.get('config_name'), self.config.selection,
+                                   target, self.config.user_id, 'DOWNLOAD'):
                 self.c.download(result, target=output)
 
     @property
@@ -298,7 +301,8 @@ class FakeClient(Client):
     """A client that writes the selection arguments to the output file."""
 
     def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest, target: str) -> None:
-        with manifest.transact(self.config.selection, target, self.config.user_id, 'RETRIEVE'):
+        with manifest.transact(self.config.kwargs.get('config_name'), self.config.selection,
+                               target, self.config.user_id, 'RETRIEVE'):
             self.logger.debug(f'Downloading {dataset} to {output}')
             with open(output, 'w') as f:
                 json.dump({dataset: selection}, f)
--- a/weather_dl/download_pipeline/clients.py
+++ b/weather_dl/download_pipeline/clients.py
@@ -30,7 +30,7 @@ import urllib3
 from ecmwfapi import ECMWFService, api
 
 from .config import Config, optimize_selection_partition
-from .manifest import Manifest
+from .manifest import Manifest, Stage
 from .util import retry_with_exponential_backoff
 
 warnings.simplefilter(
@@ -55,7 +55,7 @@ class Client(abc.ABC):
         self.logger.setLevel(level)
 
     @abc.abstractmethod
-    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest, target: str) -> None:
+    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest) -> None:
         """Download from data source."""
         pass
 
@@ -105,11 +105,10 @@ class CdsClient(Client):
             error_callback=self.logger.error,
         )
 
-    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest, target: str) -> None:
+    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest) -> None:
         selection_ = optimize_selection_partition(selection)
-        with manifest.transact(self.config.kwargs.get('config_name'), self.config.selection,
-                               target, self.config.user_id, 'RETRIEVE'):
-            self.c.retrieve(dataset, selection_, output)
+        manifest.set_stage(Stage.RETRIEVE)
+        self.c.retrieve(dataset, selection_, output)
 
     @property
     def license_url(self):
@@ -259,7 +258,7 @@ class MarsClient(Client):
     def __init__(self, config: Config, level: int = logging.INFO) -> None:
         super().__init__(config, level)
 
-    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest, target: str) -> None:
+    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest) -> None:
         self.c = MARSECMWFServiceExtended(
             "mars",
             key=self.config.kwargs.get('api_key', os.environ.get("MARSAPI_KEY")),
@@ -270,12 +269,10 @@ class MarsClient(Client):
         )
         selection_ = optimize_selection_partition(selection)
         with StdoutLogger(self.logger, level=logging.DEBUG):
-            with manifest.transact(self.config.kwargs.get('config_name'), self.config.selection,
-                                   target, self.config.user_id, 'FETCH'):
-                result = self.c.fetch(req=selection_)
-            with manifest.transact(self.config.kwargs.get('config_name'), self.config.selection,
-                                   target, self.config.user_id, 'DOWNLOAD'):
-                self.c.download(result, target=output)
+            manifest.set_stage(Stage.FETCH)
+            result = self.c.fetch(req=selection_)
+            manifest.set_stage(Stage.DOWNLOAD)
+            self.c.download(result, target=output)
 
     @property
     def license_url(self):
@@ -300,12 +297,11 @@ class MarsClient(Client):
 class FakeClient(Client):
     """A client that writes the selection arguments to the output file."""
 
-    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest, target: str) -> None:
-        with manifest.transact(self.config.kwargs.get('config_name'), self.config.selection,
-                               target, self.config.user_id, 'RETRIEVE'):
-            self.logger.debug(f'Downloading {dataset} to {output}')
-            with open(output, 'w') as f:
-                json.dump({dataset: selection}, f)
+    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest) -> None:
+        manifest.set_stage(Stage.RETRIEVE)
+        self.logger.debug(f'Downloading {dataset} to {output}')
+        with open(output, 'w') as f:
+            json.dump({dataset: selection}, f)
 
     @property
     def license_url(self):
--- a/weather_dl/download_pipeline/clients.py
+++ b/weather_dl/download_pipeline/clients.py
@@ -16,6 +16,7 @@
 import abc
 import collections
 import contextlib
+import datetime
 import io
 import json
 import logging
@@ -108,6 +109,12 @@ class CdsClient(Client):
     def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest) -> None:
         selection_ = optimize_selection_partition(selection)
         manifest.set_stage(Stage.RETRIEVE)
+        precise_retrieve_start_time = (
+            datetime.datetime.utcnow()
+            .replace(tzinfo=datetime.timezone.utc)
+            .isoformat(timespec='seconds')
+        )
+        manifest.prev_stage_precise_start_time = precise_retrieve_start_time
         self.c.retrieve(dataset, selection_, output)
 
     @property
@@ -270,8 +277,20 @@ class MarsClient(Client):
         selection_ = optimize_selection_partition(selection)
         with StdoutLogger(self.logger, level=logging.DEBUG):
             manifest.set_stage(Stage.FETCH)
+            precise_fetch_start_time = (
+                datetime.datetime.utcnow()
+                .replace(tzinfo=datetime.timezone.utc)
+                .isoformat(timespec='seconds')
+            )
+            manifest.prev_stage_precise_start_time = precise_fetch_start_time
             result = self.c.fetch(req=selection_)
             manifest.set_stage(Stage.DOWNLOAD)
+            precise_download_start_time = (
+                datetime.datetime.utcnow()
+                .replace(tzinfo=datetime.timezone.utc)
+                .isoformat(timespec='seconds')
+            )
+            manifest.prev_stage_precise_start_time = precise_download_start_time
             self.c.download(result, target=output)
 
     @property
@@ -299,6 +318,12 @@ class FakeClient(Client):
 
     def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest) -> None:
         manifest.set_stage(Stage.RETRIEVE)
+        precise_retrieve_start_time = (
+            datetime.datetime.utcnow()
+            .replace(tzinfo=datetime.timezone.utc)
+            .isoformat(timespec='seconds')
+        )
+        manifest.prev_stage_precise_start_time = precise_retrieve_start_time
         self.logger.debug(f'Downloading {dataset} to {output}')
         with open(output, 'w') as f:
             json.dump({dataset: selection}, f)
--- a/weather_dl/download_pipeline/clients.py
+++ b/weather_dl/download_pipeline/clients.py
@@ -30,7 +30,7 @@ import urllib3
 from ecmwfapi import ECMWFService, api
 
 from .config import Config, optimize_selection_partition
-from .manifest import MANIFESTS, Manifest, Location, NoOpManifest
+from .manifest import Manifest
 from .util import retry_with_exponential_backoff
 
 warnings.simplefilter(
@@ -48,15 +48,14 @@ class Client(abc.ABC):
         level: Default log level for the client.
     """
 
-    def __init__(self, config: Config, manifest_obj: Manifest, level: int = logging.INFO) -> None:
+    def __init__(self, config: Config, level: int = logging.INFO) -> None:
         """Clients are initialized with the general CLI configuration."""
         self.config = config
-        self.manifest_obj = manifest_obj
         self.logger = logging.getLogger(f'{__name__}.{type(self).__name__}')
         self.logger.setLevel(level)
 
     @abc.abstractmethod
-    def retrieve(self, dataset: str, selection: t.Dict, output: str, target: str) -> None:
+    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest, target: str) -> None:
         """Download from data source."""
         pass
 
@@ -95,8 +94,8 @@ class CdsClient(Client):
     """Name patterns of datasets that are hosted internally on CDS servers."""
     cds_hosted_datasets = {'reanalysis-era'}
 
-    def __init__(self, config: Config, manifest: t.Dict[str, Location], level: int = logging.INFO) -> None:
-        super().__init__(config, MANIFESTS.get(manifest['type'], NoOpManifest)(manifest['location']), level)
+    def __init__(self, config: Config, level: int = logging.INFO) -> None:
+        super().__init__(config, level)
         self.c = cdsapi.Client(
             url=config.kwargs.get('api_url', os.environ.get('CDSAPI_URL')),
             key=config.kwargs.get('api_key', os.environ.get('CDSAPI_KEY')),
@@ -106,9 +105,9 @@ class CdsClient(Client):
             error_callback=self.logger.error,
         )
 
-    def retrieve(self, dataset: str, selection: t.Dict, output: str, target: str) -> None:
+    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest, target: str) -> None:
         selection_ = optimize_selection_partition(selection)
-        with self.manifest_obj.transact(self.config.selection, target, self.config.user_id, 'RETRIEVE'):
+        with manifest.transact(self.config.selection, target, self.config.user_id, 'RETRIEVE'):
             self.c.retrieve(dataset, selection_, output)
 
     @property
@@ -256,10 +255,10 @@ class MarsClient(Client):
         level: Default log level for the client.
     """
 
-    def __init__(self, config: Config, manifest: t.Dict[str, Location], level: int = logging.INFO) -> None:
-        super().__init__(config, MANIFESTS.get(manifest['type'], NoOpManifest)(manifest['location']), level)
+    def __init__(self, config: Config, level: int = logging.INFO) -> None:
+        super().__init__(config, level)
 
-    def retrieve(self, dataset: str, selection: t.Dict, output: str, target: str) -> None:
+    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest, target: str) -> None:
         self.c = MARSECMWFServiceExtended(
             "mars",
             key=self.config.kwargs.get('api_key', os.environ.get("MARSAPI_KEY")),
@@ -270,9 +269,9 @@ class MarsClient(Client):
         )
         selection_ = optimize_selection_partition(selection)
         with StdoutLogger(self.logger, level=logging.DEBUG):
-            with self.manifest_obj.transact(self.config.selection, target, self.config.user_id, 'FETCH'):
+            with manifest.transact(self.config.selection, target, self.config.user_id, 'FETCH'):
                 result = self.c.fetch(req=selection_)
-            with self.manifest_obj.transact(self.config.selection, target, self.config.user_id, 'DOWNLOAD'):
+            with manifest.transact(self.config.selection, target, self.config.user_id, 'DOWNLOAD'):
                 self.c.download(result, target=output)
 
     @property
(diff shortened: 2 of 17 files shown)


weather_dl/README.md
```diff
@@ -44,8 +44,11 @@ _Common options_:
   and `--direct_num_workers`
 * `-m, --manifest-location MANIFEST_LOCATION`: Location of the manifest. By default, it will use Cloud Logging
   (stdout for direct runner). You can set the name of the manifest as the hostname of a URL with the 'cli' protocol.
-  For example, 'cli://manifest' will prefix all the manifest logs as '[manifest]'. In addition, users can specify a GCS
-  bucket URI, or 'noop://<name>' for an in-memory location.
+  For example, `cli://manifest` will prefix all the manifest logs as '[manifest]'. In addition, users can specify 
+  either a BigQuery table (`bq://<project-id>.<dataset-name>.<table-name>`) or GCS bucket URI, or `noop://<name>`
```
alxmrs, line 48: +1 thanks, this is perfect.

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,22 +436,121 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+                        insert_dml.append(f"JSON'{status[col]}'")
+                    elif col == 'size' or col == 'error':
+                        update_dml.append(f"{col} = {status[col]}")
+                        insert_dml.append(status[col])
+                    else:
+                        update_dml.append(f"{col} = '{status[col]}'")
+                        insert_dml.append(f"'{status[col]}'")
+
+                # Build the merge statement as a string
+                merge_statement = f"""
```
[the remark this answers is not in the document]
alxmrs: CC: @blackvvine I bet you'd also have some opinions here :)

alxmrs (commented):
Made a first pass for a review. I recommend slightly less deviation from the existing class structure. 

So far, this looks great! I really like what I've seen in the demo DB :)

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,22 +436,121 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+            bigquery.SchemaField('size', 'FLOAT', mode='NULLABLE'),
+            bigquery.SchemaField('scheduled_time', 'TIMESTAMP', mode='NULLABLE'),
+            bigquery.SchemaField('retrieve_start_time', 'TIMESTAMP', mode='NULLABLE'),
+            bigquery.SchemaField('retrieve_end_time', 'TIMESTAMP', mode='NULLABLE'),
+            bigquery.SchemaField('fetch_start_time', 'TIMESTAMP', mode='NULLABLE'),
+            bigquery.SchemaField('fetch_end_time', 'TIMESTAMP', mode='NULLABLE'),
+            bigquery.SchemaField('download_start_time', 'TIMESTAMP', mode='NULLABLE'),
+            bigquery.SchemaField('download_end_time', 'TIMESTAMP', mode='NULLABLE'),
+            bigquery.SchemaField('upload_start_time', 'TIMESTAMP', mode='NULLABLE'),
+            bigquery.SchemaField('upload_end_time', 'TIMESTAMP', mode='NULLABLE'),
```
alxmrs, line 466: Let's add descriptions for each of these fields. e.g. users should have an answer to the question: "are these sizes in bytes or GB?" A description would make it easy to know.

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,22 +436,121 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+
+    This is an append-only implementation, the latest value in the manifest
+    represents the current state of a download.
+    """
+
+    _lock = threading.Lock()
+
+    def __init__(self, location: Location) -> None:
+        super().__init__(Location(location[5:]))
+        TABLE_SCHEMA = [
```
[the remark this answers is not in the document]
alxmrs: @TylerTCF may have some opinions, too.

weather_dl/download_pipeline/manifest.py
```diff
@@ -106,76 +202,161 @@ class Manifest(abc.ABC):
...
             error = None
         else:
-            status = 'failure'
+            status = 'FAILURE'
             # For explanation, see https://docs.python.org/3/library/traceback.html#traceback.format_exception
             error = '\n'.join(traceback.format_exception(exc_type, exc_inst, exc_tb))
 
+        current_utc_time = datetime.datetime.utcnow() \
+                                            .replace(tzinfo=datetime.timezone.utc) \
+                                            .isoformat(timespec='seconds')
```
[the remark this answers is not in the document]
mahrsee1997: Done. Thanks.

weather_dl/download_pipeline/manifest.py
```diff
@@ -106,76 +202,161 @@ class Manifest(abc.ABC):
...
+        retrieve_start_time = prev_download_status.retrieve_start_time
+        retrieve_end_time = prev_download_status.retrieve_end_time
+        fetch_start_time = prev_download_status.fetch_start_time
+        fetch_end_time = prev_download_status.fetch_end_time
+        download_start_time = prev_download_status.download_start_time
+        download_end_time = prev_download_status.download_end_time
+        upload_start_time = prev_download_status.upload_start_time
+        upload_end_time = prev_download_status.upload_end_time
+
+        current_utc_time = datetime.datetime.utcnow() \
```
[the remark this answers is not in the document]
mahrsee1997: Done.

weather_dl/download_pipeline/manifest.py
```diff
@@ -38,32 +42,124 @@ class ManifestException(Exception):
     pass
 
 
-class DownloadStatus(t.NamedTuple):
+class RequestStage(Enum):
```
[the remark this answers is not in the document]
mahrsee1997: Done.

weather_dl/download_pipeline/manifest.py
```diff
@@ -38,32 +42,124 @@ class ManifestException(Exception):
...
     """Location of the downloaded data."""
-    location: str
+    location: str = ""
+
+    """Current stage of request : 'fetch', 'download', 'retrieve', 'upload' or None."""
+    stage: t.Optional[RequestStage] = RequestStage.NONE
 
     """Download status: 'scheduled', 'in-progress', 'success', or 'failure'."""
-    status: str
+    status: RequestStatus = RequestStatus.SCHEDULED
```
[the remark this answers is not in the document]
mahrsee1997: Done.

weather_dl/download_pipeline/manifest.py
mahrsee1997: Done.

weather_dl/download_pipeline/manifest.py
```diff
@@ -106,76 +202,161 @@ class Manifest(abc.ABC):
 
     def __post_init__(self):
         """Initialize the manifest."""
-        self.start = 0
-        self.scheduled_times = {}
+        pass
```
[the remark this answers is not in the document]
mahrsee1997: umm this is required because we are using __post_init__() in ConsoleManifest (inherited from our base Manifest dataclass, defining __post_init__() in just the child class seems to be not working w.r.t to dataclasses). I do have a cleaner solution for the same.

**BUT** as per this approach of manifest, we can't really use ConsoleManifest because we need to read the state from the DB before updating the manifest.
Hence I was thinking of getting rid of ConsoleManifest all together and making LocalManifest as the default manifest system when running the tool in the local system & BQ manifest as the default manifest system when running the tool in Dataflow runner.
Furthermore we anyways log the manifest on the console after writing it so ConsoleManifest will be there always.

WDYT @alxmrs ?

weather_dl/download_pipeline/manifest.py
```diff
@@ -106,76 +202,161 @@ class Manifest(abc.ABC):
...
+                client = storage.Client()
+                bucket = client.bucket(bucket_name)
+                blob = bucket.get_blob(object_name)
+
+                size = blob.size / (10**9) if blob else 0
+            # In case of local system (--local-run).
+            except ValueError as e:
+                e_str = str(e)
+                if not (f"Invalid GCS location: {path!r}." in e_str):
+                    raise
```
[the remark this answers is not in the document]
mahrsee1997: This makes code look clean. Thanks.
Necessary code changes done.

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,22 +436,121 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+            with bigquery.Client() as client:
+                query = f"SELECT * FROM {self.location} WHERE location = {location!r}"
+                query_job = client.query(query)
+                result = query_job.result().to_dataframe().to_dict('records')
+                row = {n: to_json_serializable_type(v) for n, v in result[0].items()}
+                return DownloadStatus.from_dict(row)
+
+    def _update(self, download_status: DownloadStatus) -> None:
+        """Writes the JSON data to a manifest."""
+        with BQManifest._lock:
```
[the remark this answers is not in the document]
mahrsee1997: Yeah after thinking a bit, we actually don't need the lock in case of BQ Manifest.

Also BigQuery, does not provide traditional database transactions that follow ACID (Atomicity, Consistency, Isolation, Durability) properties.
**BUT** we are handling this by making the next transaction only when first one is complete.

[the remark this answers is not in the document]
mahrsee1997: Necessary code changes done.

weather_dl/download_pipeline/manifest.py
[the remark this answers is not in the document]
alxmrs: I think we should keep the console manifest. Even when we're running on Dataflow, we want the tool to not have a huge amount of infrastructural dependencies to get up and running. i.e. to get started, a new user would have to set up one additional GCP service (BigQuery) to use the tool. @blackvvine can explain this reasoning better than I can: 
https://github.com/google/weather-tools/issues/246#issue-1401678061

What you're saying makes sense wrt to keeping post_init. I'm happy to keep it -- can you just add a comment explaining that this is used in the subclasses?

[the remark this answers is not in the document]
mahrsee1997: > BUT as per this approach of manifest, we can't really use ConsoleManifest because **we need to read the state from the DB before updating the manifest.**

weather_dl/download_pipeline/fetcher.py
```diff
@@ -45,17 +45,17 @@ class Fetcher(beam.DoFn):
     """
 
     client_name: str
-    manifest: Manifest = NoOpManifest(Location('noop://in-memory'))
+    manifest: t.Dict[str, Location]
```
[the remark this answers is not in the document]
mahrsee1997: The state is indeed serializable. However, when passing the manifest instance from pipeline.py to other classes, a new instance is created internally every time. This is why I initially thought about making it more apparent to users. 
However, upon further consideration, I realized that creating the manifest instance separately may not be the most elegant solution.

Therefore, I have made the necessary code changes to revert back to our original approach.

weather_dl/download_pipeline/clients.py
```diff
@@ -47,14 +48,15 @@ class Client(abc.ABC):
         level: Default log level for the client.
     """
 
-    def __init__(self, config: Config, level: int = logging.INFO) -> None:
+    def __init__(self, config: Config, manifest_obj: Manifest, level: int = logging.INFO) -> None:
```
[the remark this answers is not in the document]
mahrsee1997: Resolved.

weather_dl/download_pipeline/manifest.py
```diff
@@ -38,32 +42,124 @@ class ManifestException(Exception):
...
     """Copy of selection section of the configuration."""
-    selection: t.Dict
+    selection: t.Dict = dataclasses.field(default_factory=dict)
 
     """Location of the downloaded data."""
-    location: str
+    location: str = ""
+
+    """Current stage of request : 'fetch', 'download', 'retrieve', 'upload' or None."""
+    stage: t.Optional[RequestStage] = RequestStage.NONE
```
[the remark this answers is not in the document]
mahrsee1997: Done. 

Removed `None` enum & set stage to None by default.

weather_dl/download_pipeline/manifest.py
```diff
@@ -106,76 +202,161 @@ class Manifest(abc.ABC):
...
+            size=None,
+            scheduled_time=scheduled_time,
+            retrieve_start_time=retrieve_start_time,
+            retrieve_end_time=retrieve_end_time,
+            fetch_start_time=fetch_start_time,
+            fetch_end_time=fetch_end_time,
+            download_start_time=download_start_time,
+            download_end_time=download_end_time,
+            upload_start_time=upload_start_time,
+            upload_end_time=upload_end_time,
```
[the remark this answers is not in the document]
mahrsee1997: Done. Thanks for suggesting.

weather_dl/download_pipeline/manifest.py
```diff
@@ -106,76 +202,161 @@ class Manifest(abc.ABC):
 
     def __post_init__(self):
         """Initialize the manifest."""
-        self.start = 0
-        self.scheduled_times = {}
+        pass
 
     def schedule(self, selection: t.Dict, location: str, user: str) -> None:
```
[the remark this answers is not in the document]
mahrsee1997: The necessary code modifications have been made.

weather_dl/download_pipeline/manifest.py
```diff
@@ -106,76 +202,161 @@ class Manifest(abc.ABC):
...
+
+        scheduled_time = prev_download_status.scheduled_time
+        retrieve_start_time = prev_download_status.retrieve_start_time
+        retrieve_end_time = prev_download_status.retrieve_end_time
+        fetch_start_time = prev_download_status.fetch_start_time
+        fetch_end_time = prev_download_status.fetch_end_time
+        download_start_time = prev_download_status.download_start_time
+        download_end_time = prev_download_status.download_end_time
+        upload_start_time = prev_download_status.upload_start_time
+        upload_end_time = prev_download_status.upload_end_time
```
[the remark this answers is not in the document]
mahrsee1997: Actually not if we are using dataclasses.replace(). 

Code modifications complete.

weather_dl/download_pipeline/manifest.py
```diff
@@ -106,76 +202,161 @@ class Manifest(abc.ABC):
...
+            fetch_end_time=fetch_end_time,
+            download_start_time=self.status.download_start_time,
+            download_end_time=download_end_time,
+            upload_start_time=self.status.upload_start_time,
+            upload_end_time=upload_end_time,
         )
         self._update(self.status)
 
-    def transact(self, selection: t.Dict, location: str, user: str) -> 'Manifest':
+    def transact(self, selection: t.Dict, location: str, user: str, stage: str) -> 'Manifest':
```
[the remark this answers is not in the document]
mahrsee1997: The necessary code modifications have been made.

Implemented (2) for now.

weather_dl/download_pipeline/clients.py
```diff
@@ -30,6 +30,7 @@
 from ecmwfapi import ECMWFService, api
 
 from .config import Config, optimize_selection_partition
+from .manifest import MANIFESTS, Manifest, Location, NoOpManifest
```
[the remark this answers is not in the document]
mahrsee1997: Yes, that's correct.

Our client's interface has a `retrieve()` method, which is called from `fetcher.py`. For the CDS client, this method performs both the fetch and download actions, while for the MARS client, it performs the fetch and download steps separately.

To capture these different stages, we need to have some coupling. While we may be able to separate the fetch and download for the CDS client by extending its API, as we did with the MARS client, there may be future clients for which this is not possible due to complex APIs.

weather_dl/download_pipeline/clients.py
```diff
@@ -267,8 +270,10 @@ def retrieve(self, dataset: str, selection: t.Dict, output: str) -> None:
         )
         selection_ = optimize_selection_partition(selection)
         with StdoutLogger(self.logger, level=logging.DEBUG):
-            result = self.c.fetch(req=selection_)
-            self.c.download(result, target=output)
+            with self.manifest_obj.transact(self.config.selection, target, self.config.user_id, 'FETCH'):
```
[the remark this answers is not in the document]
mahrsee1997: Check my other comment related to this.

weather_dl/download_pipeline/fetcher.py
```diff
@@ -68,11 +69,18 @@ def fetch_data(self, config: Config, *, worker_name: str = 'default') -> None:
...
+            with self.manifest.transact(config.kwargs.get('config_name'), config.selection, target, config.user_id):
                 self.retrieve(client, config.dataset, config.selection, temp.name)
 
+                self.manifest.set_stage(Stage.UPLOAD)
+                precise_upload_start_time = (
+                    datetime.datetime.utcnow()
+                    .replace(tzinfo=datetime.timezone.utc)
+                    .isoformat(timespec='seconds')
+                )
+                self.manifest.prev_stage_precise_start_time = precise_upload_start_time
```
alxmrs, line 83: It seems like this logic gets duplicated a bunch. Is there some way that we can include it in `set_stage`?

weather_dl/download_pipeline/fetcher_test.py
```diff
@@ -32,6 +33,7 @@ def setUp(self) -> None:
     @patch('cdsapi.Client.retrieve')
     def test_fetch_data(self, mock_retrieve):
         with tempfile.TemporaryDirectory() as tmpdir:
+            # self.dummy_manifest = LocalManifest(Location(tmpdir))
```
alxmrs, line 36: Remove?

weather_dl/download_pipeline/manifest.py
```diff
@@ -16,15 +16,18 @@
 import abc
 import collections
 import dataclasses
+import datetime
+from enum import Enum
```
alxmrs, line 20: Optional nit: let's use `import enum` and `class Foo(enum.Enum)`.

weather_dl/download_pipeline/manifest.py
```diff
@@ -102,80 +199,140 @@ class Manifest(abc.ABC):
...
+            new_status.fetch_end_time = current_utc_time
+        elif new_status.stage == Stage.RETRIEVE:
+            new_status.retrieve_start_time = self.prev_stage_precise_start_time
+            new_status.retrieve_end_time = current_utc_time
+        elif new_status.stage == Stage.DOWNLOAD:
+            new_status.download_start_time = self.prev_stage_precise_start_time
+            new_status.download_end_time = current_utc_time
+        else:
+            new_status.upload_start_time = self.prev_stage_precise_start_time
+            new_status.upload_end_time = current_utc_time
```
alxmrs, line 282: Can we make this more succinct with a `setattr` call? This is a bit reflective, but I bet we can make the update to new status something like: 

```
setattr(new_status, f'{stage}_start_time', self.prev_stage_precise_start_time)
setattr(new_status, f'{stage}_end_time', current_utc_time)
```

weather_dl/download_pipeline/manifest.py
```diff
@@ -102,80 +199,140 @@ class Manifest(abc.ABC):
...
+
+        path = new_status.location
+        parsed_gcs_path = urlparse(path)
+        if parsed_gcs_path.scheme != 'gs' or parsed_gcs_path.netloc == '':
+            new_status.size = LocalSystemFileSizeStrategy().get_file_size(path)
+        else:
+            new_status.size = GCSBlobSizeStrategy().get_file_size(parsed_gcs_path)
```
alxmrs, line 289: This is much cleaner. However, could we apply something like the strategy pattern on the manifest subclass itself? e.g. the GCS Manifest has a `clac_file_size()` method that gets overridden? I see no need to create a separate class hierarchy, this one should do fine.

alxmrs: For example, our `_read` and `_update` calls use the strategy pattern in my book.

weather_dl/download_pipeline/manifest.py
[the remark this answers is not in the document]
alxmrs: Before we land this change, I was thinking Tyler could build a prototype dashboard out of sample data you produce. That should be a good test of the schema.

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,22 +444,138 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+                ON T.location = S.location
+                WHEN MATCHED THEN
+                UPDATE SET
+                    {', '.join(str(v) for v in update_dml)}
+                WHEN NOT MATCHED THEN
+                INSERT
+                    ({", ".join(columns)})
+                VALUES
+                    ({', '.join(str(v) for v in insert_dml)})
+            """
```
alxmrs, line 543: Does BQ include some sort of protection against SQL injection? I think this system of formatting strings is unsafe.

alxmrs: It is necessary to use a library to sanitize user inputs.

weather_dl/download_pipeline/parsers.py
```diff
@@ -492,4 +494,7 @@ def validate_all_configs(configs: t.List[Config]) -> None:
     require(all_equal(clients), f'All configs must request data from the same client, {clients[0]!r}.')
 
     kwargs = [conf.kwargs for conf in configs]
+    # Since the config file names will vary between configs,
+    # there's no need to perform an equality check for this.
```
alxmrs, line 498: Thanks for the comment!

alxmrs (commented):
More feedback. Overall it's looking good!

weather_dl/download_pipeline/parsers.py
mahrsee1997: Thanks. Though this code has been removed since we moved the `config_name` property to the top level of the config dataclass.

weather_dl/download_pipeline/manifest.py
mahrsee1997: Done.

weather_dl/download_pipeline/fetcher.py
```diff
@@ -68,11 +69,18 @@ def fetch_data(self, config: Config, *, worker_name: str = 'default') -> None:
         client = CLIENTS[self.client_name](config)
         target = prepare_target_name(config)
 
-        with self.manifest.transact(config.selection, target, config.user_id):
-            with tempfile.NamedTemporaryFile() as temp:
-                logger.info(f'[{worker_name}] Fetching data for {target!r}.')
+        with tempfile.NamedTemporaryFile() as temp:
+            logger.info(f'[{worker_name}] Fetching data for {target!r}.')
+            with self.manifest.transact(config.kwargs.get('config_name'), config.selection, target, config.user_id):
```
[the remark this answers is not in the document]
mahrsee1997: Done.

weather_dl/download_pipeline/manifest.py
```diff
@@ -352,24 +351,56 @@ def _update(self, download_status: DownloadStatus) -> None:
...
 
 class GCSManifest(Manifest):
-    """Writes a JSON representation of the manifest to GCS.
+    """Writes a JSON representation of the manifest to GCS."""
 
-    This is an append-only implementation, the latest value in the manifest
-    represents the current state of a download.
-    """
+    def __init__(self, location: Location) -> None:
+        super().__init__(Location('{}{}manifest.json'.format(location, os.sep)))
```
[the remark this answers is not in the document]
mahrsee1997: Done.

weather_dl/download_pipeline/clients.py
```diff
@@ -54,7 +56,7 @@ def __init__(self, config: Config, level: int = logging.INFO) -> None:
         self.logger.setLevel(level)
 
     @abc.abstractmethod
-    def retrieve(self, dataset: str, selection: t.Dict, output: str) -> None:
+    def retrieve(self, dataset: str, selection: t.Dict, output: str, manifest: Manifest) -> None:
```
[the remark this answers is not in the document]
mahrsee1997: By default, it will use NoOpManifest since this argument is passed from fetcher.py, and if we check there, we can see that the manifest is initialized to NoOpManifest by default."

weather_dl/download_pipeline/manifest.py
```diff
@@ -352,24 +351,56 @@ def _update(self, download_status: DownloadStatus) -> None:
...
-    # Ensure no race conditions occurs on appends to objects in GCS
-    # (i.e. JSON manifests).
-    _lock = threading.Lock()
+    def get_json(self):
+        """This function will get the json object from GCS bucket."""
+        storage_client = storage.Client()
+
+        parsed_file_name = urlparse(self.location)
+        if parsed_file_name.scheme != 'gs' or parsed_file_name.netloc == '':
+            raise ValueError(f'Invalid GCS location: {self.location!r}.')
```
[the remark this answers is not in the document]
mahrsee1997: Done. Thanks for pointing out.

weather_dl/download_pipeline/manifest.py
```diff
@@ -352,24 +351,56 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+        bucket_name = parsed_file_name.netloc
+        BUCKET = storage_client.get_bucket(bucket_name)
+
+        # Get the blob
+        blob = BUCKET.get_blob(parsed_file_name.path[1:])
+        return json.loads(blob.download_as_string()) if blob else {}
```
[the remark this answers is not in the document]
mahrsee1997: No such reason.
Modified the code to use `from apache_beam.io.gcp import gcsio` -- code looks much cleaner using this.

weather_dl/download_pipeline/manifest.py
```diff
@@ -352,24 +351,56 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+
+        BUCKET = storage_client.get_bucket(bucket_name)
+
+        # Create a blob
+        blob = BUCKET.blob(parsed_file_name.path[1:])
+        # Upload the blob
+        blob.upload_from_string(
+            data=json.dumps(json_object),
+            content_type='application/json'
+        )
```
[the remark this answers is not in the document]
mahrsee1997: Done. Check my above comment.

weather_dl/download_pipeline/manifest.py
mahrsee1997: I changed the implementation a bit w.r.t to getting file size. WRYT ?

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,22 +447,138 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+
+    This is an append-only implementation, the latest value in the manifest
+    represents the current state of a download.
+    """
+    def __init__(self, location: Location) -> None:
+        super().__init__(Location(location[5:]))
+        TABLE_SCHEMA = [
+            bigquery.SchemaField('config_name', 'STRING', mode='REQUIRED',
+                                 description="The name of the config file associated with the request."),
+            bigquery.SchemaField('selection', 'JSON', mode='REQUIRED',
```
alxmrs, line 461: Hey Rahul. I'm talking to Tyler about his dashboard (that doesn't use this DB... yet). He needs some way to calculate the total number of shards in order to display the percentage complete. Any ideas to accomplish this?

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,22 +447,138 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+            bigquery.SchemaField('status', 'STRING', mode='REQUIRED',
+                                 description="Download status: 'scheduled', 'in-progress', 'success', or 'failure'."),
+            bigquery.SchemaField('error', 'STRING', mode='NULLABLE',
+                                 description="Cause of error, if any."),
+            bigquery.SchemaField('user', 'STRING', mode='REQUIRED',
+                                 description="Identifier for the user running the download."),
+            bigquery.SchemaField('size', 'FLOAT', mode='NULLABLE',
+                                 description="File size in GB."),
+            bigquery.SchemaField('scheduled_time', 'TIMESTAMP', mode='NULLABLE',
+                                 description="A UTC datetime when download was scheduled."),
```
alxmrs, line 477: Another idea from Tyler + I: Let's include a nullable Geospatial Polygon. This column would represent the `area` lat/lngs from the selection. If we record the area covered by a shard, we can use BQ to quickly check if we have data coverage. 

If the `area` is omitted, let's assume it has global coverage (put in a default global polygon).

weather_dl/download_pipeline/fetcher.py
mahrsee1997: This was done to reduce the impact of _read() and _update() calls on the stage start time, effectively eliminating the time taken by insert and update query operations.

weather_dl/download_pipeline/manifest.py
mahrsee1997: Yeah we can do that. Basically in partition step as soon as the shards are created it will update the manifest stating it is `scheduled` (i.e. status: scheduled). This allows us to obtain the total number of shards for a particular config.
To determine the number of shards that have been successfully downloaded, we can check for shards with a stage of 'upload' and a status of 'success'.

weather_dl/download_pipeline/manifest.py
alxmrs: Thanks Rahul :)

weather_dl/download_pipeline/manifest.py
mahrsee1997: Thanks Alex for pointing this.

I have made the necessary changes for this -- by using parameterized queries, we ensure that the user input is properly sanitized and prevent SQL injection attacks, because the user input is never directly embedded in the SQL query string.

weather_dl/download_pipeline/manifest.py
alxmrs: Brilliant. I like this clean, simple solution.

weather_dl/download_pipeline/clients.py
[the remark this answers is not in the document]
alxmrs: SGTM, thanks!

weather_dl/download_pipeline/fetcher.py
alxmrs: If there's a good technical reason to include this, I'm OK with it.

weather_dl/README.md
```diff
@@ -44,8 +44,11 @@ _Common options_:
   and `--direct_num_workers`
 * `-m, --manifest-location MANIFEST_LOCATION`: Location of the manifest. By default, it will use Cloud Logging
   (stdout for direct runner). You can set the name of the manifest as the hostname of a URL with the 'cli' protocol.
-  For example, 'cli://manifest' will prefix all the manifest logs as '[manifest]'. In addition, users can specify a GCS
-  bucket URI, or 'noop://<name>' for an in-memory location.
+  For example, `cli://manifest` will prefix all the manifest logs as '[manifest]'. In addition, users can specify 
+  either a BigQuery table (`bq://<project-id>.<dataset-name>.<table-name>`) or GCS bucket URI, or `noop://<name>` 
+  for an in-memory location.
+> Note: Tool will create the BQ table itself, if not already present.
```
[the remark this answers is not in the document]
mahrsee1997: `arg docs` ?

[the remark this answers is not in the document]
alxmrs: Sorry, I mean the help message of the CLI associated with this argument.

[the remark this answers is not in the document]
mahrsee1997: Yeah, we have already mentioned the same in help message of CLI.

Also moved it below the `_Common options_`.

weather_dl/download_pipeline/manifest.py
mahrsee1997: Done. Included the `area` (GeoJSON geometry object) as STRING in BQ schema.

Some useful functions for analysis purpose:
* [ST_GEOGFROMGEOJSON](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_geogfromgeojson) : To convert a GeoJSON geometry object into a GEOGRAPHY value. 
* ST_COVERS(geography_expression, ST_GEOGPOINT(longitude, latitude)): To check if a point lies in the given area or not.

weather_dl/download_pipeline/manifest.py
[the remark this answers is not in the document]
alxmrs: As discussed in a 1:1 today, the console manifest doesn't have to be stateful (or perfect). We can merely update the user with the information we have available.

weather_dl/download_pipeline/manifest.py
alxmrs: The above is optional. What Rahul has is perfectly fine.

weather_dl/download_pipeline/manifest.py
```diff
@@ -184,31 +347,13 @@ def _update(self, download_status: DownloadStatus) -> None:
...
-    """Writes a JSON representation of the manifest to GCS.
-
-    This is an append-only implementation, the latest value in the manifest
-    represents the current state of a download.
-    """
-
-    # Ensure no race conditions occurs on appends to objects in GCS
-    # (i.e. JSON manifests).
-    _lock = threading.Lock()
+    def _read(self, location: str) -> DownloadStatus:
```
alxmrs, line 352: LGTM

weather_dl/download_pipeline/manifest.py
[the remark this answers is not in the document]
alxmrs: I'll check with Tyler one last time. Otherwise, I think we're good to go.

weather_dl/download_pipeline/manifest.py
alxmrs: Looks good to me!

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,22 +395,141 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+                                 description="A UTC datetime when the upload state ends."),
+        ]
+        table = bigquery.Table(self.location, schema=TABLE_SCHEMA)
+        with bigquery.Client() as client:
+            client.create_table(table, exists_ok=True)
+
+    def _read(self, location: str) -> DownloadStatus:
+        """Reads the JSON data from a manifest."""
+        with bigquery.Client() as client:
+            query = f"SELECT * FROM {self.location} WHERE location = {location!r}"
```
alxmrs, line 456: Can you sanitize these inputs too?

alxmrs: I think this will only be necessary if the `query()` command can write or delete data in the DB. 

Reading the docs, it looks like this is the case.

https://cloud.google.com/python/docs/reference/bigquery/latest/google.cloud.bigquery.client.Client#google_cloud_bigquery_client_Client_query

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,22 +395,141 @@ def _update(self, download_status: DownloadStatus) -> None:
...
+            query_job = client.query(query)
+            result = query_job.result().to_dataframe().to_dict('records')
+            row = {n: to_json_serializable_type(v) for n, v in result[0].items()}
+            return DownloadStatus.from_dict(row)
+
+    def _update(self, download_status: DownloadStatus) -> None:
```
alxmrs, line 462: I like this method -- very well written.

weather_dl/download_pipeline/pipeline.py
```diff
@@ -168,7 +170,9 @@ def run(argv: t.List[str], save_main_session: bool = True) -> PipelineArgs:
     configs = []
     for cfg in known_args.config:
         with open(cfg, 'r', encoding='utf-8') as f:
-            config = process_config(f)
+            # configs/example.cfg -> example.cfg
```
alxmrs, line 173: Thanks, I like this comment.

weather_dl/download_pipeline/util.py
```diff
@@ -72,3 +80,67 @@ def copy(src: str, dst: str) -> None:
     except subprocess.CalledProcessError as e:
         logger.error(f'Failed to copy file {src!r} to {dst!r} due to {e.stderr.decode("utf-8")}')
         raise
+
+
+# TODO(#245): Group with common utilities (duplicated)
```
alxmrs, line 85: So sorry to ask you to do this, but: Can you duplicate this method in the `loader_pipeline` util.py file until we fix #245?

alxmrs: Oh, I see -- this is the copy! 

In that case, can you add this comment to the loader pipeline duplicate :)

alxmrs (approved):
I have a few critical points of feedback. Once these are fixed, I'm happy to land these changes. Great work, Rahul!

weather_dl/download_pipeline/manifest.py
mahrsee1997: Thanks Alex.

weather_dl/download_pipeline/manifest.py
mahrsee1997: Done. Thanks for pointing this out.

weather_dl/download_pipeline/util.py
mahrsee1997: Done.

mahrsee1997:
Thanks Alex for reviewing.

Issue #159, opened 2022-05-30
Explore ways to capture required metadata from TIF file using xarray


mahrsee1997:
Currently `rasterio` has been used to fetch metadata from TIF file because `xarray's` open_dataset method is not providing the same.

mahrsee1997:
Can make use of rioxarray here. This will load bands in a raster to separate variables & preserves the attrs.
```
import rioxarray
ds = rioxarray.open_rasterio('./S10090600100918001.tiff', band_as_variable=True) # Magical call
```
Ref: https://github.com/corteva/rioxarray/pull/600

PR #296, opened 2023-02-16
Improvements on parsing time from tiffs.

alxmrs:
Fixes and improved codepath for parsing time from Tiffs. Useful for performing round-trip tests.

### DIFF
--- a/weather_mv/loader_pipeline/sinks.py
+++ b/weather_mv/loader_pipeline/sinks.py
@@ -163,8 +163,11 @@ def _preprocess_tif(ds: xr.Dataset, filename: str, tif_metadata_for_datetime: st
     ds = xr.merge(band_data_list)
     ds.attrs['is_normalized'] = ds_is_normalized_attr
 
-    end_time = None
-    if initialization_time_regex and forecast_time_regex:
+    start_time, end_time = None, None
+    time_attrs_present = 'start_time' in ds.attrs and 'end_time' in ds.attrs
+    if time_attrs_present:
+        start_time, end_time = ds.attrs['start_time'], ds.attrs['end_time']
+    elif initialization_time_regex and forecast_time_regex:
         try:
             start_time = match_datetime(uri, initialization_time_regex)
         except Exception:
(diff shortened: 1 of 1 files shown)


weather_mv/loader_pipeline/sinks.py
```diff
@@ -178,15 +181,17 @@ def _get_band_data(i):
 
     # TODO(#159): Explore ways to capture required metadata using xarray.
     with rasterio.open(filename) as f:
-        datetime_value_ms = None
+        datetime_value_s = None
```
mahrsee1997, line 184: +1 nice catch.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -163,8 +163,11 @@ def _get_band_data(i):
     ds = xr.merge(band_data_list)
     ds.attrs['is_normalized'] = ds_is_normalized_attr
 
-    end_time = None
-    if initialization_time_regex and forecast_time_regex:
+    start_time, end_time = None, None
+    time_attrs_present = 'start_time' in ds.attrs and 'end_time' in ds.attrs
+    if time_attrs_present:
+        start_time, end_time = ds.attrs['start_time'], ds.attrs['end_time']
+    elif initialization_time_regex and forecast_time_regex:
```
mahrsee1997, line 170: Considering the user-specified flags `initialization_time_regex` and `forecast_time_regex`, we should prioritize the case where these flags are used over the case where `time_attrs_present` is used, and therefore move this case above the `time_attrs_present` case.

WDYT ?

mahrsee1997 (commented):
Overall looks good ... just a minor feedback point.

weather_mv/loader_pipeline/sinks.py
alxmrs: Can you explain by what you mean by prioritize? Shall I switch the order of the conditional blocks (user input gets checked first)?

mahrsee1997: > Shall I switch the order of the conditional blocks (user input gets checked first)?

Yes, exactly. I think user inputs should be given priority over the default.

alxmrs:
Thanks for the feedback. The changes are now updated.

mahrsee1997 (approved):
LGTM !

weather_mv/loader_pipeline/sinks.py
```diff
@@ -175,18 +175,22 @@ def _get_band_data(i):
             raise RuntimeError("Wrong regex passed in --forecast_time_regex.")
         ds.attrs['start_time'] = start_time
         ds.attrs['end_time'] = end_time
+    elif 'start_time' in ds.attrs and 'end_time' in ds.attrs:
```
deepgabani8, line 178: 3 points
- Our tiff files do not preserve attrs so reading them with `xr.open_dataset(file.tiff, engine='rasterio')` will have empty `ds.attrs`.
- 'start_time' and 'end_time' in dataset attrs is in `2021-01-01T00:00:00Z` format so we need to parse it into datetime.datetime() for below datetime_value_s variable to initialize.
- `start_time` is not being used. Can be skipped.

mahrsee1997: Good points Deep.
May be we can pick this https://github.com/google/weather-tools/issues/159#issuecomment-1342490273 next & restructure `_preprocess_tif` altogether.

PR #297, opened 2023-02-22
`--update-manifest` : to update the manifest for the already downloaded shards and exit.

mahrsee1997:
The `--update-manifest` flag allows users to update the manifest for only the already downloaded shards by providing a config, and then exit the tool. If this flag is not provided, the tool will proceed with its usual functionality and also update the manifest for the already downloaded shards.


weather_dl/download_pipeline/partition.py
```diff
@@ -193,13 +198,19 @@ def prepare_partitions_from_index(config: Config, indexes: t.List[Index]) -> t.I
...
 
-def new_downloads_only(candidate: Config, store: t.Optional[Store] = None) -> bool:
+def new_downloads_only(candidate: Config, store: t.Optional[Store] = None,
+                       manifest: Manifest = NoOpManifest(Location('noop://in-memory')),
+                       update_manifest: bool = False) -> bool:
     """Predicate function to skip already downloaded partitions."""
     if store is None:
         store = FSStore()
-    should_skip = skip_partition(candidate, store)
+    should_skip = skip_partition(candidate, store, manifest)
```
alxmrs, line 207: I'm concerned about having a side effect in a `Filter` operation. I'd rather let this be a predicate, and include the update logic in a separate beam step later on.

weather_dl/download_pipeline/manifest.py
```diff
@@ -242,6 +241,45 @@ def schedule(self, config_name: str, selection: t.Dict, location: str, user: str
             )
         self._update(self.status)
 
+    def skip(self, config_name: str, selection: t.Dict, location: str, user: str) -> None:
+        """Updates the manifest to mark the shards that were skipped in the current job
```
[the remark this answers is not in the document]
alxmrs: I don't think we need a new API on the manifest. Instead, let's use the same transaction logic in a new Beam step in a different part of the pipeline (more comments on this below).

alxmrs (commented):
Overall feedback: I'd like a different pipeline pathway for this feature. I'm concerned about coupling side effects with a filter operation, and in general, for this subtle approach to cause problems later on.

weather_dl/download_pipeline/partition.py
mahrsee1997: Updated the code.

weather_dl/download_pipeline/pipeline.py
```diff
@@ -110,6 +110,7 @@ def subsection_and_request(it: Config) -> t.Tuple[str, int]:
                                 args.manifest,
                                 args.known_args.schedule,
                                 args.known_args.partition_chunks,
+                                args.known_args.update_manifest,
```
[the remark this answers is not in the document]
mahrsee1997: Done.

weather_dl/download_pipeline/partition.py
```diff
@@ -148,6 +152,7 @@ def skip_partition(config: Config, store: Store) -> bool:
     target = prepare_target_name(config)
     if store.exists(target):
         logger.info(f'file {target} found, skipping.')
+        manifest.skip(config.config_name, config.selection, target, config.user_id)
```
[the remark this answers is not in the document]
mahrsee1997: A read from DB is required to make sure we are not overwriting the unintentional row.

The manifest needs to be updated for a skipped shard if its entry is not present, or  if the stage is not 'upload', or if the stage is 'upload' but the status is not 'success'.

alxmrs (approved):
LGTM, thanks Rahul.

weather_dl/download_pipeline/pipeline_test.py
```diff
@@ -56,7 +56,8 @@
                                   num_requests_per_key=-1,
                                   partition_chunks=None,
                                   schedule='in-order',
-                                  check_skip_in_dry_run=False),
+                                  check_skip_in_dry_run=False,
+                                  update_manifest=False),
```
alxmrs, line 60: Let's add a(n integration) test either here or in the partitions_tests file that exercises the update manifest feature.

alxmrs:
These changes LGTM, Rahul. In light of our discussion in chat, I'll be satisfied landing this once we have an addition test for the feature.

weather_dl/download_pipeline/pipeline_test.py
```diff
@@ -135,6 +135,12 @@ def test_local_run(self):
             )
         )
 
+    def test_update_manifest(self):
```
alxmrs, line 138: This is good. I'd also like to see coverage that `--update-manifest` correctly updates the manifest as we would expect. A couple ways we could do this: 
- here: let's write to a local manifest and assert that the output is correct. We also can also make multiple runs to show that the manifest handles state correctly (e.g. skips already downloaded data). 
- in a unit tests: abstract out the manifest layer and ensure that independent aspects of the feature are implemented correctly.

mahrsee1997: Yes, definitely thats the plan -- will update the unit test.

weather_dl/download_pipeline/pipeline_test.py
mahrsee1997: Added unit tests in partitions_tests.

alxmrs:
Thanks Rahul! I agree that the CL is ready to be merged.

mahrsee1997:
Thanks for the review Alex.

PR #300, opened 2023-03-07
Added retry logic to handle concurrency issue in BQ.

mahrsee1997:
- [x] Do e2e testing for the same.

### DIFF
--- a/weather_dl/download_pipeline/manifest.py
+++ b/weather_dl/download_pipeline/manifest.py
@@ -27,9 +27,9 @@ import traceback
 import typing as t
 
 from .util import (
-    to_json_serializable_type, 
-    fetch_geo_polygon, 
-    get_file_size, 
+    to_json_serializable_type,
+    fetch_geo_polygon,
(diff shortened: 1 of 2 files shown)


weather_dl/download_pipeline/manifest.py
```diff
@@ -26,7 +26,7 @@
 import traceback
 import typing as t
 
-from .util import to_json_serializable_type, fetch_geo_polygon, get_file_size
+from .util import to_json_serializable_type, fetch_geo_polygon, get_file_size, retry_with_exponential_backoff
```
alxmrs, line 29: Since this line is getting a bit long, let's refactor it to use a multi-line import with parens: 
```suggestion
from .util import (
    to_json_serializable_type, 
    fetch_geo_polygon, 
    get_file_size, 
    retry_with_exponential_backoff
)
```

alxmrs (approved):
One nit; otherwise this LGTM. 

Please merge after e2e testing.

Issue #264, opened 2022-12-02
Add optional feature to group common hypercubes into image collections when loading grib2 data into Earth Engine.


alxmrs:
There is a significant performance problem that occurs when trying to convert a large grib file into a single COG. It seems like pyramiding data in GDAL takes exponential time (probably a bug). 

Thus, there's an advantage to splitting up large grib files into multiple ImageCollections / COGS. At the same time, we'd like to group COGs that come from the same source together, so it's easy to access groups of variables as bands. For example, we don't want to write one COG per band, as that would make the access pattern more difficult. 

The solution that @fredzyda, @j9sh264 and I have discussed for this is to make one COG / ImageCollection per layer of the grib file. As described [here](https://github.com/ecmwf/cfgrib#automatic-filtering), `cfgrib` will naturally return multiple datasets when opening a mixed hypercubes of data in a single grib. These offer natural groupings to break up the expensive COG operation. 

Let's implement this capability generically, where we default to normalized grib schemas (#184), and let users opt-in to creating multiple image collections in Earth Engine.

PR #304, opened 2023-03-17
Added a flag to allow for level-wise ingestion of data.

j9sh264:
In this PR I have introduced a new flag named `--group_common_hypercubes` which allows to ingest data in a level-wise format into the Earth engine.

For more context refer to this [issue](https://github.com/google/weather-tools/issues/264).


weather_mv/loader_pipeline/ee.py
```diff
@@ -388,77 +392,85 @@ class ConvertToAsset(beam.DoFn):
...
-                            # Making the channel name EE-safe before adding it as a band name.
-                            f.set_band_description(i+1, get_ee_safe_name(channel_names[i]))
-                            f.update_tags(i+1, band_name=channel_names[i])
-
-                        # Write attributes as tags in tiff.
-                        f.update_tags(**attrs)
-
-                    # Copy in-memory tiff to gcs.
+                          self.disable_grib_schema_normalization,
+                          group_common_hypercubes=self.group_common_hypercubes) as ds_list:
```
j9sh264, line 404: The new `open_dataset` function returns a list of Xarray datasets (instead of just a single dataset before). This will affect the real-time data pipelines.
cc @alxmrs

weather_mv/loader_pipeline/sinks.py
```diff
@@ -354,35 +383,52 @@ def open_dataset(uri: str,
                  open_dataset_kwargs: t.Optional[t.Dict] = None,
                  disable_grib_schema_normalization: bool = False,
                  tif_metadata_for_datetime: t.Optional[str] = None,
+                 group_common_hypercubes: t.Optional[bool] = False,
                  band_names_dict: t.Optional[t.Dict] = None,
                  initialization_time_regex: t.Optional[str] = None,
-                 forecast_time_regex: t.Optional[str] = None) -> t.Iterator[xr.Dataset]:
+                 forecast_time_regex: t.Optional[str] = None) -> t.Iterator[t.List[xr.Dataset]]:
```
alxmrs, line 389: Just thinking aloud: what are ways that we could implement this to get backwards compatibility? Is there a way we could introduce this change that would insulate us from breaking internal pipelines? 

One idea: What if we return a union type here (`t.Union[xr.Dataset, t.List[xr.Dataset]]`, just like the above)? For most dataset (e.g. at least all non-grib pipelines, and most grib pipelines) we'll use the first output of the union. This may save us from making code changes to a lot of areas of the codebase, for example, we won't have to implement this in our BQ code for now. 

It's not clear to me, for example, that the BQ pipeline here is correct. For this change, we'd need to test that this support creates the correct BQ schema as well as writes rows accurately. I'd say it's safer to assume all BQ paths that try to open lists of Datasets throw an error for now, and have this only apply to the EE pipeline (I'd probably be more comfortable with the EE pipeline AND when the `group_common_hypercubes` flag is set, which is more predictable). 

CC: @fredzyda -- LMK if this makes sense to you or if you have any thoughts here.

alxmrs (commented):
A thought on the design of this change.

weather_mv/loader_pipeline/sinks.py
fredzyda: I think you're right @alxmrs. I think maintaining backwards compatibility would be a good thing here.

fredzyda (commented):
I agree with Alex.

weather_mv/loader_pipeline/bq.py
```diff
@@ -52,17 +52,14 @@
 @dataclasses.dataclass
 class ToBigQuery(ToDataSink):
     """Load weather data into Google BigQuery.
-
```
alxmrs, line 55: Nit: I think we should keep this whitespace.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -281,26 +292,44 @@ def __normalize_grib_dataset(filename: str) -> xr.Dataset:
...
-    ds_attrs.update(**dv_units_dict)
+                _level_data_dict[level].append(copied_da)
+
+    _data_array_list = []
+    for level, ds in _level_data_dict.items():
+        if len(ds) == 1:
+            dataset = ds[0].to_dataset(promote_attrs=True)
+        else:
+            dataset = xr.merge(ds)
+        _data_array_list.append(dataset)
```
alxmrs, line 304: Nice

alxmrs (approved):
Let's fix a few small issues first; otherwise I think this is good to go :)

weather_mv/loader_pipeline/ee.py
```diff
@@ -436,75 +440,85 @@ def convert_to_asset(self, queue: Queue, uri: str):
...
+                channel_names = [da.name for da in data]
+                start_time, end_time, is_normalized = (attrs.get(key) for key in
+                                                       ('start_time', 'end_time', 'is_normalized'))
+                dtype, crs, transform = (attrs.pop(key) for key in ['dtype', 'crs', 'transform'])
+                attrs.update({'is_normalized': str(is_normalized)})  # EE properties does not support bool.
+                # Make attrs EE ingestable.
+                attrs = make_attrs_ee_compatible(attrs)
+
+                if self.group_common_hypercubes:
+                    level, height = (attrs.pop(key) for key in ['level', 'height'])
```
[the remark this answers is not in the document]
j9sh264: These attributes were not present in the final asset to be uploaded before. I just added them in the case we have `group_common_hypercubes` enabled. We are using them internally in code. Hence I used `pop` instead of `get`.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -380,35 +409,50 @@ def open_dataset(uri: str,
...
+            if group_common_hypercubes:
+                total_size_in_bytes = 0
+
+                for xr_dataset in xr_datasets:
+                    xr_dataset.attrs.update({'dtype': dtype, 'crs': crs, 'transform': transform})
+                    total_size_in_bytes += xr_dataset.nbytes
+
+                logger.info(f'opened dataset size: {total_size_in_bytes}')
+            else:
+                if uri_extension in ['.tif', '.tiff']:
```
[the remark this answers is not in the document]
j9sh264: Resolved

weather_mv/loader_pipeline/sinks.py
j9sh264: Thank you.

weather_mv/loader_pipeline/ee.py
```diff
@@ -436,75 +440,85 @@ def convert_to_asset(self, queue: Queue, uri: str):
...
-                            # Making the channel name EE-safe before adding it as a band name.
-                            f.set_band_description(i+1, get_ee_safe_name(channel_names[i]))
-                            f.update_tags(i+1, band_name=channel_names[i])
-                            f.update_tags(i+1, **da.attrs)
-                        # Write attributes as tags in tiff.
-                        f.update_tags(**attrs)
-
-                    # Copy in-memory tiff to gcs.
+                          forecast_time_regex=self.forecast_time_regex) as ds_list:
+            if type(ds_list) is not list:
```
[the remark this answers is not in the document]
j9sh264: Resolved

weather_mv/loader_pipeline/bq.py
j9sh264: Resolved

weather_mv/test_data/test_data_grib_multi_levels.grib2
mahrsee1997, line 1: Is this necessary ? Can we make use of existing test data file ?

weather_mv/loader_pipeline/sinks.py
```diff
@@ -394,29 +424,42 @@ def open_dataset(uri: str,
...
-                dtype, crs, transform = (f.profile.get(key) for key in ['dtype', 'crs', 'transform'])
-                xr_dataset.attrs.update({'dtype': dtype, 'crs': crs, 'transform': transform})
+            xr_dataset.attrs.update({'dtype': dtype, 'crs': crs, 'transform': transform})
 
             logger.info(f'opened dataset size: {xr_dataset.nbytes}')
 
             beam.metrics.Metrics.counter('Success', 'ReadNetcdfData').inc()
-            yield xr_dataset
+            yield xr_datasets if group_common_hypercubes else xr_dataset
             xr_dataset.close()
```
mahrsee1997, line 461: Please do necessary changes here in case of `group_common_hypercubes`.

weather_mv/loader_pipeline/ee.py
```diff
@@ -268,6 +269,8 @@ def add_parser_arguments(cls, subparser: argparse.ArgumentParser):
                                help='The expected latency per requests, in seconds. Default: 0.5')
         subparser.add_argument('--ee_max_concurrent', type=int, default=10,
                                help='Maximum concurrent api requests to EE allowed for your project. Default: 10')
+        subparser.add_argument('--group_common_hypercubes', action='store_true', default=False,
```
mahrsee1997, line 272: Please update README.md.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -394,29 +424,42 @@ def open_dataset(uri: str,
...
                                              initialization_time_regex,
                                              forecast_time_regex)
+            else:
+                xr_dataset = xr_datasets
 
-            # Extracting dtype, crs and transform from the dataset & storing them as attributes.
-            with rasterio.open(local_path, 'r') as f:
-                dtype, crs, transform = (f.profile.get(key) for key in ['dtype', 'crs', 'transform'])
-                xr_dataset.attrs.update({'dtype': dtype, 'crs': crs, 'transform': transform})
+            xr_dataset.attrs.update({'dtype': dtype, 'crs': crs, 'transform': transform})
```
mahrsee1997, line 455: Looks off to me. Let's discuss this over call.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -281,26 +292,44 @@ def __normalize_grib_dataset(filename: str) -> xr.Dataset:
...
                         disable_grib_schema_normalization: bool,
-                        open_dataset_kwargs: t.Optional[t.Dict] = None) -> xr.Dataset:
+                        open_dataset_kwargs: t.Optional[t.Dict] = None,
+                        group_common_hypercubes: t.Optional[bool] = False) -> t.Union[xr.Dataset, t.List[xr.Dataset]]:
     """Opens the dataset at 'uri' and returns a xarray.Dataset."""
+    # add a flag to group common hypercubes
+    if group_common_hypercubes:
+        return __normalize_grib_dataset(filename, group_common_hypercubes)
+
+    # add a flag to use cfgrib
```
mahrsee1997, line 332: Is this required `# add a flag to use cfgrib`?

mahrsee1997 (changes requested):

weather_mv/loader_pipeline/sinks.py
j9sh264: Removed

weather_mv/loader_pipeline/sinks.py
j9sh264: Resolved

weather_mv/loader_pipeline/ee.py
j9sh264: Updated

weather_mv/loader_pipeline/sinks.py
j9sh264: Resolved

weather_mv/loader_pipeline/sinks.py
```diff
@@ -281,26 +292,43 @@ def __normalize_grib_dataset(filename: str) -> xr.Dataset:
...
-    ds_attrs['forecast_hour'] = _data_array_list[0].attrs['forecast_hour']
-    ds_attrs['start_time'] = _data_array_list[0].attrs['start_time']
-    ds_attrs['end_time'] = _data_array_list[0].attrs['end_time']
-    ds_attrs.update(**dv_units_dict)
+                _level_data_dict[level].append(copied_da)
+
+    _data_array_list = []
+    for level, ds in _level_data_dict.items():
+        if len(ds) == 1:
+            dataset = ds[0].to_dataset(promote_attrs=True)
```
deepgabani8, line 301: xr.merge also works for a list of one dataArray. I don't think we need to have a separate condition for it.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -281,26 +292,43 @@ def __normalize_grib_dataset(filename: str) -> xr.Dataset:
...
+        else:
+            dataset = xr.merge(ds)
+        _data_array_list.append(dataset)
```
deepgabani8, line 304: This can be: `_data_array_list = [xr.merge(list_da) for list_da in _level_data_dict.values()]`

j9sh264: Resolved

weather_mv/loader_pipeline/sinks.py
j9sh264: Resolved

weather_mv/loader_pipeline/sinks.py
```diff
@@ -281,26 +292,43 @@ def __normalize_grib_dataset(filename: str) -> xr.Dataset:
...
-    # Stick the forecast hour, start_time, end_time, data variables units
-    # in the ds attrs as well, that's useful.
-    ds_attrs['forecast_hour'] = _data_array_list[0].attrs['forecast_hour']
-    ds_attrs['start_time'] = _data_array_list[0].attrs['start_time']
-    ds_attrs['end_time'] = _data_array_list[0].attrs['end_time']
-    ds_attrs.update(**dv_units_dict)
+                _level_data_dict[level].append(copied_da)
+
+    _data_array_list = []
+    for level, ds in _level_data_dict.items():
```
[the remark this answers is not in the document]
j9sh264: Resolved

weather_mv/loader_pipeline/sinks.py
j9sh264: Have maintained backwards compatibility with the previous pipelines. The function will return a list of datasets only in case the user passes the `--group_common_hypercubes` flag.

deepgabani8 (approved):

mahrsee1997 (approved):
LGTM.

PR #307, opened 2023-03-27
Skipping ingestion of uris for which the asset file already exists in the GCS bucket

deepgabani8:


weather_mv/loader_pipeline/ee.py
```diff
@@ -364,13 +375,23 @@ def __init__(self,
...
         """Yields uri if the asset does not already exist."""
         self.check_setup()
-
         asset_name = get_ee_safe_name(uri)
+
+        # Checks if the asset is already present in the GCS bucket or not.
+        target_path = os.path.join(
+            self.asset_location, f'{asset_name}{ASSET_TYPE_TO_EXTENSION_MAPPING[self.ee_asset_type]}')
+        if not self.force_overwrite and FileSystems.exists(target_path):
+            logger.info(f'Asset file {target_path} already exists in GCS bucket. Skipping...')
```
fredzyda, line 392: It would appear this is going to just drop the file if the matching COG already exists, regardless of the state of the asset in EE. That's fine, but I think the readme description of what happens could be more clear.

fredzyda (commented):
I think the readme could be a bit more clear about what the flag does. Otherwise it is fine.

weather_mv/README.md
```diff
@@ -322,6 +323,7 @@ _Command options_:
 * `--private_key`: To use a private key for earth engine authentication. Only used with the `service_account` flag.
 * `--xarray_open_dataset_kwargs`: Keyword-args to pass into `xarray.open_dataset()` in the form of a JSON string.
 * `-s, --skip-region-validation` : Skip validation of regions for data migration. Default: off.
+* `-f, --force`: A flag that allows overwriting of existing asset files in the GCS bucket. Default: off.
```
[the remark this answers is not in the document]
deepgabani8: It will not ingest the asset for such uri, and simply skip it.

weather_mv/loader_pipeline/ee.py
deepgabani8: Updated the README statement.

weather_mv/loader_pipeline/ee.py
```diff
@@ -249,6 +253,9 @@ def add_parser_arguments(cls, subparser: argparse.ArgumentParser):
                                help='Skip validation of regions for data migration. Default: off')
         subparser.add_argument('-u', '--use_personal_account', action='store_true', default=False,
                                help='To use personal account for earth engine authentication.')
+        subparser.add_argument('-f', '--force', action='store_true', default=False,
```
alxmrs, line 256: I like the explanation you have in the README. Make sure it is also found here, so that users know all the context when they invoke `-h,--help`.

alxmrs (approved):
One nit, otherwise this LGTM.

weather_mv/loader_pipeline/ee.py
deepgabani8: Updated it 👍

PR #308, opened 2023-04-03
Handling a scenario when the convert-to-asset process terminates unexpectedly

deepgabani8:


weather_mv/loader_pipeline/ee.py
```diff
@@ -520,18 +521,22 @@ def process(self, uri: str) -> t.Iterator[AssetData]:
...
-        while not kill_subprocess:
+        while True:
             if not queue.empty():
                 asset_data = queue.get_nowait()
 
-                if asset_data is not None:
-                    yield asset_data
-                else:
-                    kill_subprocess = True
+                # Not needed now but keeping this check for backwards compatibility.
```
alxmrs, line 529: Why is this not needed?

alxmrs (approved):

weather_mv/loader_pipeline/ee.py
deepgabani8: Now we are checking below if the process is alive or not and the loop will break when the process finishes.

PR #315, opened 2023-04-12
Faster data transfer in `weather-mv rg`.

alxmrs:
Using gsutil, if not gcloud alpha storage, to copy data during regridding. Further, this PR upgrades `debug` level logs to the `info` level.

### DIFF
--- a/weather_mv/loader_pipeline/sinks.py
+++ b/weather_mv/loader_pipeline/sinks.py
@@ -22,6 +22,7 @@ import logging
 import os
 import re
 import shutil
+import subprocess
 import tempfile
 import typing as t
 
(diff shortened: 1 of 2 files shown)


weather_mv/loader_pipeline/sinks.py
```diff
@@ -338,15 +336,27 @@ def __open_dataset_file(filename: str,
...
+    errors = []
+    for cmd in ['gcloud alpha storage cp', 'gsutil cp']:
+        try:
+            subprocess.run(cmd.split() + [src, dst], check=True, capture_output=True)
+            return
+        except subprocess.CalledProcessError as e:
+            errors.append(e)
+
+    msg = f'Failed to copy file {src!r} to {dst!r}'
+    logger.error(f'{msg} due to {errors[0].stderr.decode("utf-8")}')
```
floraxue, line 350: Maybe we should log `errors[-1]` instead?

floraxue (approved):
Thanks for the changes!

weather_mv/loader_pipeline/sinks.py
floraxue: Actually, if both commands are failing, we probably want to log that we've tried both commands and also log both error messages?

PR #316, opened 2023-04-13
Update weather-mv's start_time, end_time parameters to string.

deepgabani8:


weather_mv/loader_pipeline/ee.py
```diff
@@ -182,15 +182,15 @@ class AssetData:
         name: The EE-safe name of the asset.
         target_path: The location of the asset in GCS.
         channel_names: A list of channel names in the asset.
-        start_time: Image start time in floating point seconds since epoch.
-        end_time: Image end time in floating point seconds since epoch.
+        start_time: Image start time in '%Y-%m-%dT%H:%M:%SZ' format.
```
alxmrs, line 185: Is this the same as isoformat? If it is, let's just call it that.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -199,9 +199,11 @@ def _get_band_data(i):
     return ds
 
 
-def _to_utc_timestring(np_time: np.datetime64) -> str:
+def _to_utc_timestring(time: t.Union[np.datetime64, datetime.datetime]) -> str:
     """Turn a numpy datetime64 into UTC timestring."""
-    timestamp = float((np_time - np.datetime64(0, 's')) / np.timedelta64(1, 's'))
+    if isinstance(time, datetime.datetime):
+        return time.strftime('%Y-%m-%dT%H:%M:%SZ')
```
alxmrs, line 205: If this is isofromat, let's use `time.isoformat()`

alxmrs: Same with the line below.

alxmrs (commented):
A nit to try to make the code simpler. Otherwise, this LGTM.

I have a slight hesitation that this will break some pipelines, though I realize it will make weather-mv compatible with others, still.

PR #317, opened 2023-04-14
Adding extensible weather-mv ee transform.

alxmrs:
WIP. Adding a custom transform to make writing Xarray data to EE better.


alxmrs:
I want advice from @deepgabani8 on what the ideal interface for this module should be.

PR #319, opened 2023-04-18
Users can add Pub/Sub subscriptions to `weather-mv` pipelines.

### LINKED ISSUE #318: Add an option to provide `subscription` instead of `topic` for streaming `weather-mv` jobs
While using `weather-mv regrid` as a streaming job, we found that the auto-created subscription by beam in [this line](https://github.com/google/weather-tools/blob/046f675aaa39fccca51d47188995f5f57eae58d0/weather_mv/loader_pipeline/pipeline.py#L64) cannot get messages from the topic. However, attaching a manually created subscription to `beam.ReadFromPubSub` works fine (gets all the messages).

Can we add an option to provide a subscription instead of topic for this case? i.e.
```
if known_args.topic or known_args.subscription:
  if known_args.topic:
    assert known_args.subscription is None, "When topic is provided, subscription must be None."
  if known_args.subscription:
    assert known_args.topic is None, "When subscription is provided, topic must be None."
  beam.ReadFromPubSub(topic=known_args.topic, subscription=known_args.subscription)
```
And corresponding documentations should be updated accordingly.

alxmrs:
Fixes #318.

### DIFF
--- a/weather_mv/README.md
+++ b/weather_mv/README.md
@@ -42,7 +42,9 @@ _Common options_
 
 * `-i, --uris`: (required) URI glob pattern matching input weather data, e.g. 'gs://ecmwf/era5/era5-2015-*.gb'.
 * `--topic`: A Pub/Sub topic for GCS OBJECT_FINALIZE events, or equivalent, of a cloud bucket. E.g.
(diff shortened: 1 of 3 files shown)


floraxue (approved):

alxmrs:
Thanks, Flora!

PR #322, opened 2023-04-27
Added 'dataset' field in manifest.

mahrsee1997:


weather_dl/download_pipeline/manifest.py
```diff
@@ -96,6 +96,9 @@ class DownloadStatus():
     """The name of the config file associated with the request."""
     config_name: str = ""
 
+    """Represents the dataset field of the configuration."""
```
alxmrs, line 99: Sanity check: do we need to update any documentation? My instinct is that the answer is "no".

alxmrs (approved):

weather_dl/download_pipeline/manifest.py
mahrsee1997: You're correct. There is no documentation that needs updating.

mahrsee1997:
Thanks for the review Alex.

Issue #291, opened 2023-02-03
weather-mv: Improve tool's efficiency in terms of time & memory. 


mahrsee1997:
### Time Efficient: 
Make use of `gcloud alpha storage` in open_local() method, sinks.py.

Findings -- using `gsutil` for downloading the data from gcs to the local file system is 5 times slower compared to `gcloud alpha storage`.

### Memory Efficient:
Every time when we log `xr_dataset.nbytes` it will take complete dataset in-memory which is causing OOM killer invocation.
TODO: Find a better way for logging the dataset size.

mahrsee1997:
A temporary fix has been implemented on the mv-optimization branch ([link](https://github.com/google/weather-tools/tree/mv-optimization)). Further work is required to prepare the changes for merging.

mahrsee1997:
Fixed:
* Time Efficiency: #315 
* Memory Efficiency: #323

PR #323, opened 2023-04-27
xarray version update

dabhicusp:
Solution for the TODO of issue #291 `TODO: Find a better way for logging the dataset size.` is **just updating the version of xarray**.

The problem with the current version of xarray which we are using is it takes too much memory while calculating `dataset.nbytes`  (eg:  [`Line #408 weather-mv`](https://github.com/google/weather-tools/blob/main/weather_mv/loader_pipeline/sinks.py#L408)) and kills the process as well. 

To address the problem, I reported the issue on the official xarray GitHub page  [https://github.com/pydata/xarray/issues/7772](https://github.com/pydata/xarray/issues/7772) and they suggested to use the latest version of xarray. After testing, I found that xarray `v2023.1.0` worked well, and we no longer experienced any memory issues.

ps: This is the last `python 3.8` supported version of xarray.

### DIFF
--- a/ci3.8.yml
+++ b/ci3.8.yml
@@ -30,6 +30,6 @@ dependencies:
   - pygrib=2.1.4
   - pip:
     - earthengine-api==0.1.329
-    - xarray=2023.1.0
+    - xarray==2023.1.0
     - ruff==0.0.260
     - .[test]
--- a/ci3.8.yml
+++ b/ci3.8.yml
@@ -16,7 +16,6 @@ dependencies:
   - requests=2.28.1
   - netcdf4=1.6.1
   - rioxarray=0.12.2
-  - xarray=2023.1.0
   - xarray-beam=0.3.1
   - ecmwf-api-client=1.6.3
   - fsspec=2022.10.0
@@ -31,5 +30,6 @@ dependencies:
   - pygrib=2.1.4
   - pip:
     - earthengine-api==0.1.329
+    - xarray=2023.1.0
     - ruff==0.0.260
     - .[test]
--- a/ci3.8.yml
+++ b/ci3.8.yml
@@ -16,7 +16,7 @@ dependencies:
   - requests=2.28.1
   - netcdf4=1.6.1
   - rioxarray=0.12.2
-  - xarray=2022.10.0
+  - xarray=2023.1.0
   - xarray-beam=0.3.1
   - ecmwf-api-client=1.6.3
   - fsspec=2022.10.0
--- a/weather_mv/loader_pipeline/sinks_test.py
+++ b/weather_mv/loader_pipeline/sinks_test.py
@@ -11,12 +11,12 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import contextlib
 import datetime
 from functools import wraps
 import numpy as np
-import os
-import psutil
 import tempfile
+import tracemalloc
 import unittest
 import xarray as xr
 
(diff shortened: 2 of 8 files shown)


weather_mv/loader_pipeline/sinks_test.py
```diff
@@ -38,6 +43,32 @@ def decorated(*args, **kwargs):
     return decorated
 
 
+def generate_dataset() -> str:
```
alxmrs, line 46: Would you mind generating the data with Xarray? It should have the ability built-in to write netCDF files.

weather_mv/loader_pipeline/sinks_test.py
```diff
@@ -67,6 +98,15 @@ def test_opens_tif_files(self):
             self.assertIsNotNone(ds)
             self.assertDictContainsSubset({'is_normalized': False}, ds.attrs)
 
+    def test_open_dataset__fits_memory_bounds(self):
```
alxmrs, line 101: Nice! 

Can you extract this into a decorator or context manager? I think it would be a good test util to have.

alxmrs (commented):
Overall, this looks great! Thanks for the PR.

ci3.8.yml
```diff
@@ -31,5 +30,6 @@ dependencies:
   - pygrib=2.1.4
   - pip:
     - earthengine-api==0.1.329
+    - xarray==2023.1.0
```
alxmrs, line 33: On second though: we may get weird link errors with our conda deploy if we don't include this as a conda dependency. Has this been published on conda forge yet?

weather_mv/loader_pipeline/sinks_test.py
```diff
@@ -38,6 +43,25 @@ def decorated(*args, **kwargs):
     return decorated
 
 
+def generate_dataset() -> str:
+    """Generates temporary netCDF file using xarray."""
+    lat_dim = 3210
+    lon_dim = 3440
+    lat = [0] * lat_dim
+    lon = [0] * lon_dim
```
alxmrs, line 51: Let's do something like this instead: 
```suggestion
    lat_dim = 3210
    lon_dim = 3440
    lat = np.linspace(-90, 90, lat_dim)
    lon = np.linspace(-180, 180, lon_dim)
```

weather_mv/loader_pipeline/sinks_test.py
```diff
@@ -38,6 +43,25 @@ def decorated(*args, **kwargs):
     return decorated
 
 
+def generate_dataset() -> str:
```
alxmrs, line 46: Let's name this something like "write_netcdf".

alxmrs: Looking at how this is used: what do you think about making this into a contextmanager? The difference, besides adding a decorator, is that we'd `yield` from the tempfile `with` statement. The benefit is that we'd get cleanup "for free". Would this negatively affect your memory test?

alxmrs (commented):
Added a concern.

weather_mv/loader_pipeline/sinks_test.py
dabhicusp: Appropriate Changes were done.

weather_mv/loader_pipeline/sinks_test.py
dabhicusp: Added context manager in `test_open_dataset__fits_memory_bounds`.

ci3.8.yml
dabhicusp: While trying to install the `xarray` package via the `conda-forge` channel, I encountered an error when running the installation in the **CI/CD pipeline**. However, the installation process worked without any issues when I tried it **locally**. Here's an error message that I received during the CI/CD run:
image
 [**Link**](https://github.com/google/weather-tools/actions/runs/4829131037/jobs/8603795741)
Do you have any suggestions for resolving this issue?

alxmrs: It looks like 2023.4.2 is available on conda (I think this is the latest release). How about we try upgrading to that version? 

https://github.com/conda-forge/xarray-feedstock#current-release-info

dabhicusp: As mentioned in the PR Description `This is the last updated version of xarray that python 3.8 supported`.

dabhicusp: "Hello @alxmrs, I have tested both the `local` and `docker` deployment and did not encounter any errors."

alxmrs: Thanks for confirming!

alxmrs (approved):
LGTM

PR #325, opened 2023-05-03
Fixed `gcloud alpha command` interactive prompt.

### LINKED ISSUE #324: `gcloud alpha commands` used but not installed in enviroment
**What is the issue?**
`gcloud alpha storage cp` is being used in `weather-mv` but it is not installed in conda enviroment.

https://github.com/google/weather-tools/blob/374153ba3bc34491ca414b88e3004003e63ec852/weather_mv/loader_pipeline/sinks.py#L344-L349

**Bug this is causing**
The `weather-mv` pipeline hangs at this command as gcloud tries to install it and waits for user input.
```
$ python weather_mv/weather-mv bq -i "/local/path/data.nc" -o  myproject.mydataset.mytable --dry-run
warning: sdist: standard file not found: should have one of README, README.rst, README.txt, README.md

INFO:root:Missing pipeline option (runner). Executing pipeline using the default runner: DirectRunner.
```

aniketinfocusp:
Possible solution for this was adding the `gcloud components install alpha --quiet` to `weather_mv/setup.py`.
Ref-
https://github.com/google/weather-tools/blob/f2a255460ef23e960a7fc669ecdc27ff7d58c9c9/weather_mv/setup.py#L99-L104
But going forward we would be moving to conda instead of pip hence setup files would be removed eventually.

For now, the `subprocess.run()` command is passed with a negative input for any user prompt, this will not let the program hang at user input. Fixes #324.

Additionally, for the docker image, gcloud alpha commands are explicitly installed so that it's already in the environment when someone runs the tool on Dataflow providing the docker image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,10 @@ ARG CONDA_ENV_NAME=weather-tools
 RUN echo "source activate ${CONDA_ENV_NAME}" >> ~/.bashrc
 ENV PATH /opt/conda/envs/${CONDA_ENV_NAME}/bin:$PATH
 
+# Install gcloud alpha
+RUN apt-get update -y
+RUN gcloud components install alpha --quiet
+
 # Copy files from official SDK image, including script/dependencies.
 COPY --from=beam_sdk /opt/apache/beam /opt/apache/beam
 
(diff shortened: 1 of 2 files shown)


alxmrs (approved):
LGTM! Has the deployment of this been tested so far?

aniketinfocusp:
> LGTM! Has the deployment of this been tested so far?

Yes, it has been tested end to end. Also created a docker image locally and it had installed gcloud alpha.

alxmrs:
Excellent, thank you!

PR #326, opened 2023-05-03
fixed support for floating point numbers when specifying geographic region

### LINKED ISSUE #183: Support floating point numbers when specifying geographic region
Users should be able to specify the floating point number while specifying `-a, --area: Target area in [N, W, S, E].`

Eg: --area 49.34 -124.68 24.74 -66.95

Piyush-Ingale:
The issue  #183 was users were not able to pass floating point for ``-a, --area`` argument. It was throwing error `weather-mv bigquery: error: argument -a/--area: invalid int value: '49.34'.`  This error was because the type for argument was specified as `int.`

changing the type for argument `-a, --area` to `float`  Fixes #183

### DIFF
--- a/weather_mv/loader_pipeline/bq.py
+++ b/weather_mv/loader_pipeline/bq.py
@@ -87,7 +87,7 @@ class ToBigQuery(ToDataSink):
     """
     output_table: str
     variables: t.List[str]
-    area: t.Tuple[int, int, int, int]
+    area: t.List[float]
     import_time: t.Optional[datetime.datetime]
     infer_schema: bool
     xarray_open_dataset_kwargs: t.Dict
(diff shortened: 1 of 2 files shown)


weather_mv/loader_pipeline/bq.py
```diff
@@ -87,7 +87,7 @@ class ToBigQuery(ToDataSink):
     """
     output_table: str
     variables: t.List[str]
-    area: t.Tuple[int, int, int, int]
+    area: t.List[float]
```
alxmrs, line 90: This LGTM. A quick question: is there early validation to check that users must specify 4 floats?

alxmrs (approved):

weather_mv/loader_pipeline/bq.py
Piyush-Ingale: Yes, there is [validation](https://github.com/google/weather-tools/blob/e67f361899aebe6efaac23628cbe75bd68ce31c5/weather_mv/loader_pipeline/bq.py#L133-L135) that user must specify 4 values for `--area`.

alxmrs: Thanks, perfect.

PR #327, opened 2023-05-04
Unit Tests for JSON Serialization

### LINKED ISSUE #106: Thoroughly unit tests JSON Serialization
Check out TODO in `unit_test.py`.

aniketinfocusp:
Fixes #106. Unit tests for `to_json_serializable_type()` in `weather_mv/loader_pipeline/utils.py`. 

The current implementation of `to_json_serializable_type()`  throws error when a `ndarray` of size > 1 is passed.  Also, it did not handle arrays with NaNs and numbers mixed. 
Both of these issues are fixed in this PR.

### DIFF
--- a/weather_mv/loader_pipeline/util.py
+++ b/weather_mv/loader_pipeline/util.py
@@ -86,11 +86,11 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
     logger.debug('Serializing to JSON')
 
     # pd.isna() returns ndarray if input is not scalar therefore checking if not ndarray.
-    if type(value) != np.ndarray and (pd.isna(value) or value is None):
+    if (np.isscalar(value) and pd.isna(value)) or value is None:
         return None
     elif np.issubdtype(type(value), np.floating):
         return float(value)
-    elif type(value) == np.ndarray:
+    elif isinstance(value, np.ndarray):
         # Will return a scaler if array is of size 1, else will return a list.
         # Replace all NaNs, NaTs with None.
         return np.where(pd.isna(value), None, value).tolist()
--- a/weather_mv/loader_pipeline/util.py
+++ b/weather_mv/loader_pipeline/util.py
@@ -85,13 +85,13 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
     # Note: The order of processing is significant.
     logger.debug('Serializing to JSON')
 
-    if pd.isna(value) or value is None:
+    if type(value) != np.ndarray and (pd.isna(value) or value is None):
         return None
     elif np.issubdtype(type(value), np.floating):
         return float(value)
     elif type(value) == np.ndarray:
         # Will return a scaler if array is of size 1, else will return a list.
-        return value.tolist()
+        return np.where(pd.isna(value), None, value).tolist()
     elif type(value) == datetime.datetime or type(value) == str or type(value) == np.datetime64:
         # Assume strings are ISO format timestamps...
         try:
--- a/weather_mv/loader_pipeline/util.py
+++ b/weather_mv/loader_pipeline/util.py
@@ -85,12 +85,14 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
     # Note: The order of processing is significant.
     logger.debug('Serializing to JSON')
 
+    # pd.isna() returns ndarray is input is not scalar therefore checking if not ndarray.
     if type(value) != np.ndarray and (pd.isna(value) or value is None):
         return None
     elif np.issubdtype(type(value), np.floating):
         return float(value)
     elif type(value) == np.ndarray:
         # Will return a scaler if array is of size 1, else will return a list.
+        # Replace all NaNs, NaTs with None
         return np.where(pd.isna(value), None, value).tolist()
     elif type(value) == datetime.datetime or type(value) == str or type(value) == np.datetime64:
         # Assume strings are ISO format timestamps...
--- a/weather_mv/loader_pipeline/util.py
+++ b/weather_mv/loader_pipeline/util.py
@@ -92,7 +92,7 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
         return float(value)
     elif type(value) == np.ndarray:
         # Will return a scaler if array is of size 1, else will return a list.
-        # Replace all NaNs, NaTs with None
+        # Replace all NaNs, NaTs with None.
         return np.where(pd.isna(value), None, value).tolist()
     elif type(value) == datetime.datetime or type(value) == str or type(value) == np.datetime64:
         # Assume strings are ISO format timestamps...
--- a/weather_mv/loader_pipeline/util.py
+++ b/weather_mv/loader_pipeline/util.py
@@ -85,7 +85,7 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
     # Note: The order of processing is significant.
     logger.debug('Serializing to JSON')
 
-    # pd.isna() returns ndarray if input is not scalar therefore checking if not ndarray.
+    # pd.isna() returns ndarray if input is not scalar therefore checking if value is scalar.
     if (np.isscalar(value) and pd.isna(value)) or value is None:
         return None
     elif np.issubdtype(type(value), np.floating):
--- a/weather_mv/loader_pipeline/util.py
+++ b/weather_mv/loader_pipeline/util.py
@@ -90,11 +90,14 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
         return None
     elif np.issubdtype(type(value), np.floating):
         return float(value)
+    elif isinstance(value, set):
+        value = list(value)
+        return np.where(pd.isna(value), None, value).tolist()
     elif isinstance(value, np.ndarray):
         # Will return a scaler if array is of size 1, else will return a list.
         # Replace all NaNs, NaTs with None.
         return np.where(pd.isna(value), None, value).tolist()
-    elif type(value) == datetime.datetime or type(value) == str or type(value) == np.datetime64:
+    elif isinstance(value, datetime.datetime) or isinstance(value, str) or isinstance(value, np.datetime64):
         # Assume strings are ISO format timestamps...
         try:
             value = datetime.datetime.fromisoformat(value)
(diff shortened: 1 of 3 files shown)


weather_mv/loader_pipeline/util.py
```diff
@@ -85,13 +85,15 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
     # Note: The order of processing is significant.
     logger.debug('Serializing to JSON')
 
-    if pd.isna(value) or value is None:
+    # pd.isna() returns ndarray if input is not scalar therefore checking if not ndarray.
+    if type(value) != np.ndarray and (pd.isna(value) or value is None):
```
alxmrs, line 89: What do you think about the following? 
```suggestion
    if (np.isscalar(value) and pd.isna(value)) or value is None:
```

If you prefer your method, my recommendation is to use `isinstance()` instead of type equality. I've been meaning to clean this up for a bit.

weather_mv/loader_pipeline/util.py
```diff
@@ -85,13 +85,15 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
...
         return None
     elif np.issubdtype(type(value), np.floating):
         return float(value)
     elif type(value) == np.ndarray:
         # Will return a scaler if array is of size 1, else will return a list.
-        return value.tolist()
+        # Replace all NaNs, NaTs with None.
```
alxmrs, line 95: As mentioned above: can we replace all `type(x) == klass` with `isinstance(x, klass)`?

alxmrs, line 95: Nice! Good thinking.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -178,5 +185,99 @@ def test_make_attrs_ee_compatible_a2(self):
...
+
+    def test_to_json_serializable_type_none(self):
+        attrs = None
+
+        expected = None
+
+        actual = to_json_serializable_type(attrs)
+
+        self.assertEqual(actual, expected)
+        self.assertEqual(type(actual), type(expected))
```
alxmrs, line 197: I don't think we need to include a type check; equality should be sufficient

weather_mv/loader_pipeline/util_test.py
```diff
@@ -178,5 +185,99 @@ def test_make_attrs_ee_compatible_a2(self):
...
-    pass
+
+    def test_to_json_serializable_type_none(self):
+        attrs = None
+
+        expected = None
+
+        actual = to_json_serializable_type(attrs)
+
+        self.assertEqual(actual, expected)
```
alxmrs, line 196: I know AAA is standard, but I think it would be more readable if we had a series of checks that all fit on one line that were mostly the same, where we only needed to check in the value being compared against. For example: 

```suggestion
    self.assertIsNone(self.convert(None))
    self.assertIsNone(self.convert(float('NaN')))
    self.assertIsNone(self.convert(np.nan))
    self.assertIsNotNone(self.convert([]))
```

weather_mv/loader_pipeline/util_test.py
```diff
@@ -178,5 +185,99 @@ def test_make_attrs_ee_compatible_a2(self):
...
+        self.assertEqual(type(actual), type(expected))
+
+    def test_to_json_serializable_type_npfloat(self):
+        attrs = np.float32(1)
+
+        expected = float(1)
+
+        actual = to_json_serializable_type(attrs)
```
alxmrs, line 204: Similar to my suggestion above. To help you out, here are a bunch of cases to check: 

```suggestion
    self.assertIsInstance(self.convert(0.1), float)
    self.assertIsInstance(self.convert(np.float32('0.1')), float)
    self.assertIsInstance(self.convert(np.float32('1')), float)
    self.assertIsInstance(self.convert(np.float16('0.1')), float)
    self.assertIsInstance(self.convert(np.single('0.1')), float)
    self.assertIsInstance(self.convert(np.double('0.1')), float)
    self.assertNotIsInstance(self.convert(1), float)
    self.assertNotIsInstance(self.convert(np.csingle('0.1')), float)
    self.assertNotIsInstance(self.convert(np.cdouble('0.1')), float)
    self.assertNotIsInstance(self.convert(np.intc('1')), float)
```

weather_mv/loader_pipeline/util_test.py
```diff
@@ -178,5 +185,99 @@ def test_make_attrs_ee_compatible_a2(self):
...
+        self.assertEqual(type(actual), type(expected))
+
+    def test_to_json_serializable_type_ndarray(self):
+        attrs = np.arange(5)
+
+        expected = [0, 1, 2, 3, 4]
+
+        actual = to_json_serializable_type(attrs)
+
+        self.assertEqual(actual, expected)
```
alxmrs, line 216: OK, I'll make an admission: I've written tests for these internally, but I have not yet had the time to port them to be external. I've not made these up on the fly, and I really appreciate that you're taking the initiative to fix this open issue :)

```suggestion
        self.assertIsInstance(self.convert(np.array(list(range(10)))), list)
    self.assertEqual(self.convert(np.array(list(range(10)))), list(range(10)))
    self.assertEqual(self.convert(np.array([1])), [1])
    self.assertEqual(
        self.convert(np.array([[1, 2, 3], [4, 5, 6]])), [[1, 2, 3], [4, 5, 6]]
    )
```

weather_mv/loader_pipeline/util_test.py
```diff
@@ -178,5 +185,99 @@ def test_make_attrs_ee_compatible_a2(self):
...
+        self.assertEqual(type(actual), type(expected))
+
+    def test_to_json_serializable_type_ndarray_sizeone(self):
+        attrs = np.arange(1)
```
alxmrs, line 220: This is a good idea for a test.

weather_mv/loader_pipeline/util_test.py
```diff
@@ -178,5 +185,99 @@ def test_make_attrs_ee_compatible_a2(self):
...
+    def test_to_json_serializable_type_nptimedelta(self):
+        attrs = np.timedelta64(1, 'm')  # timedelta of 1 minute.
+
+        expected = float(60)  # seconds.
+
+        actual = to_json_serializable_type(attrs)
+
+        self.assertEqual(actual, expected)
+        self.assertEqual(type(actual), type(expected))
+
```
alxmrs, line 274: Here are my internal tests that you're free to port here: 

```

  def test_time_to_isostring(self):
    now = datetime.datetime.now()
    now_str = now.isoformat()
    now_np = np.datetime64(now)

    self.assertEqual(self.convert(now), self.convert(now_str))
    self.assertEqual(self.convert(now), self.convert(now_np))

  def test_np_timedelta_to_float(self):
    delta = datetime.timedelta(seconds=33)
    delta_np = np.timedelta64(delta)
    self.assertEqual(self.convert(delta_np), delta.total_seconds())

  def test_np_integer_to_int(self):
    self.assertEqual(self.convert(np.int64(1)), 1)
    self.assertEqual(self.convert(np.int64(-10_000)), -10_000)
    self.assertEqual(self.convert(np.uint16(25)), 25)
```

weather_mv/loader_pipeline/util_test.py
```diff
@@ -178,5 +185,99 @@ def test_make_attrs_ee_compatible_a2(self):
...
+    def test_to_json_serializable_type_npinteger(self):
```
alxmrs, line 275: I like this idea for a test. Can you think of more corner cases to check?

alxmrs (commented):
Added some feedback. Thanks for your fixes here!

weather_mv/loader_pipeline/util.py
aniketinfocusp: I think using `np.isscalar` looks much cleaner and it handles other sequence objects like sets and tuples.

weather_mv/loader_pipeline/util_test.py
aniketinfocusp: I agree, it looks much better and compressed when we have the checks inline. I went ahead and created a wrapper function `_convert` to wrap `to_json_serializable_type`.

weather_mv/loader_pipeline/util.py
aniketinfocusp: Updated in https://github.com/google/weather-tools/pull/327/commits/dac23a047e2357b4f0dce9fb0b17969ae7f972a5.

weather_mv/loader_pipeline/util.py
aniketinfocusp: Updated in https://github.com/google/weather-tools/pull/327/commits/dac23a047e2357b4f0dce9fb0b17969ae7f972a5.

weather_mv/loader_pipeline/util_test.py
aniketinfocusp: Updated in https://github.com/google/weather-tools/pull/327/commits/d9cf97add436e462d01fb43fe2d32ac7c0511d17.

weather_mv/loader_pipeline/util_test.py
aniketinfocusp: Added in https://github.com/google/weather-tools/pull/327/commits/d9cf97add436e462d01fb43fe2d32ac7c0511d17.

weather_mv/loader_pipeline/util_test.py
aniketinfocusp: Thanks for these cases. Added in https://github.com/google/weather-tools/pull/327/commits/d9cf97add436e462d01fb43fe2d32ac7c0511d17.

weather_mv/loader_pipeline/util_test.py
aniketinfocusp: I added some more I could think of. Added in https://github.com/google/weather-tools/pull/327/commits/d9cf97add436e462d01fb43fe2d32ac7c0511d17.

weather_mv/loader_pipeline/util_test.py
aniketinfocusp: Thanks.

weather_mv/loader_pipeline/util_test.py
aniketinfocusp: I have added some more cases (and used all that you have provided).

weather_mv/loader_pipeline/util.py
```diff
@@ -85,14 +85,19 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
     # Note: The order of processing is significant.
     logger.debug('Serializing to JSON')
 
-    if pd.isna(value) or value is None:
+    # pd.isna() returns ndarray if input is not scalar therefore checking if not ndarray.
```
alxmrs, line 88: Comment is a bit stale.

alxmrs (approved):
LGTM, thanks!

weather_mv/loader_pipeline/util.py
aniketinfocusp: updated in https://github.com/google/weather-tools/pull/327/commits/af655e7d4c63b3dc2e39e87271f5528016981d50

PR #301, opened 2023-03-07
`weather-mv bq` can now load Zarr into BigQuery.

alxmrs:
This CL adds a Zarr reading option to `open_dataset` and to the BigQuery data sink. In addition, we provide some small refactoring to the BigQuery transform to make it more succinct.

PR #333, opened 2023-05-22
Updated: weather-mv bq can now load Zarr into BigQuery.

DarshanSP19:
Follow Up PR for https://github.com/google/weather-tools/pull/301.

This CL adds a Zarr reading option to open_dataset and to the BigQuery data sink. In addition, we provide some small refactoring to the BigQuery transform to make it more succinct.


weather_mv/loader_pipeline/pipeline.py
```diff
@@ -54,9 +54,7 @@ def pipeline(known_args: argparse.Namespace, pipeline_args: t.List[str]) -> None
     known_args.first_uri = next(iter(all_uris))
 
     with beam.Pipeline(argv=pipeline_args) as p:
-        if known_args.zarr:
-            paths = p
```
alxmrs, line 58: Isn't this needed? How do we select a single Zarr URI?

alxmrs (approved):
Happy to merge this if my question is answered and the tests pass. 

Has this been tested e2e?

weather_mv/loader_pipeline/pipeline.py
DarshanSP19: For above query following URI Patterns will work for `zarr`.
- `../test_data/test_data.zarr` (For a single zarr)
- `../test_data/*.zarr` (For multiple zarr in a directory)

DarshanSP19: This will be handled in `else` part of this condition.
Currently it's working for a single `zarr` file.

But for URI like `gs://gcp-public-data-arco-era5/co/model-level-*.zarr` it's not running as intended.

alxmrs: Thanks for the ptr! That makes sense to me. 

For that test Zarr: I'm happy to meet over VC to debug what's going on.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -387,7 +387,7 @@ def open_dataset(uri: str,
     """Open the dataset at 'uri' and return a xarray.Dataset."""
     try:
         if is_zarr:
-            ds = xr.open_dataset(uri, engine='zarr', chunks=None)
+            ds: xr.Dataset = xr.open_dataset(uri, engine='zarr', **open_dataset_kwargs)
```
alxmrs, line 390: Are `chunks=None` set by default? It's important to make sure this is the default behavior, since it will open the Zarr lazily.

weather_mv/loader_pipeline/bq.py
```diff
@@ -150,6 +150,8 @@ def validate_arguments(cls, known_args: argparse.Namespace, pipeline_args: t.Lis
 
     def __post_init__(self):
         """Initializes Sink by creating a BigQuery table based on user input."""
+        if self.zarr:
+            self.xarray_open_dataset_kwargs = self.zarr_kwargs
```
alxmrs, line 154: Does this let us specify a range of years to open? If so, that's excellent and general purpose!

DarshanSP19: Currently don't have that kind of support. But we can introduce a separate flag specific to `zarr` dataset to specify the time range. What do you think?

alxmrs: Yes, that would be great. For now, I've seen that the pipeline runs for a quite a while with no apparent errors. I say: let's merge this CL for now, and implement the time range flags in a future PR.

PR #336, opened 2023-05-29
Refactored DB layer

aniketinfocusp:
Updated Db layer:
- Separated out database operations based on collections
- Removed global db connection (added it on route level)
- Used dependency injection to get data handling classes


mahrsee1997:
Not required as of now.

Issue #266, opened 2022-12-06
The BigQuery schema should ingest grid points as polygons (covering the area a point represents). 


alxmrs:
Inspired by this [post], instead of representing a grid point in a BigQuery dataset as an infinitesimally small point, we should represent it as a polygon that covers a specific area. This should help users understand the coarseness of grids, as well as apply geospatial SQL operators in BigQuery. 

Thanks @rabernat for the idea. 

[post]: https://discourse.pangeo.io/t/conservative-region-aggregation-with-xarray-geopandas-and-sparse/2715

blackvvine:
IIUC the idea is to use a [`geojson.Polygon`](https://python-geojson.readthedocs.io/en/latest/#polygon) instead of a `Point` in the `geo_point` field:

https://github.com/google/weather-tools/blob/67affd68b509f22d3761c9e1c131ebc7c5ccd46e/weather_mv/loader_pipeline/bq.py#L333

If that is the case, what is the logic for determining the area boundaries? Do we trivially take the middle of adjacent coordinates?

In the [post](https://discourse.pangeo.io/t/conservative-region-aggregation-with-xarray-geopandas-and-sparse/2715), they are [using](https://gist.github.com/rsignell-usgs/273d5736e522017cf5d4e0d7ae691b76) a dataset which specifies the area boundaries explicitly with `lot_bounds`, `lat_bounds`, and `nv`.

alxmrs:
Thinking about it some more: Can we add a polygon instead of replacing the point? 

For the polygon: the lat/lng point should be the center. We should check in common places within the XArray metadata (maybe, add a flag to parse arbitrary locations?) to look up the range of a grid point. 

We can more or less assume the data follows ECMWF's conventions. We aim to apply this to this dataset in particular: https://github.com/google-research/arco-era5

IIUC, I believe their coordinates are defined by the CF conventions: https://cfconventions.org/cf-conventions/cf-conventions.html#grid-mappings-and-projections

@wundersooner: I think we can decipher a way to make this issue work for ARCO datasets, but do you have any advise about what metadata we could look for to build this in a general way?

wundersooner:
[...]

PR #337, opened 2023-05-29
Ingested grid point as a polygon in BigQuery.

dabhicusp:
Hi @alxmrs, as per our earlier conversation in discussion #266, I have implemented the code to ingest grid points as `polygons` in BigQuery. Currently, the code creates a polygon by considering the given `latitude` and `longitude` as the `center` of the polygon. 

Note: This implementation only works for a regular_ll grid.


weather_mv/loader_pipeline/bq.py
```diff
@@ -305,6 +328,17 @@ def extract_rows(uri: str,
 
     with open_dataset(uri, open_dataset_kwargs, disable_grib_schema_normalization,
                       tif_metadata_for_datetime) as ds:
+
+        # Find the grid_resolution. In case of single point we can't find grid resolution.
+        should_create_polygon = False
+        if ds['latitude'].size > 1 and ds['longitude'].size > 1:
+            # consider that Grid is regular.
```
alxmrs, line 335: Maybe, launch a warning log message with this? Related: in the BQ constructor, let's add a check on the test data point to ensure that the grid is regular. If it isn't, let's omit adding the polygon.

alxmrs: Last thought on this: In the constructor, if we can't create polygons, let's add a warning log message.

alxmrs (commented):
Looks good so far!

weather_mv/loader_pipeline/bq.py
```diff
@@ -266,6 +268,27 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+    """Create a Polygon based on latitude and longitude and resolution."""
+    lower_left = [latitude - lat_grid_resolution, longitude - lon_grid_resolution]
+    upper_left = [latitude - lat_grid_resolution, longitude + lon_grid_resolution]
+    upper_right = [latitude + lat_grid_resolution, longitude + lon_grid_resolution]
+    lower_right = [latitude + lat_grid_resolution, longitude - lon_grid_resolution]
+    lat_lon_bound = [lower_left, upper_left, upper_right, lower_right]
+
+    for i, _ in enumerate(lat_lon_bound):
+        if lat_lon_bound[i][1] >= 180:
+            lat_lon_bound[i][1] = lat_lon_bound[i][1] - 360
```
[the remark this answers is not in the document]
wundersooner: I would be cautious here.  Some coordinate systems use x,y coordinates in meters instead of degree-based lon, lat.

weather_mv/loader_pipeline/bq.py
```diff
@@ -266,6 +268,27 @@ def fetch_geo_point(lat: float, long: float) -> str:
     return point
 
 
+def fetch_geo_polygon(latitude: float, longitude: float, lat_grid_resolution: float, lon_grid_resolution: float) -> str:
+    """Create a Polygon based on latitude and longitude and resolution."""
```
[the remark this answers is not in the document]
dabhicusp: necessary changes were done.

weather_mv/loader_pipeline/bq.py
```diff
@@ -266,6 +268,27 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+
+    for i, _ in enumerate(lat_lon_bound):
+        if lat_lon_bound[i][1] >= 180:
+            lat_lon_bound[i][1] = lat_lon_bound[i][1] - 360
+    polygon = geojson.dumps(geojson.Polygon([
+        (lat_lon_bound[0][0], lat_lon_bound[0][1]),
+        (lat_lon_bound[1][0], lat_lon_bound[1][1]),
+        (lat_lon_bound[2][0], lat_lon_bound[2][1]),
+        (lat_lon_bound[3][0], lat_lon_bound[3][1]),
+        (lat_lon_bound[0][0], lat_lon_bound[0][1])
```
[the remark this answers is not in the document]
dabhicusp: Appropriate changes are done.

weather_mv/loader_pipeline/bq.py
```diff
@@ -310,3 +331,24 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+
+    for i in range(len(lat_lon_bound)):
+        if lat_lon_bound[i][1] >= 180:
+            lat_lon_bound[i][1] = lat_lon_bound[i][1] - 360
+    polygon = geojson.dumps(geojson.Polygon([
+        (lat_lon_bound[0][0], lat_lon_bound[0][1]),  # lower_left
+        (lat_lon_bound[1][0], lat_lon_bound[1][1]),  # upper_left
+        (lat_lon_bound[2][0], lat_lon_bound[2][1]),  # upper_right
+        (lat_lon_bound[3][0], lat_lon_bound[3][1]),  # lower_right
+        (lat_lon_bound[0][0], lat_lon_bound[0][1]),  # lower_left
```
alxmrs, line 352: Thanks!

weather_mv/loader_pipeline/bq.py
```diff
@@ -154,6 +158,20 @@ def __post_init__(self):
...
                           is_zarr=self.zarr) as open_ds:
+
+            # Find the grid_resolution. In case of single point we can't find grid resolution.
+            if open_ds['latitude'].size > 1 and open_ds['longitude'].size > 1:
+                # consider that Grid is regular.
+                latitude_length = len(open_ds['latitude'])
+                longitude_length = len(open_ds['longitude'])
+                self.lat_grid_resolution = abs((open_ds["latitude"][-1].values - open_ds["latitude"][0].values
+                                                )/latitude_length)/2
+                self.lon_grid_resolution = abs((open_ds["longitude"][-1].values - open_ds["longitude"][0].values
```
alxmrs, line 169: I think aI think there may be a succinct way to do this in numpy. For example, consider using this method: 

https://numpy.org/doc/stable/reference/generated/numpy.ptp.html

weather_mv/loader_pipeline/bq.py
```diff
@@ -154,6 +158,20 @@ def __post_init__(self):
...
+                                                )/longitude_length)/2
+                self.should_create_polygon = True
+            else:
+                logger.warning("Polygon can't be genereated as dataset has a single point.")
```
alxmrs, line 173: Is this case common?

weather_mv/loader_pipeline/bq.py
```diff
@@ -154,6 +158,20 @@ def __post_init__(self):
         with open_dataset(self.first_uri, self.xarray_open_dataset_kwargs,
                           self.disable_grib_schema_normalization, self.tif_metadata_for_datetime,
                           is_zarr=self.zarr) as open_ds:
+
+            # Find the grid_resolution. In case of single point we can't find grid resolution.
+            if open_ds['latitude'].size > 1 and open_ds['longitude'].size > 1:
+                # consider that Grid is regular.
```
alxmrs, line 164: That the latitudes and longitudes are larger than 1 doesn't seem like that rigorous of a test that the grid is regular.

alxmrs: Maybe the comment and log message is wrong?

weather_mv/loader_pipeline/bq.py
```diff
@@ -309,3 +330,49 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+        (lat_lon_bound[1][0], lat_lon_bound[1][1]),  # upper_left
+        (lat_lon_bound[2][0], lat_lon_bound[2][1]),  # upper_right
+        (lat_lon_bound[3][0], lat_lon_bound[3][1]),  # lower_right
+        (lat_lon_bound[0][0], lat_lon_bound[0][1]),  # lower_left
+    ]))
+    return polygon
+
+
+def bound_point(latitude, longitude, lat_grid_resolution, lon_grid_resolution) -> t.List:
+    '''Calculate the bound point based on latitude, longitude and grid resolution.'''
```
alxmrs, line 351: Nit, please use `"""`.

weather_mv/loader_pipeline/bq.py
```diff
@@ -309,3 +330,49 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+    is_lat_out_of_bound = True if latitude in [90.0, -90.0] else False
+    is_lon_out_of_bound = True if longitude in [-180.0, 180.0] else False
```
alxmrs, line 353: There is no need for an if statement here. These could simply be `latitude in [90.0, -90.0]`, etc.

alxmrs: maybe, `bool(latitude in [90.0, -90.0])`

alxmrs: Also: is this variable named correctly? Shouldn't it be "latitude_in_bounds"? Or, do I misunderstand?

alxmrs (commented):
Review part 1...

weather_mv/loader_pipeline/bq.py
```diff
@@ -309,3 +330,49 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+                                  lat_grid_resolution, lon_grid_resolution)
+    lower_left = [lon_range[1], lat_range[1]]
+    upper_left = [lon_range[0], lat_range[1]]
+    upper_right = [lon_range[0], lat_range[0]]
+    lower_right = [lon_range[1], lat_range[0]]
+    return [lower_left, upper_left, upper_right, lower_right]
+
+
+def get_lat_lon_range(value, lat_lon, is_point_out_of_bound, lat_grid_resolution, lon_grid_resolution):
+    '''Calculate the latitude, longitude point range point latitude, longitude and grid resolution.'''
```
alxmrs, line 367: Same here: `"""`.

weather_mv/loader_pipeline/bq.py
```diff
@@ -309,3 +330,49 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+    lon_range = get_lat_lon_range(longitude, "longitude", is_lon_out_of_bound,
+                                  lat_grid_resolution, lon_grid_resolution)
+    lower_left = [lon_range[1], lat_range[1]]
+    upper_left = [lon_range[0], lat_range[1]]
+    upper_right = [lon_range[0], lat_range[0]]
+    lower_right = [lon_range[1], lat_range[0]]
+    return [lower_left, upper_left, upper_right, lower_right]
+
+
+def get_lat_lon_range(value, lat_lon, is_point_out_of_bound, lat_grid_resolution, lon_grid_resolution):
```
alxmrs, line 366: I think the logic for normalizing the lat/lng ranges to spread out across too many places. Can we correct for it once near the logical beginning? 

For example, can we just compute the following line before getting the input lat/lng values? 

https://github.com/google/weather-tools/pull/337/files#diff-938d4c3999aa9d0bd9f832240de92dc8ea660bf71f8b9b6d751b3a16773ce4b1L309

weather_mv/loader_pipeline/bq.py
alxmrs, line 367: Same here: `"""`.

alxmrs (commented):
Review part 2. (Splitting up the review b/c I have intermittent internet).

weather_mv/loader_pipeline/bq.py
```diff
@@ -94,6 +95,9 @@ class ToBigQuery(ToDataSink):
     skip_region_validation: bool
     disable_grib_schema_normalization: bool
     coordinate_chunk_size: int = 10_000
+    should_create_polygon: bool = False
+    lat_grid_resolution: t.Optional[float] = None
+    lon_grid_resolution: t.Optional[float] = None
 
     @classmethod
     def add_parser_arguments(cls, subparser: argparse.ArgumentParser):
```
alxmrs, line 103: Let's add a flag in the subparser to let users turn this feature on or off.

weather_mv/loader_pipeline/bq.py
dabhicusp: Code updated as per your suggestion. Thanks for the input!

weather_mv/loader_pipeline/bq.py
dabhicusp: Code updated as per your suggestion. Thanks for the input!

weather_mv/loader_pipeline/bq.py
dabhicusp: Code updated as per your suggestion. Thanks for the input!

weather_mv/loader_pipeline/bq.py
dabhicusp: Code updated as per your suggestion. Thanks for the input!

dabhicusp: Code updated as per your suggestion. Thanks for the input!

weather_mv/loader_pipeline/bq.py
dabhicusp: Code updated as per your suggestion. Thanks for the input!

weather_mv/loader_pipeline/bq.py
dabhicusp: Code updated as per your suggestion. Thanks for the input!

weather_mv/loader_pipeline/bq.py
dabhicusp: Hey @alxmrs, I wanted to discuss the different possibilities in our pipeline when the user provides a flag(`create_polygon`) but the input file contains grid points as a **single point**. Here are the options we have:

1. We can display a warning message to the user, ingest `None` in the Polygon, and allow the pipeline to continue running smoothly.
2. Alternatively, we can give an error to the user at this stage and terminate the pipeline.

Currently, I have implemented the `first option` in the code. I would appreciate your input on whether this approach is suitable, or if we should consider moving towards the `second option`.

weather_mv/loader_pipeline/bq.py
```diff
@@ -94,6 +95,10 @@ class ToBigQuery(ToDataSink):
     skip_region_validation: bool
     disable_grib_schema_normalization: bool
     coordinate_chunk_size: int = 10_000
+    create_polygon: bool = False
+    should_create_polygon: bool = False
```
alxmrs, line 99: Can this be one flag?

weather_mv/loader_pipeline/bq.py
```diff
@@ -154,6 +163,24 @@ def __post_init__(self):
         with open_dataset(self.first_uri, self.xarray_open_dataset_kwargs,
                           self.disable_grib_schema_normalization, self.tif_metadata_for_datetime,
                           is_zarr=self.zarr) as open_ds:
+
+            if self.create_polygon:
```
alxmrs, line 167: I think the user-defined boolean should override the default behavior.

weather_mv/loader_pipeline/bq.py
```diff
@@ -306,6 +338,49 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+    lower_left = [lon_range[1], lat_range[1]]
+    upper_left = [lon_range[0], lat_range[1]]
+    upper_right = [lon_range[0], lat_range[0]]
+    lower_right = [lon_range[1], lat_range[0]]
+    return [lower_left, upper_left, upper_right, lower_right]
+
+
+def get_lat_lon_range(value, lat_lon, is_point_out_of_bound, lat_grid_resolution, lon_grid_resolution):
+    """Calculate the latitude, longitude point range point latitude, longitude and grid resolution."""
+    if is_point_out_of_bound:
```
alxmrs, line 376: Is this necessary if we are also preprocessing the lat/lngs like we do here? 
https://github.com/google/weather-tools/pull/337/files#diff-938d4c3999aa9d0bd9f832240de92dc8ea660bf71f8b9b6d751b3a16773ce4b1R261

Can we merely do the same thing for latitude?

alxmrs: Like, given that we have that 1-liner, will the data ever be out of bounds? Can you check?

weather_mv/loader_pipeline/bq.py
dabhicusp: Could you please review the discussion at https://github.com/google/weather-tools/pull/337#discussion_r1246421324 and provide your feedback? I will make the necessary changes to the flag based on your input.

weather_mv/loader_pipeline/bq.py
dabhicusp: Hello @alxmrs, I'd like to discuss the two approaches regarding this flag. Here are the options we have:
1. The current implementation sets the default value of [`create_polygon`](https://github.com/google/weather-tools/pull/337/files#diff-938d4c3999aa9d0bd9f832240de92dc8ea660bf71f8b9b6d751b3a16773ce4b1R98) as False. When the user passes the flag, the value becomes True, and a Polygon is created if the input file contains more than a single grid point.
2. An alternative approach is to have the default value of `create_polygon` as True. In this case, a Polygon is created based on the input file by default, and if the user passes the flag, we will not create a polygon.

Currently, I have implemented the first option in the code. I would appreciate your input on whether this approach is suitable, or if we should consider moving towards the second option.

weather_mv/loader_pipeline/bq.py
dabhicusp: I kindly request a meeting to discuss this logic further and provide clarity on this.

alxmrs: I set up a meeting for next week!

weather_mv/loader_pipeline/bq.py
alxmrs: As discussed, let's do option (1). However, to address my original comment: Let's update this comment. "Regular" is an overloaded term: it can both mean "more than one grid point" (in this context) and "on an even latitude longitude grid" (what I've seen "regular" typically means).

weather_mv/loader_pipeline/bq.py
alxmrs: As discussed over VC, let's do option (2).

weather_mv/loader_pipeline/bq.py
dabhicusp: Code updated as per your suggestion. Thanks for the input!

weather_mv/loader_pipeline/bq.py
dabhicusp: Code updated as per your suggestion. Thanks for the input!

weather_mv/loader_pipeline/bq.py
dabhicusp: Example and Logic for the creation of polygon is updated as per your suggestion. Thanks for the input!

weather_mv/loader_pipeline/bq.py
```diff
@@ -230,8 +261,12 @@ def extract_rows(self, uri: str, coordinates: t.List[t.Dict]) -> t.Iterator[t.Di
                 row[DATA_IMPORT_TIME_COLUMN] = self.import_time
                 row[DATA_URI_COLUMN] = uri
                 row[DATA_FIRST_STEP] = first_time_step
-                row[GEO_POINT_COLUMN] = fetch_geo_point(row['latitude'], row['longitude'])
 
+                longitude = ((row['longitude'] + 180) % 360) - 180
+                row[GEO_POINT_COLUMN] = fetch_geo_point(row['latitude'], longitude)
+                row[GEO_POLYGON_COLUMN] = fetch_geo_polygon(row['latitude'], longitude,
+                                                            self.lat_grid_resolution, self.lon_grid_resolution
+                                                            ) if not self.skip_creating_polygon else None
```
alxmrs, line 269: I find this a bit hard to read. Would you mind making it two lines?

weather_mv/loader_pipeline/bq.py
```diff
@@ -306,6 +342,77 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
         raise ValueError(f"Invalid latitude value '{lat}'")
-    long = ((long + 180) % 360) - 180
+    if long > LONGITUDE_RANGE[1] or long < LONGITUDE_RANGE[0]:
+        raise ValueError(f"Invalid longitude value '{long}'")
     point = geojson.dumps(geojson.Point((long, lat)))
     return point
+
+
+def fetch_geo_polygon(latitude: float, longitude: float, lat_grid_resolution: float, lon_grid_resolution: float) -> str:
+    """Create a Polygon based on latitude, longitude and resolution.
```
alxmrs, line 352: Nit: add an extra space after the first line.

weather_mv/loader_pipeline/bq.py
```diff
@@ -306,6 +342,77 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+        (lat_lon_bound[1][0], lat_lon_bound[1][1]),  # upper_left
+        (lat_lon_bound[2][0], lat_lon_bound[2][1]),  # upper_right
+        (lat_lon_bound[3][0], lat_lon_bound[3][1]),  # lower_right
+        (lat_lon_bound[0][0], lat_lon_bound[0][1]),  # lower_left
+    ]))
+    return polygon
+
+
+def bound_point(latitude: float, longitude: float, lat_grid_resolution: float, lon_grid_resolution: float) -> t.List:
+    """Calculate the bound point based on latitude, longitude and grid resolution.
```
alxmrs, line 375: Nit: add an extra space after the first line.

weather_mv/loader_pipeline/bq.py
```diff
@@ -306,6 +342,77 @@ def fetch_geo_point(lat: float, long: float) -> str:
...
+    lower_left = [lon_range[1], lat_range[1]]
+    upper_left = [lon_range[1], lat_range[0]]
+    upper_right = [lon_range[0], lat_range[0]]
+    lower_right = [lon_range[0], lat_range[1]]
+    return [lower_left, upper_left, upper_right, lower_right]
+
+
+def get_lat_lon_range(value: float, lat_lon: str, is_point_out_of_bound: bool,
+                      lat_grid_resolution: float, lon_grid_resolution: float) -> t.List:
+    """Calculate the latitude, longitude point range point latitude, longitude and grid resolution.
```
alxmrs, line 400: Nit: add an extra space after the first line.

alxmrs (approved):
A few nits. Otherwise, this looks ready to merge!

weather_mv/loader_pipeline/bq.py
dabhicusp: Code updated as per your suggestion.

PR #342, opened 2023-06-05
Using `rioxarray` to open `.tif` files

aniketinfocusp:
Using `rioxarray` to open .tif file to get required meta data in the data set which was missing when the .tif file was opened through `xr.open_dataset`.
Also upgrading rioxarray version from `0.12.0 -> 0.13.4` as that's the latest version that supports python 3.8 and has implemented the band_as_variable argument.
Updated CI as well. Turned off `use-only-tar-bz2` flag as rioxarray doesn't support .tar.bz2 artifacts after 0.13.1

fixes https://github.com/google/weather-tools/issues/159

### DIFF
--- a/weather_mv/loader_pipeline/sinks.py
+++ b/weather_mv/loader_pipeline/sinks.py
@@ -147,10 +147,7 @@ def _preprocess_tif(ds: xr.Dataset, filename: str, tif_metadata_for_datetime: st
     """
 
     def _replace_dataarray_names_with_long_names(ds: xr.Dataset):
-        rename_dict = {}
-        for var_name in ds.variables:
-            rename_dict[var_name] = ds[var_name].attrs.get('long_name', var_name)
-        print(rename_dict)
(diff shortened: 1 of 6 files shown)


weather_mv/loader_pipeline/sinks.py
```diff
@@ -145,14 +146,9 @@ def _preprocess_tif(ds: xr.Dataset, filename: str, tif_metadata_for_datetime: st
...
-        if not band_names_dict:
-            band = ds.band_data[i]
-            band.name = ds.band_data.attrs['long_name'][i]
-        else:
-            band = ds.band_data
-            band.name = band_names_dict.get(band.name)
-        return band
+    def _replace_dataarray_names_with_long_names(ds: xr.Dataset):
+        rename_dict = {var_name: ds[var_name].attrs.get('long_name', var_name) for var_name in ds.variables}
+        return ds.rename(rename_dict)
```
alxmrs, line 151: Nice

alxmrs (approved):
LGTM

Issue #311, opened 2023-04-06
When streaming jobs with `weather-mv`, should not require URIs exist before launching the streaming


floraxue:
When the user launches a `weather-mv` job with topics (meaning it's watching for streaming events like new objects being created), it is possible that the provided URI pattern doesn't match any existing objects in the bucket yet. 

However, we currently always [require all URIs exist](https://github.com/google/weather-tools/blob/main/weather_mv/loader_pipeline/pipeline.py#L51) before we launch the beam job [code](https://github.com/google/weather-tools/blob/main/weather_mv/loader_pipeline/pipeline.py#L64-L66).

This will cause immediate failures for the user so they are not able to launch the streaming job at all. 

The check for all URIs exist should only happen for the case when no topics are created. Similarly, for the `first_uri` argument, we only use it for the non-streaming case. With `topics`, we should not create `first_uri` to be clearer.

PR #343, opened 2023-06-08
Streaming in `weather-mv`

aniketinfocusp:
fixes: #311
As mentioned in this pr https://github.com/google/weather-tools/pull/250#issuecomment-1277339173 skipping file prefix validation does not work in case of big query as it was dependent on first_uri to infer schema for big query.

Fixed this issue by creating a pipeline stage that infers the schema and creates the bigquery table using the first uri. This stage is skipped for all subsequent uris.

### DIFF
--- a/weather_mv/loader_pipeline/bq.py
+++ b/weather_mv/loader_pipeline/bq.py
@@ -148,8 +148,10 @@ class ToBigQuery(ToDataSink):
                             region=pipeline_options_dict.get('region'))
             logger.info('Region validation completed successfully.')
 
-    def __post_init__(self):
-        """Initializes Sink by creating a BigQuery table based on user input."""
+    def create_bq_table(self, uri: str) -> str:
+        if self.table:
+            return uri
+        
         if self.zarr:
             self.xarray_open_dataset_kwargs = self.zarr_kwargs
         with open_dataset(self.first_uri, self.xarray_open_dataset_kwargs,
@@ -176,6 +178,7 @@ class ToBigQuery(ToDataSink):
         try:
             table = bigquery.Table(self.output_table, schema=table_schema)
             self.table = bigquery.Client().create_table(table, exists_ok=True)
+            return uri
         except Exception as e:
             logger.error(f'Unable to create table in BigQuery: {e}')
             raise
@@ -243,6 +246,7 @@ class ToBigQuery(ToDataSink):
         """Extract rows of variables from data paths into a BigQuery table."""
         extracted_rows = (
                 paths
+                | 'CreateTable' >> beam.Map(self.create_bq_table)
                 | 'PrepareCoordinates' >> beam.FlatMap(self.prepare_coordinates)
                 | beam.Reshuffle()
                 | 'ExtractRows' >> beam.FlatMapTuple(self.extract_rows)
--- a/weather_mv/loader_pipeline/bq.py
+++ b/weather_mv/loader_pipeline/bq.py
@@ -133,9 +133,9 @@ class ToBigQuery(ToDataSink):
 
         if known_args.output_table:
             # checking if the output table is in format (<project>.<dataset>.<table>).
-            pattern = r'^[\w-]+\.[\w-]+\.[\w-]+$'
-            if not bool(re.match(pattern, known_args.output_table)):
-                raise RuntimeError("Output table is not in correct format.")
+            output_table_pattern = r'^[\w-]+\.[\w-]+\.[\w-]+$'
+            if not bool(re.match(output_table_pattern, known_args.output_table)):
+                raise RuntimeError("output_table is not in correct format.")
 
         if known_args.area:
             assert len(known_args.area) == 4, 'Must specify exactly 4 lat/long values for area: N, W, S, E boundaries.'
@@ -156,16 +156,20 @@ class ToBigQuery(ToDataSink):
             logger.info('Region validation completed successfully.')
 
     def __post_init__(self):
-        self.table =  None
-        project, dataset, table_name = self.output_table.split('.')
+        """Initializes BigQuery table based on user input."""
+        project, dataset_id, table_id = self.output_table.split('.')
         self.project = project
-        self.dataset = dataset
-        self.table_name = table_name
+        self.dataset_id = dataset_id
+        self.table_id = table_id
+        self.table =  None
+
         if self.zarr:
             self.xarray_open_dataset_kwargs = self.zarr_kwargs
 
     def create_bq_table(self, uri: str) -> str:
+        """Create a big query table for the first uri. After table is created, subsequent uris are returned."""
         if self.table:
+            # Skip table creation.
             return uri
 
         with open_dataset(uri, self.xarray_open_dataset_kwargs,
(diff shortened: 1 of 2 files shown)


weather_mv/loader_pipeline/bq.py
```diff
@@ -149,10 +156,23 @@ def validate_arguments(cls, known_args: argparse.Namespace, pipeline_args: t.Lis
             logger.info('Region validation completed successfully.')
 
     def __post_init__(self):
-        """Initializes Sink by creating a BigQuery table based on user input."""
+        """Initializes BigQuery table based on user input."""
+        project, dataset_id, table_id = self.output_table.split('.')
+        self.project = project
+        self.dataset_id = dataset_id
+        self.table_id = table_id
```
alxmrs, line 163:
```suggestion
        self.project, self.dataset_id, self.table_id = self.output_table.split('.')
```

weather_mv/loader_pipeline/bq.py
```diff
@@ -149,10 +156,20 @@ def validate_arguments(cls, known_args: argparse.Namespace, pipeline_args: t.Lis
             logger.info('Region validation completed successfully.')
 
     def __post_init__(self):
-        """Initializes Sink by creating a BigQuery table based on user input."""
+        """Initializes BigQuery table based on user input."""
+        self.project, self.dataset_id, self.table_id = self.output_table.split('.')
+        self.table =  None
```
alxmrs, line 161: extra space.

weather_mv/loader_pipeline/bq.py
```diff
@@ -252,9 +271,9 @@ def expand(self, paths):
             (
                     extracted_rows
                     | 'WriteToBigQuery' >> WriteToBigQuery(
-                        project=self.table.project,
-                        dataset=self.table.dataset_id,
-                        table=self.table.table_id,
+                        project=self.project,
+                        dataset=self.dataset_id,
+                        table=self.table_id,
```
alxmrs, line 276: If we wanted to create the BQ table in the pipeline, a simpler solution would be to change this step's disposition: https://beam.apache.org/documentation/io/built-in/google-bigquery/#create-disposition

weather_mv/loader_pipeline/bq.py
```diff
@@ -149,10 +156,20 @@ def validate_arguments(cls, known_args: argparse.Namespace, pipeline_args: t.Lis
...
-        """Initializes Sink by creating a BigQuery table based on user input."""
+        """Initializes BigQuery table based on user input."""
+        self.project, self.dataset_id, self.table_id = self.output_table.split('.')
+        self.table =  None
+
         if self.zarr:
             self.xarray_open_dataset_kwargs = self.zarr_kwargs
-        with open_dataset(self.first_uri, self.xarray_open_dataset_kwargs,
+
+    def create_bq_table(self, uri: str) -> str:
```
alxmrs, line 166: Can you explain the reasoning for creating the table in the pipeline and not the `__init__`? I like doing so in the constructor so that we get a signal early on. However, I recognize that this slows down the deployment of the pipeline (when we need to copy and read data).

weather_mv/loader_pipeline/bq.py
alxmrs: On second thought, we may want to keep your step since we create an opinionated schema...

weather_mv/loader_pipeline/bq.py
aniketinfocusp: https://github.com/google/weather-tools/blob/ec9a71422f82d34d7e2a7e5497217e87d0b6afa4/weather_mv/loader_pipeline/pipeline.py#L49-L55
because in case of a streaming pipeline, there can be a case that target bucket has no files at all. That will cause `all_uri` to be empty and `first_ur` (which is used to infer the schema would be none too). 
hence putting the table creation stage inside the pipeline as an when a file arrives, it will create the table.

alxmrs: Thanks! That is sound reasoning to me. Ok, I'm on board with this approach and will resolve related issues.

weather_mv/loader_pipeline/bq.py
```diff
@@ -243,6 +261,7 @@ def expand(self, paths):
         """Extract rows of variables from data paths into a BigQuery table."""
         extracted_rows = (
                 paths
+                | 'CreateTable' >> beam.Map(self.create_bq_table)
```
alxmrs, line 264: Concern: I am pretty sure this will try to create a BQ table for every element of `paths`. Remember, `self.tables` won't refer to the state of the class, since global state is not really a think for parallel steps like this.

alxmrs: Consider using something like `sample`: https://beam.apache.org/documentation/transforms/python/aggregation/sample/

weather_mv/loader_pipeline/bq.py
```diff
@@ -130,6 +131,12 @@ def validate_arguments(cls, known_args: argparse.Namespace, pipeline_args: t.Lis
         pipeline_options = PipelineOptions(pipeline_args)
         pipeline_options_dict = pipeline_options.get_all_options()
 
+        if known_args.output_table:
+            # checking if the output table is in format (<project>.<dataset>.<table>).
+            output_table_pattern = r'^[\w-]+\.[\w-]+\.[\w-]+$'
```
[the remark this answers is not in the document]
aniketinfocusp: project_ids can contain `hypens` https://cloud.google.com/resource-manager/docs/creating-managing-projects
\w+ doesn't match when word is like `my-project`.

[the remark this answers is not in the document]
alxmrs: Thanks, that makes sense! Then, would it be `[\w\-]+`? IIRC, `[]` and `-` together are often used to express a range.

weather_mv/loader_pipeline/bq.py
aniketinfocusp: @alxmrs 
i tried putting the create table as a transform in the pipeline but faced many issues
- I tried the `sample` transform as you suggested but there is no way to ensure table creation happens before the usual pipeline flow. (apache_beam python sdk doesn't have support for `Wait.on` which could have made this possible)
- I also tried stateful processing but as we are windowing out pub sub reads, and a state is discarded when the window is expired, the solution didn't work.

I think there is potential in using the `create-disposition` flag in `WriteToBigQuery`. Can you please elaborate on what do you mean by `opinionated schema`

cc: @mahrsee1997

alxmrs: Now that you mention it, I agree that using the create disposition is easiest! We need to make sure that we pass in our computed `schema` (say, from the `init`) into this transform instead of having it automatically make the schema. Thinking it over now, that's what my concern was about: I wasn't sure if we'd get the schema we wanted if it computed it automatically.

mahrsee1997: Was surfing web for some other issue & found this. It might be helping here.

https://beam.apache.org/releases/pydoc/current/apache_beam.io.gcp.bigquery.html#module-apache_beam.io.gcp.bigquery

PR #349, opened 2023-06-23
Added timedelta to json serialzation.

aniketinfocusp:
Current implementation of `to_json_serializable_type()` only handles `np.timedelta` and was missing case of `datetime.timedelta`

### DIFF
--- a/weather_mv/loader_pipeline/util.py
+++ b/weather_mv/loader_pipeline/util.py
@@ -117,6 +117,8 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
 
         # We assume here that naive timestamps are in UTC timezone.
         return value.replace(tzinfo=datetime.timezone.utc).isoformat()
+    elif isinstance(value, datetime.timedelta):
+        return value.total_seconds()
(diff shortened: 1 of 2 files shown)


weather_mv/loader_pipeline/util.py
```diff
@@ -117,6 +117,8 @@ def to_json_serializable_type(value: t.Any) -> t.Any:
 
         # We assume here that naive timestamps are in UTC timezone.
         return value.replace(tzinfo=datetime.timezone.utc).isoformat()
+    elif isinstance(value, datetime.timedelta):
```
mahrsee1997, line 120: Please add test case for the same.

mahrsee1997 (approved):
LGTM ! Please add the test case & we can merge this.

weather_mv/loader_pipeline/util.py
aniketinfocusp: added in [18d632c](https://github.com/google/weather-tools/pull/349/commits/18d632c2af4aa587886c12312af1cdf7dc0e57db)

Issue #350, opened 2023-06-24
Find a way to exclude test data when building docker image.


alxmrs:

alxmrs:
Since we git clone weather tools in the docker image, we should also find a way to prevent test data being checked in.

https://github.com/google/weather-tools/blob/main/Dockerfile#L29

I think this could make our image more lightweight!

alxmrs:
Here's an idea! Let's replace hosting our own test data by using examples from the Xarray tutorial: https://docs.xarray.dev/en/stable/generated/xarray.tutorial.open_dataset.html

PR #351, opened 2023-06-24
Removing test data after clone in Docker image.

alxmrs:
I believe that our docker images are larger than they should be since they also include checked-in test data. This change adds a short-term fix to delete all `test_data` folders in each weather_tool before building the rest of the image.

In small experiments, this can save ~190 MiBs of disk space.

This is a band aid for #350.


mahrsee1997 (approved):
LGTM.

PR #352, opened 2023-06-26
Small fix to return correct hdate for leap days

raspstephan:
ECMWF API expects day==28 for all hdates if input date is Feb 29, even if target hdate is also a leap year. Added small correction to fix this.


mahrsee1997 (approved):
LGTM. The modifications appear to be in accordance with the stated requirements.

Comment: Please resolve lint error.

alxmrs (approved):
LGTM, thanks for the patch!

PR #354, opened 2023-06-28
Refactored coordinate serialization

aniketinfocusp:
As we are performing json serialization on coordinates in `get_coordinates()` here https://github.com/google/weather-tools/blob/edd742d591cb428aa41f6c4d713c969e39f8dbbe/weather_mv/loader_pipeline/util.py#L214

Indexing by dimension coordinate labels fails for coordinates that are serialized to types not index-able by xarray (eg `datetime.timedelta` is serialized to floats; so when timedelta is a coordinate, `.loc[]` fails as it gives type mis match error)
https://github.com/google/weather-tools/blob/edd742d591cb428aa41f6c4d713c969e39f8dbbe/weather_mv/loader_pipeline/bq.py#L213-L215
  
This is fixed by performing json serialization _after_ the indexing is done.

### DIFF
--- a/weather_mv/loader_pipeline/bq.py
+++ b/weather_mv/loader_pipeline/bq.py
@@ -219,7 +219,7 @@ class ToBigQuery(ToDataSink):
                 row = {n: to_json_serializable_type(ensure_us_time_resolution(v.values))
                        for n, v in row_ds.data_vars.items()}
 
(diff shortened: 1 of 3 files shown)


alxmrs (approved):
LGTM, thank you!

Issue #353, opened 2023-06-27
All tools should make use of public runtime container image to manage dependencies


alxmrs:
We've finally been able to host a public Docker image for weather-tools. The first version can be pulled with this command: 

```
docker pull gcr.io/weather-tools-prod/weather-tools:0.0.0
```

This unblocks our ongoing efforts to make weather-tools easy to install! 

In brief, here is what's needed to make installation better: 
- For each tool's associated script, let's add an argument that points to our runtime container. Right now, we specify extra_packages. I believe this can go away. See https://github.com/google/weather-tools/blob/edd742d591cb428aa41f6c4d713c969e39f8dbbe/weather_dl/weather-dl#L76. I believe we can remove a lot of the packaging and installing logic in these scripts.
- We will most likely need to create a new runtime container with an incremented version to get the latest dependencies. 

The next steps to fully land a clean install experience will involve: 
- Landing the `conda install google-weather-tools` experience via this staged recipe: https://github.com/conda-forge/staged-recipes/pull/20892
- Following our release process and tagging a new public release.
- Setting up CI to keep our conda environment and docker image versions consistent via Google Cloud Build and GCR.

CC: @blackvvine @DarshanSP19

PR #355, opened 2023-07-05
weather-* scripts default to runtime container image.

DarshanSP19:
- Removed `--extra_package` and relevant build logic from all three tools.
- Injecting the `args` related to the runtime container (`--sdk_container_image` and `--experiments`) if not provided by the user.
- `--sdk_container_image` will be the public docker image like below.
- New flag for suppling the local code to DataflowRunner `-–use-local-code`.
```
gcr.io/weather-tools-prod/weather-tools:0.0.0
```

- Resolves the first part of #353


weather_dl/download_pipeline/pipeline.py
```diff
@@ -46,6 +46,7 @@
 from .stores import TempFileStore, LocalFileStore
 
 logger = logging.getLogger(__name__)
+SDK_CONTAINER_IMAGE='gcr.io/weather-tools-prod/weather-tools:0.0.0'
```
alxmrs, line 49: Let's make this an environment variable with this GCR image as the default.

weather_dl/download_pipeline/pipeline.py
```diff
@@ -180,6 +181,10 @@ def run(argv: t.List[str], save_main_session: bool = True) -> PipelineArgs:
 
     known_args, pipeline_args = parser.parse_known_args(argv[1:])
 
+    if "DataflowRunner" in pipeline_args and  "--sdk_container_image" not in pipeline_args:
```
alxmrs, line 184: I think this should go in the `weather-dl` script.

weather_dl/download_pipeline/pipeline.py
alxmrs: Ah, I see that you're doing this below. Great!

weather_dl/weather-dl
```diff
@@ -49,28 +45,4 @@ if __name__ == '__main__':
...
-
-            with tarfile.open(pkg_archive, 'r') as tar:
-                assert any([f.endswith('.py') for f in tar.getnames()]), 'extra_package must include python files!'
-
-            # cleanup memory to prevent pickling error.
-            tar = None
-            weather_dl = None
-
-            cli(['--extra_package', pkg_archive])
+    cli()
```
alxmrs, line 48: As mentioned above, I think the SDK container argument should be input here as a few list arguments in the CLI.

alxmrs (commented):
Left comments for how weather-dl should be structured; the same applies to the other tools. 

Thanks so much for this! Really excited to see this feature land.

weather_dl/download_pipeline/pipeline.py
DarshanSP19: Updated.

weather_dl/weather-dl
DarshanSP19: Updated.

weather_dl/download_pipeline/pipeline.py
```diff
@@ -177,6 +177,7 @@ def run(argv: t.List[str], save_main_session: bool = True) -> PipelineArgs:
                         help="Update the manifest for the already downloaded shards and exit. Default: 'false'.")
     parser.add_argument('--log-level', type=int, default=2,
                       help='An integer to configure log level. Default: 2(INFO)')
+    parser.add_argument('--use-local-code', action='store_true', default=False, help='Supply local code to the Runner.')
```
alxmrs, line 180: Please update the README docs with this flag. The way I do this is by running copy & pasting from the help output (`--help`).

alxmrs (approved):
One nit, but then this is good to go!

weather_dl/download_pipeline/pipeline.py
DarshanSP19: Updated.

PR #357, opened 2023-07-11
`weather-mv` will ingest data into BQ from Zarr much faster.

alxmrs:
`weather-mv bq`'s previous Zarr ingestion system only used one worker. This PR uses Xarray-Beam for Zarr ingestion, in order to distributed `xr.Dataset` chunks across beam workers. This improves ingestion into BQ. 

_Outstanding issues_: I can't find a way to incrementally load rows into BQ from Zarr. While I've used windowing on fixed intervals to break up a large ingestion job into smaller parts, it seems like the actual writing to BQ gets stuck in a reshuffle step within the `WriteToBigQuery` transform. In this PR or a future PR, let's try to find a way to incrementally write rows to BQ once they've been processed, instead of having to wait for the entire dataset to be processed. CC: @dabhicusp.

### DIFF
--- a/weather_mv/loader_pipeline/bq.py
+++ b/weather_mv/loader_pipeline/bq.py
@@ -240,7 +240,7 @@ class ToBigQuery(ToDataSink):
             data_ds: xr.Dataset = _only_target_vars(ds, self.variables)
             yield from self.to_rows(coordinates, data_ds, uri)
 
-    def to_rows(self, coordinates, ds, uri) -> t.Iterator[t.Dict]:
+    def to_rows(self, coordinates: t.List[t.Dict], ds: xr.Dataset, uri: str) -> t.Iterator[t.Dict]:
         first_ts_raw = (
             ds.time[0].values if isinstance(ds.time.values, np.ndarray)
             else ds.time.values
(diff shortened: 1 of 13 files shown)


mahrsee1997 (commented):
Overall LGTM.

@dabhicusp/@DarshanSP19, could you please provide the benchmarking results for the current main branch as compared to this specific branch?

weather_mv/loader_pipeline/bq.py
```diff
@@ -236,73 +238,90 @@ def extract_rows(self, uri: str, coordinates: t.List[t.Dict]) -> t.Iterator[t.Di
...
-            (
-                    extracted_rows
-                    | 'Log Extracted Rows' >> beam.Map(logger.debug)
+            ds, chunks = xbeam.open_zarr(self.first_uri, **self.xarray_open_dataset_kwargs)
+            ds.attrs[DATA_URI_COLUMN] = self.first_uri
+            extracted_rows = (
+                paths
+                | 'OpenChunks' >> xbeam.DatasetToChunks(ds, chunks)
+                | 'ExtractRows' >> beam.FlatMapTuple(self.chunks_to_rows)
+                | 'Window' >> beam.WindowInto(window.FixedWindows(60))
```
mahrsee1997, line 309: If incorporating these steps (window) enables real-time data ingestion into BigQuery (in batch jobs), then we should relocate these lines [here](https://github.com/google/weather-tools/pull/357/files#diff-938d4c3999aa9d0bd9f832240de92dc8ea660bf71f8b9b6d751b3a16773ce4b1R316).

dabhicusp (commented):
LGTM @alxmrs.

DarshanSP19:
Data after running pipelines for mentioned cases.

| File Size | Before | After |
|--------|--------|--------|
| ~100Mb |58 min. |41min. |
| ~1Gb | 16h 30min. | 1h 12min. | 

This changes are relatively fast than before for zarr batch ingestion.

fredzyda (approved):

PR #376, opened 2023-08-01
Update sparse data ingestion to handle huge datasets

deepgabani8:


weather_mv/loader_pipeline/ee.py
```diff
@@ -486,21 +487,46 @@ def convert_to_asset(self, queue: Queue, uri: str):
                 channel_names = []
                 file_name = f'{asset_name}.csv'
 
-                df = xr.Dataset.to_dataframe(ds)
-                df = df.reset_index()
-                # NULL and NaN create data-type mismatch issue in ee therefore replacing all of them.
-                # fillna fills in NaNs, NULLs, and NaTs but we have to exclude NaTs.
-                non_nat = df.select_dtypes(exclude=['datetime', 'timedelta', 'datetimetz'])
-                df[non_nat.columns] = non_nat.fillna(-9999)
+                shape = math.prod(list(ds.dims.values()))
```
alxmrs, line 490: Check if Xarray has a built-in method for this. I think there is one on DataArrays.

weather_mv/loader_pipeline/ee.py
```diff
@@ -486,21 +487,46 @@ def convert_to_asset(self, queue: Queue, uri: str):
...
+                # Names of dimesions, coordinates and data variables.
+                dims = list(ds.dims)
+                coords = [c for c in list(ds.coords) if c not in dims]
+                vars = list(ds.data_vars)
+                header = dims + coords + vars
```
alxmrs, line 495: There could be duplicates, esp wrt `dims` and `coords`. How is this handled? Should this be a set?

weather_mv/loader_pipeline/ee.py
```diff
@@ -486,21 +487,46 @@ def convert_to_asset(self, queue: Queue, uri: str):
...
+                                        ",".join(map(str, i))
+                                        for i in zip(*[[*get_dims_data(i), *d[i:i + ROWS_PER_WRITE]] for d in data])
+                                    ]
+                                ).replace("NULL", "-9999")
+                                + "\n"
+                            )
+
                     with FileSystems().create(target_path) as dst:
-                        shutil.copyfileobj(tmp_df, dst, WRITE_CHUNK_SIZE)
+                        shutil.copyfileobj(temp, dst, WRITE_CHUNK_SIZE)
```
alxmrs, line 529: Would it be worthwhile to use a faster copy method?

weather_mv/loader_pipeline/ee.py
```diff
@@ -486,21 +487,46 @@ def convert_to_asset(self, queue: Queue, uri: str):
...
+                        for i in range(0, shape, ROWS_PER_WRITE):
+                            f.write(
+                                "\n".join(
+                                    [
+                                        ",".join(map(str, i))
+                                        for i in zip(*[[*get_dims_data(i), *d[i:i + ROWS_PER_WRITE]] for d in data])
+                                    ]
+                                ).replace("NULL", "-9999")
+                                + "\n"
+                            )
```
alxmrs, line 526: Let's use Python's native CSV writing library, I believe it will be less error prone. 

https://docs.python.org/3/library/csv.html

weather_mv/loader_pipeline/ee.py
deepgabani8: We can use `ds.prod` or `DataArray.prod` if we convert Frozen to ds or da. I think direct multiplication is simple instead.

weather_mv/loader_pipeline/ee.py
deepgabani8: Sounds good 👍

weather_mv/loader_pipeline/ee.py
deepgabani8: Yes, we are taking coords which are not in dims.

weather_mv/loader_pipeline/ee.py
alxmrs: Ok, SGTM.

alxmrs (approved):
LGTM!

PR #377, opened 2023-08-07
Weather-sp updated.

dabhicusp:
While splitting the `Grib` file using `weather-sp`, I noticed that the file names generated after the splitting did not match the expected format as shown [here](https://github.com/google/weather-tools/tree/main/weather_sp#output-template-with-python-style-formatting). Instead of providing the actual values for {0}, {1}, and so on, it was displaying `undef`. To resolve this issue, I made some modifications to the code, and now it is working correctly. I have thoroughly tested these changes both locally, and everything is functioning as expected.


alxmrs (approved):
LGTM

PR #396, opened 2023-09-18
Added feature in weather-mv to extract specific date's data from zarr files.

dabhicusp:
I have added a feature in `weather-mv` which can allows users to access data for a specific date from Zarr files. In the existing codebase user have no choice for working with some chunks of the `zarr` file so user have to work with whole zarr file every time. 
One use-case: If user want to ingest specific date's data from the zarr file to BQ then he/she can do this using of this code.

### DIFF
--- a/weather_mv/loader_pipeline/sinks.py
+++ b/weather_mv/loader_pipeline/sinks.py
@@ -183,12 +183,12 @@ def _preprocess_tif(ds: xr.Dataset, filename: str, tif_metadata_for_start_time: 
     try:
         # if start_time/end_time is in integer milliseconds
         init_time = (int(start_time.timestamp()) if start_time is not None
-                        else int(ds.attrs[tif_metadata_for_start_time]) / 1000.0)
+                     else int(ds.attrs[tif_metadata_for_start_time]) / 1000.0)
         coords['time'] = datetime.datetime.utcfromtimestamp(init_time)
 
         if tif_metadata_for_end_time:
             forecast_time = (int(end_time.timestamp()) if end_time is not None
-                            else int(ds.attrs[tif_metadata_for_end_time]) / 1000.0)
+                             else int(ds.attrs[tif_metadata_for_end_time]) / 1000.0)
             coords['valid_time'] = datetime.datetime.utcfromtimestamp(forecast_time)
 
         ds = ds.assign_coords(coords)
@@ -198,12 +198,14 @@ def _preprocess_tif(ds: xr.Dataset, filename: str, tif_metadata_for_start_time: 
         try:
             # if start_time/end_time is in UTC string format
             init_time = (int(start_time.timestamp()) if start_time is not None
-                        else datetime.datetime.strptime(ds.attrs[tif_metadata_for_start_time], '%Y-%m-%dT%H:%M:%SZ'))
+                         else datetime.datetime.strptime(ds.attrs[tif_metadata_for_start_time],
+                                                         '%Y-%m-%dT%H:%M:%SZ'))
             coords['time'] = init_time
 
             if tif_metadata_for_end_time:
                 forecast_time = (int(end_time.timestamp()) if end_time is not None
-                        else datetime.datetime.strptime(ds.attrs[tif_metadata_for_end_time], '%Y-%m-%dT%H:%M:%SZ'))
+                                 else datetime.datetime.strptime(ds.attrs[tif_metadata_for_end_time],
+                                                                 '%Y-%m-%dT%H:%M:%SZ'))
                 coords['valid_time'] = forecast_time
 
             ds = ds.assign_coords(coords)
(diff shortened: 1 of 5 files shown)


weather_mv/loader_pipeline/sinks.py
```diff
@@ -406,7 +409,19 @@ def open_dataset(uri: str,
     """Open the dataset at 'uri' and return a xarray.Dataset."""
     try:
         if is_zarr:
+            start_date = end_date = None
+            if 'start_date' in open_dataset_kwargs:
+                start_date = open_dataset_kwargs['start_date']
+                if open_dataset_kwargs is not None:
+                    del open_dataset_kwargs['start_date']
```
alxmrs, line 416: Can you call something like `pop()` here?

weather_mv/loader_pipeline/sinks.py
```diff
@@ -406,7 +409,19 @@ def open_dataset(uri: str,
...
+            if 'end_date' in open_dataset_kwargs:
+                end_date = open_dataset_kwargs['end_date']
+                if open_dataset_kwargs is not None:
+                    del open_dataset_kwargs['end_date']
             ds: xr.Dataset = _add_is_normalized_attr(xr.open_dataset(uri, engine='zarr', **open_dataset_kwargs), False)
+            if start_date is not None and end_date is not None:
+                timedelta = pd.date_range(start_date, end_date, periods=25, inclusive='left')
+                ds = ds.sel(time=timedelta)
```
alxmrs, line 424: TIL! this is slick!

weather_mv/loader_pipeline/sinks.py
```diff
@@ -406,7 +409,19 @@ def open_dataset(uri: str,
...
+                start_date = open_dataset_kwargs['start_date']
+                if open_dataset_kwargs is not None:
+                    del open_dataset_kwargs['start_date']
+            if 'end_date' in open_dataset_kwargs:
+                end_date = open_dataset_kwargs['end_date']
+                if open_dataset_kwargs is not None:
+                    del open_dataset_kwargs['end_date']
             ds: xr.Dataset = _add_is_normalized_attr(xr.open_dataset(uri, engine='zarr', **open_dataset_kwargs), False)
+            if start_date is not None and end_date is not None:
+                timedelta = pd.date_range(start_date, end_date, periods=25, inclusive='left')
```
alxmrs, line 423: nit: let's call this `time_range` or similar.

weather_mv/loader_pipeline/sinks.py
dabhicusp: It's my pleasure. 😄

alxmrs: FWIW: I think this also may be overkill. IIUC, you could pas the `start_date` and `end_date` from the `pd.date_range` directly into the `sel`, e.g.: 

```
ds = ds.sel(time=slice(start_date, end_date))
```

weather_mv/loader_pipeline/sinks.py
dabhicusp: Thank you for suggestions.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -406,7 +409,19 @@ def open_dataset(uri: str,
     """Open the dataset at 'uri' and return a xarray.Dataset."""
     try:
         if is_zarr:
+            start_date = end_date = None
+            if 'start_date' in open_dataset_kwargs:
+                start_date = open_dataset_kwargs['start_date']
+                if open_dataset_kwargs is not None:
+                    open_dataset_kwargs.pop('start_date', None)
```
alxmrs, line 416:
```suggestion
                start_date = open_dataset_kwargs.pop('start_date', None)                    
```

weather_mv/loader_pipeline/sinks.py
```diff
@@ -406,7 +409,19 @@ def open_dataset(uri: str,
...
+            if 'end_date' in open_dataset_kwargs:
+                end_date = open_dataset_kwargs['end_date']
+                if open_dataset_kwargs is not None:
+                    open_dataset_kwargs.pop('end_date', None)
             ds: xr.Dataset = _add_is_normalized_attr(xr.open_dataset(uri, engine='zarr', **open_dataset_kwargs), False)
+            if start_date is not None and end_date is not None:
+                time_range = pd.date_range(start_date, end_date, periods=25, inclusive='left')
+                ds = ds.sel(time=time_range)
```
alxmrs, line 424: Did you see my other comment? why not just pass `start_date` and `end_date` into the `sel` via slice? 
```suggestion
                ds = ds.sel(time=slice(start_date, end_date))
```

weather_mv/loader_pipeline/sinks.py
dabhicusp: Hi @alxmrs, it's important to ensure that a dictionary is not **empty** before using the `pop` operation directly on it. Therefore, we have to must add a condition like `if open_dataset_kwargs is not None:` before using `pop` operation.

Despite making code updates, I aimed to keep the code as minimal as possible.

weather_mv/loader_pipeline/sinks.py
dabhicusp: Sorry I forgot the above comments. now it's changed.

mahrsee1997 (approved):
LGTM.

alxmrs (approved):
Perfect, change looks great!

PR #399, opened 2023-10-02
Not re-initializing ee if it's already initialized in the same worker.

deepgabani8:
Earlier it was (re)initializing earthengine client upon every ingestion call, which in turn was calling ee highvolume api.

In beam dataflow, we essentially need to initialize earthengine only once in a device, not even a dataflow worker. These changes initializes ee only if necessary. Now it makes 50% less EE calls in filter and ingestion step.

Effect: Increases ingestion speed by roughly **100%**.


alxmrs (approved):
LGTM

Issue #414, opened 2023-10-30
Faster ingestion into BQ by converting the chunk into pd.Dataframe


DarshanSP19:
In `weather-mv` we're dividing the Dataset into small chunks that's adding appropriate parallelism in the pipeline. In the next step if we convert those small chunks into pandas Dataframes it would reduce the cost of generating the flat rows as extracting the rows from Dataframe is very fast.
```
df = ds.to_dataframe().reset_index()
```
Here `ds` is a small chunk of a dataset and `reset_index()` will flatten the dataset chunk into a normalized dataframe.

PR #415, opened 2023-10-31
Faster ingestion of Zarr into BQ by converting the chunk into pd.Dataframe

DarshanSP19:
We're having small chunks of a dataset while processing zarr in `weather-mv`. These changes will convert those chunks into `dataframe` and then extract rows directly from `dataframes`. As the chunk size in our control we can control the memory consumption during the pipeline.

Considerable Points
- Works with the `zarr` dataset for now.
- Will update for all types of dataset in future.
- Users can pass cli arguments to open datasets in the specified chunk scheme.
Example
```
--input_chunks '{ "time": 1, "level": 1 }'
```
Partially Solved: #414


DarshanSP19:
The optimization changes are done in #473 .

PR #418, opened 2023-11-22
`dl-v2` semi-minor improvements

aniketinfocusp:
Fixes:
- gcs client was timing out as a global object was used. Fixed it by returning a new gcs client every time.
- added some more conditions for _get_non_successfull_downloads for refetching a download.
- added route to redeploy licenses
- simplified image build step for images.


weather_dl_v2/fastapi-server/routers/license.py
```diff
@@ -198,5 +198,39 @@ async def delete_license(
...
+
+    if client_name is not None:
+        licenses = await license_handler._get_license_by_client_name(client_name)
+
+    if len(licenses) == 0:
+        return {"message": "No license found."}
+
+    for license in licenses:
+        license_id = license['license_id']
+        logger.info(f"Terminating deployment {license['k8s_deployment_id']}")
```
mahrsee1997, line 225: Mention `license_id` in this log.

mahrsee1997 (approved):
LGTM.

PR #425, opened 2023-12-20
`dl-v2` add only_failed flag to download refetch

aniketinfocusp:


mahrsee1997:
This looks good ! What about CLI changes for the same ?

aniketinfocusp:
I can add those changes in this PR
But as these server changes would not go live till we restart server, i think it better to implement later in cli

mahrsee1997:
Let's do it in this PR only. When implementing a new feature make sure it's complete e2e.

PR #259, opened 2022-11-19
New flags for tiff file ingestions to Earth engine

j9sh264:
This PR contains changes for the weather-mover tool. Some new flags are introduced to aid in .tiff file ingestion to the Earth engine.

PR #426, opened 2023-12-20
Added logic to allow for custom band naming

j9sh264:


j9sh264:
Had previously raised a [PR](https://github.com/google/weather-tools/pull/259) to allow for custom band naming. But those changes seem to have been dropped in some iteration.

weather_mv/loader_pipeline/ee.py
```diff
@@ -459,15 +458,16 @@ def convert_to_asset(self, queue: Queue, uri: str):
                 attrs = ds.attrs
                 data = list(ds.values())
                 asset_name = get_ee_safe_name(uri)
-                channel_names = [da.name for da in data]
-                start_time, end_time, is_normalized = (attrs.get(key) for key in
-                                                       ('start_time', 'end_time', 'is_normalized'))
+                channel_names = list(self.band_names_dict.values()) \
```
deepgabani8, line 461: `band_names_dict`'s values would not always be in order.
i.e. data has 2 data_vars ['precip', 'u10']. And if the user gives band_names_dict = '{"u10": "new_u10", "precip": "new_precip"}'.

weather_mv/loader_pipeline/ee.py
```diff
@@ -459,15 +458,16 @@ def convert_to_asset(self, queue: Queue, uri: str):
...
+                    if self.band_names_dict \
+                    else [da.name for da in data]
+
                 dtype, crs, transform = (attrs.pop(key) for key in ['dtype', 'crs', 'transform'])
-                attrs.update({'is_normalized': str(is_normalized)})  # EE properties does not support bool.
```
deepgabani8, line 466: Is this removed?

j9sh264: I removed it and there was no error produced while ingesting the asset to the Earth Engine. Should we add functionality for converting `bool` values to **string** in the `to_json_serializable_type` function?

weather_mv/loader_pipeline/ee.py
j9sh264: Could we instead ask the user to input the `band_names` flag as a list of the band names (in order) instead of a dictionary? Because anyhow the mapping is something like this -
```
{
  'b0': 'band_name_1',
  'b1': 'band_name_2'
}
```

weather_mv/loader_pipeline/ee.py
deepgabani8: No need. This is fine. I was just making sure that we don't need this attribute.

j9sh264: Marking as resolved

weather_mv/loader_pipeline/ee.py
j9sh264: Resolved

deepgabani8 (approved):

PR #427, opened 2023-12-27
xql - Initial Commit

DarshanSP19:
Added tool to Query Xarray Dataset using SQL.

# Supported Features

* **Select Variables** - From a large dataset having hundreds of variables select only needed variables.
* **Apply Where Clouse** - A general where condition like SQL. Applicable for queries which includes data for specific time range or only for specific regions. (Conditions on coordinates suppoted for now.)
* **Group By And Aggregate Functions** - Aggregate functions `AVG()`, `MIN()`, `MAX()` suppoted after applying groupby on any coordinate like time.

Checkout [README.md](https://github.com/google/weather-tools/blob/xql-init/xql/README.md) for more info.

### DIFF
--- a/xql/README.md
+++ b/xql/README.md
@@ -1,0 +1,41 @@
+# `xql` - Querying Xarray Datasets with SQL
+
+Running SQL like quieries on Xarray Datasets. (Only works on zarr datasets for now.)
+
+# Supported Features
+
+* **Select Variables** - From a large dataset having hundreds of variables select only needed variables.
+* **Apply Where Clouse** - A general where condition like SQL. Applicable for queries which includes data for specific time range or only for specific regions. (Conditions on coordinates suppoted for now.)
+* **Group By And Aggregate Functions** - Aggregate functions `AVG()`, `MIN()`, `MAX()` suppoted after applying groupby on any coordinate like time.
+
+# Usage
+
+Install required packages
+```
+pip install -r xql/requirements.txt
+```
+
+Jump into xql
+```
+python xql/main.py
+```
+---
+
+Running a simple query on dataset. Comparing with SQL a data variable is like a column and table is like a dataset.
+```
+SELECT evaporation, geopotential_at_surface, temperature FROM '{TABLE}'
+```
+Replace `{TABLE}` with dataset uri. Ex. `gs://gcp-public-data-arco-era5/ar/full_37-1h-0p25deg-chunk-1.zarr-v3`:
+
+---
+Apply a conditions. Query to get temperature of arctic region in last winter:
+```
+SELECT temperature FROM '{TABLE}' WHERE time >= '2022-12-01' AND time < '2023-03-01' AND latitude >= 66.5
+```
+---
+Aggregating results using Group By and Aggregate function. Daily average of temperature of last winter in arctic region.
+```
+SELECT AVG(temperature) FROM '{TABLE}' WHERE time >= '2022-12-01' AND time < '2023-03-01' AND latitude >= 66.5
+GROUP BY time_day
+```
+Replace `time_day` to `time_month` or `time_year` if monthly or yearly average is needed. Also use MIN() and MAX() functions same way as AVG().
--- a/xql/README.md
+++ b/xql/README.md
@@ -8,7 +8,8 @@ Running SQL like queries on Xarray Datasets.
 * **`Select` Variables** - From a large dataset having hundreds of variables select only needed variables.
 * **Apply `where` clause** - A general where condition like SQL. Applicable for queries which includes data for specific time range or only for specific regions. 
 * > Note: For now, we support conditions on coordinates.
-* **`group by` and `aggregate` Functions** - Aggregate functions `AVG()`, `MIN()`, `MAX()` supported after applying `group-by` on any coordinate like time.
+* **`group by` and `order by` Functions** - Both are supported on the coordinates  only. e.g. time, latitude, longitude, etc.
+* **`aggregate` Functions** - Aggregate functions `AVG()`, `MIN()`, `MAX()`, etc. are supported on any coordinate like time.
 
 # Quickstart
 
@@ -49,3 +50,9 @@ SELECT AVG(temperature) FROM '{TABLE}' WHERE time >= '2022-12-01' AND time < '20
 GROUP BY time_day
 ```
 Replace `time_day` to `time_month` or `time_year` if monthly or yearly average is needed. Also use MIN() and MAX() functions same way as AVG().
+
+---
+Order by latitude, longitude in ascending and descending order.
+```
+SELECT surface_pressure FROM 'gs://gcp-public-data-arco-era5/ar/full_37-1h-0p25deg-chunk-1.zarr-v3' WHERE time >= '2021-06-01T00:00:00Z' AND time <= '2021-06-30T23:59:59Z' ORDER BY latitude, longitude DESC LIMIT 1
+```
--- a/xql/README.md
+++ b/xql/README.md
@@ -1,58 +1,116 @@
 # `xql` - Querying Xarray Datasets with SQL
 
-Running SQL like queries on Xarray Datasets.
+Running SQL like queries on Xarray Datasets. Consider dataset as a table and data variable as a column.
 > Note: For now, we support only zarr datasets.
 
 # Supported Features
 
 * **`Select` Variables** - From a large dataset having hundreds of variables select only needed variables.
 * **Apply `where` clause** - A general where condition like SQL. Applicable for queries which includes data for specific time range or only for specific regions. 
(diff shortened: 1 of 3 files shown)


xql/README.md
```diff
@@ -0,0 +1,41 @@
+# `xql` - Querying Xarray Datasets with SQL
+
+Running SQL like quieries on Xarray Datasets. (Only works on zarr datasets for now.)
```
mahrsee1997, line 3:
```suggestion
Running SQL like queries on Xarray Datasets.
> Note: For now, we support only zarr datasets.
```

xql/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+# Supported Features
+
+* **Select Variables** - From a large dataset having hundreds of variables select only needed variables.
+* **Apply Where Clouse** - A general where condition like SQL. Applicable for queries which includes data for specific time range or only for specific regions. (Conditions on coordinates suppoted for now.)
```
mahrsee1997, line 8:
```suggestion
* **Apply `where` clause** - A general where condition like SQL. Applicable for queries which includes data for specific time range or only for specific regions. 
> Note: For now, we support conditions on coordinates.
```

xql/README.md
```diff
@@ -0,0 +1,41 @@
+# `xql` - Querying Xarray Datasets with SQL
+
+Running SQL like quieries on Xarray Datasets. (Only works on zarr datasets for now.)
+
+# Supported Features
+
+* **Select Variables** - From a large dataset having hundreds of variables select only needed variables.
```
mahrsee1997, line 7:
```suggestion
* **`select` Variables** - From a large dataset having hundreds of variables, select only needed variables.
```

xql/README.md
```diff
@@ -0,0 +1,41 @@
...
+* **Apply Where Clouse** - A general where condition like SQL. Applicable for queries which includes data for specific time range or only for specific regions. (Conditions on coordinates suppoted for now.)
+* **Group By And Aggregate Functions** - Aggregate functions `AVG()`, `MIN()`, `MAX()` suppoted after applying groupby on any coordinate like time.
```
mahrsee1997, line 9:
```suggestion
* **`group by` and `aggregate` Functions** - Aggregate functions `AVG()`, `MIN()`, `MAX()` supported after applying `group-by` on any coordinate like time.
```

xql/README.md
```diff
@@ -0,0 +1,41 @@
...
+```
+python xql/main.py
+```
+---
+
+Running a simple query on dataset. Comparing with SQL a data variable is like a column and table is like a dataset.
+```
+SELECT evaporation, geopotential_at_surface, temperature FROM '{TABLE}'
+```
+Replace `{TABLE}` with dataset uri. Ex. `gs://gcp-public-data-arco-era5/ar/full_37-1h-0p25deg-chunk-1.zarr-v3`:
```
mahrsee1997, line 28:
```suggestion
Replace `{TABLE}` with dataset uri. Eg. `gs://gcp-public-data-arco-era5/ar/full_37-1h-0p25deg-chunk-1.zarr-v3`.
```

xql/main.py
```diff
@@ -0,0 +1,135 @@
+
```
mahrsee1997, line 1: Add license information at the top of all files.

xql/main.py
```diff
@@ -0,0 +1,135 @@
...
+    while True:
+
+        query = input("xql>")
+
+        if query == "exit":
+            break
+
+        result = parse_query(query)
+
+        if isinstance(result, xr.Dataset):
```
mahrsee1997, line 132: Just need to understand; why this is required ?

mahrsee1997 (changes requested):
First pass...

xql/main.py
DarshanSP19: If the query executes successfully it'll return xr.Dataset. This is just to handle error.

xql/main.py
DarshanSP19: Added.

xql/main.py
```diff
@@ -0,0 +1,149 @@
...
+    "eq" : lambda a, b: a == b,
+    "gt" : lambda a, b: a > b,
+    "lt" : lambda a, b: a < b,
+    "gte" : lambda a, b: a >= b,
+    "lte" : lambda a, b: a <= b,
+}
+
+aggregate_function_map = {
+    'avg': lambda x, y: x.mean(dim=y) if y else x.mean(),
+    'min': lambda x, y: x.min(dim=y) if y else x.mean(),
```
aniketinfocusp, line 34: I think you meant `x.min()`.
Description of what's happening here is helpful. It's better to create a Factory class of this instead of creating a map, that way you can apply better validations to these and future functions.

xql/main.py
```diff
@@ -0,0 +1,149 @@
...
+    'max': lambda x, y: x.max(dim=y) if y else x.mean(),
```
aniketinfocusp, line 35: `x.max()`

xql/main.py
```diff
@@ -0,0 +1,149 @@
...
+
+def parse(a, b):
+    arr_type = a.dtype.name
+    if arr_type == 'float64':
+        b = np.float64(b)
+    elif arr_type == 'float32':
+        b = np.float32(b)
+    elif arr_type == 'datetime64[ns]':
+        b = np.datetime64(b)
+    return a, b
```
aniketinfocusp, line 46: this parsing always assumes the LHS is the variable and RHS is the value (eg `temp > 5`)
general SQL query supports both `temp > 5` and `5 < temp`.
Additional checks before parsing  are required.

Or if XQL is not supporting `5 < temp` kind of conditions, should be mentioned in Docs.

xql/main.py
```diff
@@ -0,0 +1,149 @@
...
+
+def apply_group_by(fields, ds: xr.Dataset, agg_funcs):
+    grouped_ds = ds
+    for field in fields:
+        field_parts = field.split("_")
+        groupby_field = field_parts[0]
+        if len(field_parts) > 1:
+            groupby_field = f"{groupby_field}.{field_parts[1]}"
+            groups = grouped_ds.groupby(groupby_field)
+            grouped_ds = apply_aggregation(groups, list(agg_funcs.values())[0])
```
aniketinfocusp, line 85: This logic is for grouping by time.month or time.day fields
But this would break things if someone tries to group by any field that has an underscore. forecast_time or pressure_levels.
It's better to validate for special fields that we are offering (like time.day, time.month, etc) and the user should still be able to groupby on fields that have an underscore.

xql/main.py
```diff
@@ -0,0 +1,149 @@
...
+
+    mask = inorder(where, ds)
+
+    filterd_ds = ds.where(mask, drop=True)
+
+    if group_by:
+        groupby_fields = [ e.args['this'].args['this'] for e in group_by.args['expressions'] ]
+        filterd_ds = apply_group_by(groupby_fields, filterd_ds, agg_funcs)
+
+    return filterd_ds
```
aniketinfocusp, line 133: I think it's better to break down this function into multiple sub functions.
It is currently doing a lot of things like
- getting the table name (or the zarr file name)
- gettings list of aggerating functions
- validations..etc

ideally this should be a class.

xql/main.py
```diff
@@ -0,0 +1,149 @@
...
+
+if __name__ == "__main__":
```
aniketinfocusp, line 135: A general comment for all the functions:
Please add
- doc strings to functions
- comments for complex logic
- return types.

Ideally a proper class should be made for other functions that are not to be exposed to a user (eg inorder, evaluate, parse) should be made private.

xql/main.py
```diff
@@ -0,0 +1,149 @@
...
+
+    while True:
+
+        query = input("xql>")
```
aniketinfocusp, line 139: nit: add a space `"xql> "`
also i wonder if this can handle multi line input.
People generally break SQL queries into multiple lines for better readability.

aniketinfocusp (commented):
Here are some suggestions from me

xql/main.py
DarshanSP19: Will add as an enhancement later.

xql/main.py
DarshanSP19: Added Support.

xql/main.py
DarshanSP19: Will add as an enhancement later.

xql/main.py
DarshanSP19: docstrings and return types are added. Other things we will add as an enhancement later.

xql/main.py
DarshanSP19: Will do as future enhancement.

mahrsee1997 (approved):
LGTM! Great work.

PR #428, opened 2023-12-29
Added check for skipping files and a force regrid flag.

aniketinfocusp:
This PR includes
- Files are skipped if they are present at output_path
- a `force_regrid` flag to bypass this behavior which is set to False by default.

### DIFF
--- a/weather_mv/loader_pipeline/regrid.py
+++ b/weather_mv/loader_pipeline/regrid.py
@@ -239,7 +239,7 @@ class Regrid(ToDataSink):
         if self.dry_run:
             return
 
-        if not self.force_regrid and path_exists(self.target_from(uri)):
+        if path_exists(self.target_from(uri), self.force_regrid):
             logger.info(f"Skipping {uri}")
             return
 
--- a/weather_mv/loader_pipeline/regrid.py
+++ b/weather_mv/loader_pipeline/regrid.py
@@ -28,7 +28,7 @@ import dask
 import xarray as xr
 import xarray_beam as xbeam
 
-from .sinks import ToDataSink, open_local, copy
+from .sinks import ToDataSink, open_local, copy, path_exists
 
 logger = logging.getLogger(__name__)
 
@@ -179,6 +179,7 @@ class Regrid(ToDataSink):
     """
     output_path: str
     regrid_kwargs: t.Dict
+    force_regrid: bool = False
     to_netcdf: bool = False
     zarr_input_chunks: t.Optional[t.Dict] = None
     zarr_output_chunks: t.Optional[t.Dict] = None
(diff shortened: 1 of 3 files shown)


weather_mv/loader_pipeline/regrid.py
```diff
@@ -236,6 +239,10 @@ def apply(self, uri: str):
         if self.dry_run:
             return
 
+        if not self.force_regrid and path_exists(self.target_from(uri)):
+            logger.info(f"Skipping {uri}")
```
mahrsee1997, line 243: nit: `.`.

mahrsee1997: Also lets add this check for zarr as well i.e. at line [272](https://github.com/google/weather-tools/pull/428/files#diff-7ee93f0ad7c757f82a6e90ea74dc1bdd9c4d04fb1f6c6b9539cd20fe7f2964a6R272).

weather_mv/loader_pipeline/sinks.py
```diff
@@ -380,6 +380,21 @@ def upload(src: str, dst: str) -> None:
     """Uploads a file to the specified GCS bucket destination."""
     subprocess.run(f'gsutil -m cp {src} {dst}'.split(), check=True, capture_output=True, text=True, input="n/n")
 
+def path_exists(path: str) -> bool:
```
mahrsee1997, line 383: More cleaner logic: https://github.com/google/weather-tools/blob/main/weather_sp/splitter_pipeline/file_splitters.py#L85C6-L85C6.

mahrsee1997 (changes requested):
Overall looks good ...

weather_mv/loader_pipeline/regrid.py
mahrsee1997: Refer to my below comment. Let's move `self.force_regrid` check in skip function.

weather_mv/loader_pipeline/sinks.py
mahrsee1997: And for checking in local file system, we can have a local store like: https://github.com/google/weather-tools/blob/main/weather_dl/download_pipeline/stores.py#L80

aniketinfocusp: updated in 
427038e

weather_mv/loader_pipeline/regrid.py
aniketinfocusp: done

weather_mv/loader_pipeline/sinks.py
```diff
@@ -380,6 +381,24 @@ def upload(src: str, dst: str) -> None:
...
+    """Check if path exists at a certain location. Pass force_regrid to skip checking."""
+    if force_regrid:
+        return False
+    # Try on gcs file system.
+    try:
+        matches = FileSystems().match([path])
+        assert len(matches) == 1
+        if len(matches[0].metadata_list) > 0:
+            return True
+    except Exception as _:
```
mahrsee1997, line 394: Let's catch specific exception.

aniketinfocusp: removed exception

mahrsee1997 (approved):
LGTM.

PR #434, opened 2024-01-22
Make xql package external use

DarshanSP19:
- Added `setup.py` file with package metadata and required dependencies.
- Removed `requirements.txt` file as not useful any more.
- Moved the code to the `xql/` directory to use as a package.


mahrsee1997 (approved):
LGTM.

PR #438, opened 2024-02-06
Xee Integration

DarshanSP19:
- Added [Xee](https://github.com/google/xee) package for accessing `EE` catalog.
- Made a separate dataset opener function to support different kinds of Dataset.
- From now on words `EE` assets can also be queries along with `zarr` datasets.


mahrsee1997 (approved):
LGTM.

PR #441, opened 2024-02-20
Set Chunks auto while opening zarr

DarshanSP19:
- Setting `chunks="auto"` while opening the `zarr` dataset.


mahrsee1997 (approved):
LGTM.

PR #443, opened 2024-02-21
Replace open_dataset with open_zarr while opening zarr

DarshanSP19:
Replace `open_dataset` with `open_zarr` while opening `zarr`.


dabhicusp (commented):
LGTM.

dabhicusp (approved):

PR #447, opened 2024-03-04
Added OrderBy Feature in xql

DarshanSP19:
- Added `orderby` support in xql.
- Removed unnecessary methods and relevant code.

### DIFF
--- a/xql/src/xql/apply.py
+++ b/xql/src/xql/apply.py
@@ -20,7 +20,6 @@ import pandas as pd
 import typing as t
 import xarray as xr
 
-from dask import dataframe as dd
 from sqlglot import parse_one, exp
 from xarray.core.groupby import DatasetGroupBy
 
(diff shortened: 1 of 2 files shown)


xql/src/xql/apply.py
```diff
@@ -85,79 +84,31 @@ def parse(a: t.Union[xr.DataArray, str], b: t.Union[xr.DataArray, str]) -> t.Tup
...
-        return ds[expression.args['this']]
-
-    if(expression.key == "literal"):
-        return expression.args['this']
-
-    args = expression.args
+    # Extract columns and sorting orders from the parsed ORDER BY expression
+    for el in e.expressions:
+        orderby_column = el.find(exp.Column).find(exp.Identifier).this
+        order = not bool(el.args['desc'])  # Descending if desc=False
```
dikwickley, line 104: this will throw `KeyError` in case `desc` is not present.
i tested this out in a notebook

```
import sqlglot
query = "Select a, b from my_table order by a ASC, b";
expr = sqlglot.parse_one(query)
expr

output:
Select(
  expressions=[
    Column(
      this=Identifier(this=a, quoted=False)),
    Column(
      this=Identifier(this=b, quoted=False))],
  from=From(
    this=Table(
      this=Identifier(this=my_table, quoted=False))),
  order=Order(
    expressions=[
      Ordered(
        this=Column(
          this=Identifier(this=a, quoted=False)),
        desc=False,
        nulls_first=True),
      Ordered(
        this=Column(
          this=Identifier(this=b, quoted=False)),
        nulls_first=True)]))
```

as you can see in the second ordered column, desc key is not present.

xql/src/xql/apply.py
```diff
@@ -85,79 +84,31 @@ def parse(a: t.Union[xr.DataArray, str], b: t.Union[xr.DataArray, str]) -> t.Tup
...
+        orderby_columns.append(orderby_column)
+        orderby_columns_order.append(order)
 
-    left = args['this']
-    right = None
-    if 'expression' in args:
-        right = args['expression']
+    # Sort the DataFrame based on the extracted columns and orders
+    df = df.sort_values(orderby_columns, ascending=orderby_columns_order)
```
dikwickley, line 109: Also found out that sqlglot does not throw any error in case you provide extra columns in the order by clause (that are not even present in the table), it will add that column to the expression.

Example query 
```
query = "Select a, b from my_table order by a ASC, b, c";
```

You can try to print out the expression tree using this query. (Note column 'c' is not in our table)

I am not sure how pandas sort_values will handle that, but i case it doesn't it should be handled by us.

xql/src/xql/apply.py
```diff
@@ -356,34 +307,39 @@ def filter_records(ds: xr.Dataset, query: str) -> dd.DataFrame:
...
 
     # Apply limit clause if present
     if limit_clause:
         limit = int(limit_clause.expression.args['this'])
-        ddf = ddf.loc[:start_loc + limit - 1]
+        df = df.iloc[:start_loc + limit]
 
     # Compute and return the filtered DataFrame
-    return ddf.compute()
+    return df.reset_index()
```
dikwickley, line 342: not too sure here, but shouldn't this be done at L323 (right after the `df = ds.to_dataframe()` call)?

xql/src/xql/apply.py
dikwickley: The ORDER BY keyword sorts the records in ascending order by default.
https://www.w3schools.com/sql/sql_orderby.asp

dikwickley: Also 😅 ,

image

dikwickley: checked, this works

DarshanSP19: `el.args['desc']` will give `None` if `desc` is not provided.

xql/src/xql/apply.py
DarshanSP19: We can call this anywhere.

xql/src/xql/apply.py
DarshanSP19: For now let's proceed with the constraint that `orderby` can only be applied either on coordinates or selected variables.

dikwickley:
LGTM

mahrsee1997 (approved):
LGTM!

PR #454, opened 2024-04-03
XQL - Only chunk filtered dataset

DarshanSP19:
While querying a large dataset like [gs://gcp-public-data-arco-era5/ar/full_37-1h-0p25deg-chunk-1.zarr-v3](https://pantheon.corp.google.com/storage/browser/gcp-public-data-arco-era5/ar/full_37-1h-0p25deg-chunk-1.zarr-v3?project=gcp-public-data-signals) the `xr.open_zarr()` call is taking too much time and memory.
If we open the `zarr` dataset with `chunks=None` and then chunk the dataset after applying query filters that will be much faster.

### DIFF
--- a/xql/src/xql/apply.py
+++ b/xql/src/xql/apply.py
@@ -23,7 +23,7 @@ import xarray as xr
 from sqlglot import parse_one, exp
 from xarray.core.groupby import DatasetGroupBy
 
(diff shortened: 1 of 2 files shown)


mahrsee1997 (approved):
LGTM!

PR #458, opened 2024-04-29
fix license job deletion

aniketsinghrawat:
We had moved from deployments to jobs for our license service.
The deletion logic that removed license deployments on license deletion was based on deployments. This was not working any more as licenses are now jobs

This PR fixes that.


mahrsee1997 (approved):
LGTM!

PR #459, opened 2024-05-03
Fixed the version of conda in the docker-file of weather-tools.

dabhicusp:
In the existing code while creating the `environment` of the `weather-tools` through docker image it is stuck in-between due to the latest version of the conda.
I fixed the conda version in the existing code so the above error will not come in future.

While creating the `environment` for `weather-tools` using the Docker image, the process was getting stuck due to issues with the updated version of conda. I've addressed this by specifying a fixed version of conda in the existing code. This change ensures that similar issues will not occur in the future.


mahrsee1997 (approved):
LGTM.

PR #462, opened 2024-05-13
`dl-v2` download add priority

aniketsinghrawat:
This PR adds a `--priority` flag when submitting a new job that sets the priority of that config in all the queues.
If none is passed, the default behavior is to append the config at the end of the queue.


weather_dl_v2/cli/app/subcommands/download.py
```diff
@@ -75,8 +75,17 @@ def submit_download(
             help="Force redownload of partitions that were previously downloaded.",
         ),
     ] = False,
+    priority: Annotated[
+        int | None,
+        typer.Option(
+            "-p",
+            "--priority",
+            help="Set the priority for submitted config in ALL licenses. If not added, the config is added" \
+                "at the end of the queue.",
```
mahrsee1997, line 84: nit: let's mention '0' being the highest priority.

weather_dl_v2/fastapi-server/config_processing/pipeline.py
```diff
@@ -63,6 +63,9 @@ async def start_processing_config(config_file, licenses, force_download):
             config_name, "Partitioning completed."
         )
         await queue_handler._update_queues_on_start_download(config_name, licenses)
+        if priority:
+            for license_id in licenses:
+                await queue_handler._update_config_priority_in_license(license_id, config_name, priority)
```
mahrsee1997, line 68: Discussed offline, let's try to do this one call; instead of making 2 DB calls.

aniketsinghrawat: done

weather_dl_v2/cli/app/subcommands/download.py
aniketsinghrawat: done

mahrsee1997 (approved):
LGTM

PR #464, opened 2024-07-11
Letting the logs of convert_to_cog, which runs in a separate process, be visible in Dataflow.

deepgabani8:


mahrsee1997 (approved):
LGTM.

PR #469, opened 2024-08-06
`dl-v2` new cluster config

aniketsinghrawat:
This PR includes

- changes to the cluster config.
  - machines in default pool are downgraded to e2-standard-8 and now each license deployment will occupy one whole node. Same with server.
  - Licenses Jobs and Server Deployment will live in the default pool. Removed server-pool and license-pool.


mahrsee1997 (approved):
LGTM.

PR #470, opened 2024-08-07
Metrics update return type and Error handling

aniketsinghrawat:
The metrics decorator was causing apache-beam typehints error. This PR adds a return type to circumvent that.


mahrsee1997 (approved):
LGTM

PR #473, opened 2024-09-02
Improved efficiency of weather-mv bq in terms of time and cost.

mahrsee1997:
### Stats:

Tested on dataset (single file) (longitude: 1440; latitude: 721; level: 13; time: 1; number of data variables: 10):
```
<xarray.Dataset>
Dimensions:                  (longitude: 1440, latitude: 721, level: 13, time: 1)
Coordinates:
  * longitude                (longitude) float32 0.0 0.25 0.5 ... 359.5 359.8
  * latitude                 (latitude) float32 -90.0 -89.75 ... 89.75 90.0
  * level                    (level) int32 50 100 150 200 ... 700 850 925 1000
  * time                     (time) timedelta64[ns] 06:00:00
    datetime                 (time) datetime64[ns] ...
Data variables:
    10m_u_component_of_wind  (time, latitude, longitude) float32 ...
    10m_v_component_of_wind  (time, latitude, longitude) float32 ...
    2m_temperature           (time, latitude, longitude) float32 ...
    mean_sea_level_pressure  (time, latitude, longitude) float32 ...
    geopotential             (time, level, latitude, longitude) float32 ...
    specific_humidity        (time, level, latitude, longitude) float32 ...
    temperature              (time, level, latitude, longitude) float32 ...
    u_component_of_wind      (time, level, latitude, longitude) float32 ...
    v_component_of_wind      (time, level, latitude, longitude) float32 ...
    vertical_velocity        (time, level, latitude, longitude) float32 ...
```
Total number of rows to be ingested into BQ: 13,497,120 (1440 * 721 * 13 *1)

Ran on DataflowRunner.
machine_type: n1-standard1; 4gb RAM; 100GB HDD.

| Branch | Time Taken | Cost | Autoscaled max to |
| ------------- | ------------- | ------------- | ------------- |
| main | 48 min |  $9.33 | 625 workers |
| mv-opitimization  | 36 min  | $0.10 | 6 workers |

Note: In our development project, we have 1000 workers (with no resource restrictions). However, in a real-world scenario, users might not have this many workers, so the time and cost with the main branch would have been significantly higher.

### Approach:

* We calculate latitude, longitude, geo_point, and geo_polygon information upfront and dump it to a parquet file so that we do not need to process it every time we process a set of files.
* Previously, we created indexes across all the index dimensions (e.g., lat, lon, time, level) and then selected rows from the dataset based on these coordinates. This resulted in a high number of I/O calls.
* Now, we only create indexes across all the index dimensions except for latitude and longitude, thereby reducing the number of coordinates and, consequently, the number of I/O calls.
* We use pandas DataFrame and its methods to generate rows instead of iterating over each row with a for loop.
* Using `--rows_chunk_size <chunk-size>`, users can control the number of rows loaded into memory for processing, depending on their system's memory.

**Assumption:** A minimum of this much memory is available to load all the data variables for (lat × lon) plus a single indexed (apart from lat & lon) at once. I think we can make this assumption because, for a 0.1 resolution dataset (3600 × 1800) with 51 data variables, only 9 GiB of RAM is required.

ps: From the learnings of ARCO-ERA5 to BQ ingestion.


weather_mv/loader_pipeline/bq.py
```diff
@@ -101,8 +111,9 @@ class ToBigQuery(ToDataSink):
     tif_metadata_for_end_time: t.Optional[str]
     skip_region_validation: bool
     disable_grib_schema_normalization: bool
-    coordinate_chunk_size: int = 10_000
+    rows_chunk_size: int = 1000000
```
alxmrs, line 114: For readability: 1_000_000.

weather_mv/loader_pipeline/bq.py
```diff
@@ -174,6 +194,49 @@ def validate_arguments(cls, known_args: argparse.Namespace, pipeline_args: t.Lis
...
+    def generate_csv(
+        self,
+        csv_path: str,
+        lats: t.List,
+        lons: t.List,
+        lat_grid_resolution: float,
+        lon_grid_resolution: float,
+        skip_creating_polygon: bool = False,
+    ):
+        """Generates geo data CSV."""
```
alxmrs, line 206: What do you think about using parquet, or geoparquet here instead of a CVS?

weather_mv/loader_pipeline/bq.py
```diff
@@ -174,6 +194,49 @@ def validate_arguments(cls, known_args: argparse.Namespace, pipeline_args: t.Lis
...
+                if not skip_creating_polygon:
+                    geo_polygon = fetch_geo_polygon(lat, sanitized_lon, lat_grid_resolution, lon_grid_resolution)
+                    row.append(geo_polygon)
+                else:
+                    row.append(None)
+                data.append(row)
+
+            df = pd.DataFrame(data, columns=header)
+            # Write DataFrame to CSV.
+            df.to_csv(temp.name, index=False)
```
alxmrs, line 235: If this was parquet, for example, you could directly write to the bucket.

weather_mv/loader_pipeline/bq.py
```diff
@@ -227,24 +308,19 @@ def __post_init__(self):
...
 
-    def prepare_coordinates(self, uri: str) -> t.Iterator[t.Tuple[str, t.List[t.Dict]]]:
+    def prepare_coordinates(self, uri: str) -> t.Iterator[t.Tuple[str, t.Dict]]:
         """Open the dataset, filter by area, and prepare chunks of coordinates for parallel ingestion into BigQuery."""
         logger.info(f'Preparing coordinates for: {uri!r}.')
 
         with open_dataset(uri, self.xarray_open_dataset_kwargs, self.disable_grib_schema_normalization,
                           self.tif_metadata_for_start_time, self.tif_metadata_for_end_time, is_zarr=self.zarr) as ds:
             data_ds: xr.Dataset = _only_target_vars(ds, self.variables)
-            if self.area:
```
alxmrs, line 237: Nice

weather_mv/loader_pipeline/bq.py
```diff
@@ -253,57 +329,69 @@ def extract_rows(self, uri: str, coordinates: t.List[t.Dict]) -> t.Iterator[t.Di
...
-            # {'d': -2.0187, 'cc': 0.007812, 'z': 50049.8, 'rr': None}
-            row = {n: to_json_serializable_type(ensure_us_time_resolution(v.values))
-                   for n, v in row_ds.data_vars.items()}
+            # Ensure that the latitude and longitude dimensions are in sync with the geo data CSV.
+            if not BQ_EXCLUDE_COORDS - set(selected_ds.dims.keys()):
+                selected_ds = selected_ds.transpose('latitude', 'longitude')
 
-            # Serialize coordinates.
-            it = {k: to_json_serializable_type(v) for k, v in it.items()}
+            master_df = pd.read_csv(master_lat_lon)
```
alxmrs, line 351: Let’s use a different term from “master”.

weather_mv/loader_pipeline/bq.py
```diff
@@ -253,57 +329,69 @@ def extract_rows(self, uri: str, coordinates: t.List[t.Dict]) -> t.Iterator[t.Di
...
-                if not self.skip_creating_polygon
-                else None
-            )
-            # 'row' ends up looking like:
-            # {'latitude': 88.0, 'longitude': 2.0, 'time': '2015-01-01 06:00:00', 'd': -2.0187, 'cc': 0.007812,
-            #  'z': 50049.8, 'data_import_time': '2020-12-05 00:12:02.424573 UTC', ...}
-            metric.Metrics.counter('Success', 'ExtractRows').inc()
-            yield row
+            # Filter out excluded coordinates from coords.
+            filtered_coords = [c for c in selected_ds.coords if c not in BQ_EXCLUDE_COORDS]
```
alxmrs, line 361: Maybe you could do one less loop if this was a generator comprehension ().

weather_mv/loader_pipeline/bq.py
```diff
@@ -253,57 +329,69 @@ def extract_rows(self, uri: str, coordinates: t.List[t.Dict]) -> t.Iterator[t.Di
...
+            for c in filtered_coords:
+                if c not in coordinate and (not self.variables or c in self.variables):
+                    master_df[c] = to_json_serializable_type(ensure_us_time_resolution(selected_ds[c].values))
+
+            # We are not directly assigning values to dataframe because we need to consider 'None'.
```
alxmrs, line 366: Can you explain this to me more? I do think vectorization would be great if we could get away with it.

weather_mv/loader_pipeline/sinks.py
```diff
@@ -397,15 +397,15 @@ def path_exists(path: str, force_regrid: bool = False) -> bool:
 def copy(src: str, dst: str) -> None:
     """Copy data via `gcloud alpha storage` or `gsutil`."""
     errors: t.List[subprocess.CalledProcessError] = []
-    for cmd in ['gcloud alpha storage cp', 'gsutil cp']:
+    for cmd in ['gsutil -m cp']:
```
alxmrs, line 400: Nice

alxmrs (commented):
I really like this new approach, it’s very smart. The stats looks great, I’m impressed. 

I think it may be nice to revisit vecotrization later, but the optimizations in IO will have a greater effect.

weather_mv/loader_pipeline/bq.py
mahrsee1997: How about `vector_df`, `vector_data` or `rows_data` ? any other suggestion ?

weather_mv/loader_pipeline/bq.py
mahrsee1997: Done.

weather_mv/loader_pipeline/bq.py
mahrsee1997: Thanks Alex, this make sense. Moved to parquet; file size reduced to 91 MiB compare to 890 MiB.

weather_mv/loader_pipeline/bq.py
mahrsee1997: I benchmarked this, and it looks like the current approach is faster! :)

weather_mv/loader_pipeline/bq.py
mahrsee1997: umm I did not get the idea here :). Can you please elaborate.

weather_mv/loader_pipeline/bq.py
mahrsee1997: Okay couple of points to note here:

1) Assume that I have some `None` data & I simply do `master_df[var] = data`; it would convert that `None` to `nan`. That's why I used  `pd.Series(values, dtype=object)` (dtype=object) to preserve `None`.

2) I don't think we can eliminate iterating over `selected_ds.data_vars`.

3) I suppose we are doing vectorization already when doing `master_df[var] = pd.Series(values, dtype=object)`. Isn't it ?

Maybe I'm missing something ?

weather_mv/loader_pipeline/bq.py
alxmrs: I like the first two. How about `vector_df`?

weather_mv/loader_pipeline/bq.py
alxmrs: Replace the `[…]` with `(…)` and it will become a generator expression (evaluated lazily).

weather_mv/loader_pipeline/bq.py
```diff
@@ -174,6 +195,49 @@ def validate_arguments(cls, known_args: argparse.Namespace, pipeline_args: t.Lis
...
+                # Fetch the geo polygon if not skipped.
+                if not skip_creating_polygon:
+                    geo_polygon = fetch_geo_polygon(lat, sanitized_lon, lat_grid_resolution, lon_grid_resolution)
+                    row.append(geo_polygon)
+                else:
+                    row.append(None)
+                data.append(row)
+
+            df = pd.DataFrame(data, columns=header)
+            # Write DataFrame to parquet.
```
alxmrs, line 235: Sorry, I haven’t been specific enough. Parquet is a cloud optimized format, so I don’t think you need to create a temporary file and then copy it. You should be able to go from memory directly to cloud bucket withthin the pandas dataframe (maybe, at least, via a dask dataframe).

alxmrs: You know what? I did some more googling and it seems I’m mistaken. They way folks write parquet to GCS is via gcsfs. That would be a good option, but it is a departure to how we do it so far. So I’m happy either way. Sorry for the trouble!

weather_mv/loader_pipeline/bq.py
alxmrs: To be honest, I haven’t read the code too closely. I think it is good, given your explanation. This is secondary to your main improvement area, and we can investigate this later.

mahrsee1997: Yes, make sense :)

weather_mv/loader_pipeline/bq.py
mahrsee1997: ah ! my bad ... got it. Changes done.

weather_mv/loader_pipeline/bq.py
mahrsee1997: Yes, this sounds good.  Necessary code changes done.

weather_mv/loader_pipeline/bq.py
mahrsee1997: Yeah :), for now, keeping it as it is.

alxmrs:
Since I’m not employed at Google right now, I can’t in good conscience give this an approval. I think @fredzyda would be a better person to decide if this should be merged. I will say, this patch looks good to me.

fredzyda (approved):
Looks good. Very nice improvement in runtime and cost.

mahrsee1997:
Thanks @alxmrs and @fredzyda for the review!

PR #477, opened 2024-09-06
Removed default accuracy in regrid.

Piyush-Ingale:
Removed default accuracy=12 changes from weather-mv regrid. This is to maintain the Metview's default behaviour. Metview takes accuracy of input data as default if user does not pass any accuracy.


mahrsee1997 (approved):
LGTM

PR #494, opened 2025-01-27
Fixed --dry-run flag is in weather-mv ee tool

### LINKED ISSUE #493: `--dry-run` is not working in `weather-mv ee` tool.
There's a bug specifically at line: https://github.com/google/weather-tools/blob/main/weather_mv/loader_pipeline/pipeline.py#L84. The `pipeline_options_dict` also contains a `dry_run` flag, which overrides the `known_args` (user-passed) flag value.

Thanks, @shail-parekh for finding this! :)

shail-parekh:
- Fixed a bug where dry-run flag is not working as expected.

Fixes #493


mahrsee1997 (approved):
LGTM! thanks for identifying this bug & of-course for the PR :)

PR #499, opened 2025-03-04
Refactor/add project name running locally

shail-parekh:
- Added a parameter called project_name in the ee.Initialize() when running locally.

### DIFF
--- a/weather_mv/loader_pipeline/ee.py
+++ b/weather_mv/loader_pipeline/ee.py
@@ -120,9 +120,12 @@ def ee_initialize(project_id: t.Optional[str] = None,
     on_compute_engine = is_compute_engine()
     # Using the high volume api.
     if on_compute_engine:
-        if project_id is None:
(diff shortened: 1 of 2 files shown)


Piyush-Ingale (approved):
LGTM!

weather_mv/loader_pipeline/ee.py
```diff
@@ -101,7 +101,8 @@ def get_creds(use_personal_account: bool, service_account: str, private_key: str
     return creds
 
 
-def ee_initialize(use_personal_account: bool = False,
+def ee_initialize(project_id: t.Optional[str] = None,
```
mahrsee1997, line 104: Let's add the new argument at the end, because if someone uses ee_initialize without passing the keywords, it will likely fail.

shail-parekh: - Code updated with requested changes!

mahrsee1997 (approved):
LGTM!

PR #507, opened 2025-07-16
Moved the `path_exists` function from sinks.py to regrid.py file.

j9sh264:
Moved the `path_exists` function from sinks.py to regrid.py file.


mahrsee1997 (approved):
LGTM!

Please bump version.

PR #508, opened 2025-07-25
Resolving multiple issues w.r.t. `GribSplitterV2`.

shail-parekh:
Resolve time padding and multiple type support in weather-splitter - specifically supported by the `GribSplitterV2` implementation:
- Ensure time values like '600' or '0' are zero-padded to 4 digits (e.g., '0600', '0000').
- Add support for handling multiple types of values within a single Grib file such as different levels (e.g. surface and pressure levels) within a single Grib file.

### DIFF
--- a/weather_sp/splitter_pipeline/file_splitters.py
+++ b/weather_sp/splitter_pipeline/file_splitters.py
@@ -193,11 +193,11 @@ class GribSplitterV2(GribSplitter):
         split_dims_arg = ','.join(f'{dim}:s' for dim in split_dims)
         with self._copy_to_local_file() as local_file:
             self.logger.info('Skipping as needed...')
-            grib_get_args = [grib_get_cmd, '-p', split_dims_arg]
             # Append -w flag to filter GRIB messages matching the given expression
             if self.grib_filter_expression:
-                grib_get_args.extend(['-w', self.grib_filter_expression])
-            grib_get_args.append(local_file.name)
+                grib_get_args = [grib_get_cmd, '-p', split_dims_arg, '-w', self.grib_filter_expression, local_file.name]
+            else:
+                grib_get_args = [grib_get_cmd, '-p', split_dims_arg, local_file.name]
             grib_get_process = subprocess.Popen(grib_get_args, stdout=subprocess.PIPE)
             uniq_output = subprocess.check_output((uniq_cmd,), stdin=grib_get_process.stdout)
             output_paths = []
--- a/weather_sp/splitter_pipeline/file_splitters.py
+++ b/weather_sp/splitter_pipeline/file_splitters.py
@@ -50,7 +50,8 @@ class FileSplitter(abc.ABC):
     """Base class for weather file splitters."""
 
     def __init__(self, input_path: str, output_info: OutFileInfo,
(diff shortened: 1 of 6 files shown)


weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -185,18 +187,16 @@ def split_data(self) -> None:
...
-
         split_dims = self.output_info.split_dims()
-        split_dims_arg = ','.join(split_dims)
         with self._copy_to_local_file() as local_file:
             self.logger.info('Skipping as needed...')
-            grib_get_process = subprocess.Popen((grib_get_cmd, '-p', split_dims_arg, local_file.name),
-                                                stdout=subprocess.PIPE)
-            uniq_output = subprocess.check_output((uniq_cmd,), stdin=grib_get_process.stdout)
+            grib_fields = mv.read(local_file.name)
+            split_values = [tuple(mv.grib_get(grib_fields[i], split_dims)[0]) for i in range(len(grib_fields))]
```
Piyush-Ingale, line 194: I think there is no need to iterate over the GRIB messages explicitly.
`mv.grib_get(grib_fields, split_dims)`  should return the list of lists of required fields for all the GRIB messages.

weather_sp/splitter_pipeline/pipeline.py
```diff
@@ -119,7 +121,15 @@ def run(argv: t.List[str], save_main_session: bool = True):
     parser.add_argument('-f', '--force', action='store_true', default=False,
                         help='Force re-splitting of the pipeline. Turns of skipping of already split data.')
     parser.add_argument('--log-level', type=int, default=2,
-                      help='An integer to configure log level. Default: 2(INFO)')
+                        help='An integer to configure log level. Default: 2(INFO)')
+    parser.add_argument('-w', '--where', default=None,
```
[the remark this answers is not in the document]
shail-parekh: done

weather_sp/splitter_pipeline/pipeline.py
```diff
@@ -42,6 +42,7 @@ def split_file(input_file: str,
                output_dir: t.Optional[str],
                formatting: str,
                dry_run: bool,
+               grib_filter_expression: t.Optional[str],
```
mahrsee1997, line 45: Always try to add the new argument at the end to maintain backward compatibility.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -22,6 +22,7 @@
 import subprocess
 import tempfile
 import typing as t
+import metview as mv
```
mahrsee1997, line 25: Should we do something like this here ? https://github.com/google/weather-tools/blob/main/weather_mv/loader_pipeline/regrid.py#L37.

@Piyush-Ingale ?

mahrsee1997 (changes requested):
(1) Bump weather-sp version in setup.py.

(2) What about other types of splitter ? do we have similar problem in them ?

weather_sp/splitter_pipeline/pipeline.py
shail-parekh: Done.

PR #510, opened 2025-08-08
Reposition the decompressed file pointer to the start.

j9sh264:
- Sometimes when regriding some `.bz2` files they are flagged as corrupted. 
- When skipping the corrupt file check the regrid process fails with a *Segmentation fault*.
- However they are not corrupted but instead their pointer gets shifted after decompressing them in `open_local`.

These changes will help in repositioning the file pointer to the start of the file so that is gets processed properly.


mahrsee1997 (approved):
LGTM.

PR #511, opened 2025-09-04
Support compressed file objects in the splitter pipeline.

j9sh264:
Updated the `_copy_to_local_file` class method for the **FileSplitter**s to support compressed data.


mahrsee1997 (approved):
LGTM! please bump the version in setup.py

PR #536, opened 2026-05-02
Support datetime expressions in weather-sp output templates

### LINKED ISSUE #535: [weather-sp] : better formatting of date & time in output template
## Example:
For a file `gs://test-input/test-file.grib` the output file pattern is `gs://test-output/splits/test-file_{date}-{time}.grib`.
Then for date = 20200101, time = 1200, the output file for that split will be `gs://test-output/splits/test-file_20200101-1200.grib`.

But what if user want's output file as `gs://test-output/splits/test-file_20200101-12:00:00.grib` -- basically format date & time as per the need.

For such cases he can provide output file pattern as `gs://test-output/splits/test-file_{date}-{datetime.strptime(f"{time}", "%H%M").strftime("%H:%M:%S")}.grib`.

Make necessary changes for supporting this :) most probably just need to import `datetime` library and minor tweaking in logic.

zeel2104:
## Summary

Fixes #535.

Adds support for richer Python-style expressions in `weather-sp` output templates, including datetime formatting expressions like `{datetime.strptime(f"{time}", "%H%M").strftime("%H:%M:%S")}`.

This allows an output template such as:

`gs://test-output/splits/test-file_{date}-{datetime.strptime(f"{time}", "%H%M").strftime("%H:%M:%S")}.grib`

to produce:

`gs://test-output/splits/test-file_20200101-12:00:00.grib`

## Changes

- Added expression-aware output template parsing/evaluation in `OutFileInfo`.
- Updated split-dimension detection so variables used inside expressions, such as `time`, are included.
- Adjusted `GribSplitterV2` to use raw split-dimension temp filenames for `grib_copy`, then apply the richer Python formatter during upload.
- Added tests for datetime expression formatting and split-dimension extraction.

## Testing

Manual validation passed:

`['date', 'time']`

`gs://test-output/splits/test-file_20200101-12:00:00.grib`

Full pytest was not run locally because this Windows Python 3.14 environment is missing `pygrib`/ecCodes support.

### DIFF
--- a/weather_sp/splitter_pipeline/file_name_utils.py
+++ b/weather_sp/splitter_pipeline/file_name_utils.py
@@ -13,15 +13,158 @@
 # limitations under the License.
 
 from dataclasses import dataclass
+import ast
+import datetime as _datetime
 import logging
 import os
-import string
 import typing as t
 
 logger = logging.getLogger(__name__)
 
 GRIB_FILE_ENDINGS = ('.grib', '.grb', '.grb2', '.grib2', '.gb')
 NETCDF_FILE_ENDINGS = ('.nc', '.cd')
+
+
+class _DateTimeNamespace:
+    date = _datetime.date
+    datetime = _datetime.datetime
+    time = _datetime.time
+    timedelta = _datetime.timedelta
+    timezone = _datetime.timezone
+    strptime = staticmethod(_datetime.datetime.strptime)
+
+
+_TEMPLATE_GLOBALS = {
+    '__builtins__': {'__import__': __import__},
+    'datetime': _DateTimeNamespace,
+    'float': float,
+    'int': int,
+    'str': str,
+}
+
+
+def _iter_template_fields(template: str) -> t.Iterator[t.Tuple[str, str]]:
+    """Yield literal text and replacement fields from a format template."""
+    literal = []
+    i = 0
+    while i < len(template):
+        char = template[i]
+        if char == '{':
+            if i + 1 < len(template) and template[i + 1] == '{':
+                literal.append('{')
+                i += 2
+                continue
+            if literal:
+                yield ''.join(literal), ''
+                literal = []
+            field, i = _read_template_field(template, i + 1)
+            yield '', field
+        elif char == '}':
+            if i + 1 < len(template) and template[i + 1] == '}':
+                literal.append('}')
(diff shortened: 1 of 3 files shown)


mahrsee1997:
@zeel2104, can you please fix `lint` issue, thanks!

zeel2104:
@mahrsee1997 

Thanks for pointing it out! I fixed the lint issue by wrapping the long datetime output template in the test file and pushed the update.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -13,9 +13,10 @@
 # limitations under the License.
 
 from dataclasses import dataclass
+import ast
```
shail-parekh, line 16: Instead of using a complex AST-based parser, can we implement a simple native wrapper for date and time that preserves the raw GRIB date/time field formats by default?

weather_sp/splitter_pipeline/file_splitters.py
shail-parekh, line 1: With this approach, if a pipeline is initiated with an output-template using double quotes or without quotes, it can cause the pipeline to fail. We should handle these scenarios as well.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -24,6 +25,148 @@
 NETCDF_FILE_ENDINGS = ('.nc', '.cd')
 
 
+class _DateTimeNamespace:
```
shail-parekh, line 28: we can simplify the codebase by removing this class and helper methods, as we can remove the AST based approach!

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -50,13 +193,41 @@ def unformatted_output_path(self):
...
+        for _, field in _iter_template_fields(self.unformatted_output_path()):
+            if not field:
+                continue
+            field_name, _, _ = _split_field(field)
+            for name in _names_in_expression(field_name):
+                if name not in dims:
+                    dims.append(name)
+        return dims
 
     def formatted_output_path(self, splits: t.Dict[str, str]) -> str:
```
shail-parekh, line 206: Please update formatted_output_path to use native Python string formatting.

PR #540, opened 2026-05-29
Support single-message GRIBs for year-wise regridding.

j9sh264:
The changes ensure the pipeline successfully processes GRIB files containing a single message where Metview's `fs.base_date()` returns a scalar instead of a list:

- Exclude string/bytes types from the iterable check to prevent crash.
- Gracefully fallback to scalar date year extraction when no list is returned.
- Add guards to check for null/missing dates.


Piyush-Ingale (approved):
LGTM!

PR #548, opened 2026-06-12
Issue 538: Skip individual split files that already exist

### LINKED ISSUE #538: [weather-sp]: Skip creating split files that already exist
## Problem
Currently in `weather-sp`, the file splitter skips processing an input file only when all its split children already exist. If even one of the split files is missing (for example, due to a partial failure or interrupted run), the splitter re-splits the entire file and overwrites all the previously created children.

This leads to redundant processing and unnecessary I/O, especially when re-running pipelines to fill in missing data.

## Proposed Solution
Add functionality to skip the creation of individual split files if they already exist in the output location.

Instead of an all-or-nothing approach at the input file level, the splitter should:

1. Identify which specific split files are expected to be produced.
2. Check if each file already exists.
3. Only generate and upload the missing files, leaving the existing ones intact.

mmarusiak:
Fixes #538 

Previously, all three splitters used an all-or-nothing approach: if any output file existed (via wildcard match), the entire input file was skipped. If even one output file was missing, the splitter re-split the whole input and overwrote all previously created children.

Now each splitter checks existence per output file and only generates the missing ones:

- GribSplitter: replaced the upfront should_skip() guard with a per-key should_skip_file() check inside the message loop, using a skipped_keys set to avoid redundant filesystem calls for the same output path.
- GribSplitterV2: already checked per-file before running grib_copy, but the upload loop was copying every file grib_copy produced (including already-existing ones). Fixed by filtering the upload loop to only paths in output_paths.
- NetCdfSplitter: removed the upfront should_skip() guard and moved the existence check into _write_dataset(), which now returns a bool so split_data() can track how many files were actually written.

Added test_splits_only_missing_files for both GribSplitter/GribSplitterV2 (parametrized) and NetCdfSplitter: full split, delete one output file, re-split, assert only the deleted file was recreated and all others have unchanged modification times.


weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -166,10 +161,19 @@ def split_data(self) -> None:
...
+                    if key in skipped_keys:
+                        del grb
+                        continue
+
                     # Append the current grib message to a set number of output files.
                     # If the target shard doesn't exist, create it.
                     if key not in outputs:
+                        if self.should_skip_file(key):
+                            skipped_keys.add(key)
+                            del grb
```
j9sh264, line 196: We should log these skipped files to improve pipeline telemetry and simplify troubleshooting downstream.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -264,6 +275,8 @@ def split_data(self) -> None:
                 self.logger.info('Uploading %r...', self.input_path)
                 for flat_target in os.listdir(tmpdir):
                     dest_file_path = f'{prefix}{flat_target.replace(delimiter, slash)}'
+                    if dest_file_path not in output_paths_set:
+                        continue
```
j9sh264, line 321: Same here, let's log these skipped files.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -317,14 +332,18 @@ def _open_dataset_locally(self) -> t.Iterator[xr.Dataset]:
...
-    def _write_dataset(self, dataset: xr.Dataset, split_dims: t.List[str]) -> None:
-        """Write destination NetCDF file in NETCDF4 format."""
+    def _write_dataset(self, dataset: xr.Dataset, split_dims: t.List[str]) -> bool:
+        """Write destination NetCDF file in NETCDF4 format. Returns True if the file was written."""
         # Here, we need to write the file locally, since only the scipy engine supports file objects or
         # returning bytes. Further, the scipy engine does not support NETCDF4 (which is HDF5 compliant).
         # Storing data in HDF5 is advantageous since it allows opening NetCDF files with buffered readers.
+        output_path = self._get_output_for_dataset(dataset, split_dims)
+        if self.should_skip_file(output_path):
+            return False
```
j9sh264, line 393: Same here, let's log these skipped files :)

Alternatively, we can also add a log in the `should_skip_file` method.

j9sh264:
Thank you for implementing these changes @mmarusiak!

Could you please pull the latest from main and bump the version numbers in both `weather-sp/setup.py` and `VERSION.txt`?

j9sh264:
Hey @mmarusiak, checking in on this PR!

No rush, but wanted to see if you had a chance to review the feedback. If you're blocked on anything or need a hand addressing the comments, just let us know. Happy to help!

mmarusiak:
@j9sh264 sorry for the delay, but finally managed to find some time to rebase and bump version!

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -296,9 +309,17 @@ def split_data(self) -> None:
...
-                files = os.listdir(tmpdir)
-                num_files = len(files)
-                for f in files:
+                # grib_copy generates a file for every message, including ones
+                # that were already split. Drop those so copy_dir won't re-upload
+                # them (issue #538), then flatten the DELIMITER-encoded names back
+                # into their nested directory structure for the recursive copy.
+                num_files = 0
+                for f in os.listdir(tmpdir):
+                    dest_file_path = f'{prefix}{f.replace(delimiter, "/")}'
```
j9sh264, line 318: Let's use `os.path.join` to create the path.

j9sh264:
Thanks for your contributions here! Since we needed to get this merged, we incorporated your changes along with the necessary updates in #560.
Appreciate the work and input from both of you!

PR #549, opened 2026-06-17
Refactor empty filter handling in GribSplitterV2

j9sh264:
fix(splitter): gracefully handle empty GRIB filter matches

When applying a `--where` filter expression in GribSplitterV2, files containing 
zero matching messages previously caused a crash or required a broad try/except 
block in the pipeline that swallowed unexpected errors.

This change:
- Detects when `grib_get` returns an empty string (no filter matches).
- Gracefully logs a warning and returns early instead of crashing.
- Increments a new `skipped_empty_filter` Beam metric for observability.


mahrsee1997 (changes requested):
LGTM!
Please bump version of setup.py and Version.txt.

mahrsee1997 (approved):
LGTM!

Just bump the version.txt. Thanks!

PR #554, opened 2026-07-19
feat: support datetime formatting in output template

### LINKED ISSUE #535: [weather-sp] : better formatting of date & time in output template

wasim-builds:
Closes #535

## Summary

Adds datetime formatting support to the output template in `file_name_utils.py`. Users can now use arbitrary f-string expressions with `datetime` objects in file name patterns.

## Example

Before:
```
pattern = "weather_{date}"
# output: weather_2026-07-31
```

After:
```
pattern = "weather_{date:%Y%m%d_%H%M}"
# output: weather_20260731_0000
```

## Changes

- `weather_sp/splitter_pipeline/file_name_utils.py`: support `datetime` formatting in f-string evaluation
- `weather_sp/splitter_pipeline/file_name_utils_test.py`: add tests covering datetime format strings, fallback behavior, and edge cases

## Verification

Tests cover:
- Standard datetime format codes (`%Y`, `%m`, `%d`, `%H`, `%M`, `%S`)
- Fallback to string representation when format is invalid
- Preservation of existing non-dynamic template behavior


weather_sp/splitter_pipeline/file_name_utils.py
shail-parekh, line 1: Please add the corresponding unit test in `file_name_utils_test.py` to validate the new datetime formatting feature.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -56,7 +56,21 @@ def split_dims(self) -> t.List[str]:
 
     def formatted_output_path(self, splits: t.Dict[str, str]) -> str:
```
shail-parekh, line 91: Please test that the `output-template` is handled correctly regardless of whether it is used with single quotes, double quotes, or no quotes at all!

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -56,7 +56,21 @@ def split_dims(self) -> t.List[str]:
...
         """Construct output file name with formatting applied"""
-        return self.unformatted_output_path().format(*self.template_folders, **splits)
+        import datetime
```
shail-parekh: kindly move the import datetime statement to the very beginning of the file.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -56,7 +56,21 @@ def split_dims(self) -> t.List[str]:
...
+        template = self.unformatted_output_path()
+        
+        # Safe globals dictionary containing only datetime
+        safe_globals = {"datetime": datetime}
+        
+        try:
+            # First, substitute the positional args (if any) to avoid invalid f-string syntax like {0}
+            formatted_positionals = template.format(*self.template_folders, **{k: '{'+k+'}' for k in splits.keys()})
```
shail-parekh: template.format(...) doesn't support nested braces or function calls like {datetime.strptime(...), so it will raise a `ValueError`. When this happens, it falls into the except block and tries to call `template.format()` again, which will crash the pipeline.  please fix this!

weather_sp/splitter_pipeline/file_name_utils.py
shail-parekh, line 1: please check and update the other functions in the pipeline that will not work as expected when they encounter these type of output-template!

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -50,13 +53,55 @@ def unformatted_output_path(self):
         return self.file_name_template + self.formatting + self.ending
 
     def split_dims(self) -> t.List[str]:
-        all_format = list(filter(None, [field[1] for field in string.Formatter().parse(
-            self.unformatted_output_path())]))
-        return [key for key in all_format if not key.isdigit()]
+        keys = []
+        for field in string.Formatter().parse(self.unformatted_output_path()):
```
shail-parekh: Python's built-in `string.Formatter` crashes with a ValueError if a user includes a nested brace (like f"{time}") in their template.  please fix this.

weather_sp/splitter_pipeline/file_name_utils.py
```diff
@@ -50,13 +53,55 @@ def unformatted_output_path(self):
...
+        # Safe globals dictionary containing only datetime
+        safe_globals = {"datetime": datetime}
+
+        splits_with_pos = splits.copy()
+        for i, folder in enumerate(self.template_folders):
+            splits_with_pos[f'_{i}'] = folder
+
+        try:
+            # evaluate as an f-string using repr to handle quotes correctly
+            return eval('f' + repr(template), safe_globals, splits_with_pos)
```
shail-parekh: Evaluating the entire path at once as a f-string causes syntax-error. if the user has standard quotes in their template like '%H%M'.

weather_sp/splitter_pipeline/file_splitters.py
```diff
@@ -239,7 +244,14 @@ def split_data(self) -> None:
 
         # Replace { with [ and } with ] only for non-numeric values inside {} of tail
         output_str = re.sub(r'\{(\w+)\}', self.replace_non_numeric_bracket, tail)
-        output_template = output_str.format(*self.output_info.template_folders)
+        try:
+            output_template = output_str.format(*self.output_info.template_folders)
+        except (KeyError, ValueError, IndexError):
```
shail-parekh: This block passes the template to `grib_copy`, which ignores the python datetime calculation and just outputs literal strings as the filenames - please fix this!

weather_sp/splitter_pipeline/file_splitters.py
shail-parekh, line 1: When using datetime formatting in the output template like weather_{date:%Y%m%d_%H%M} or {datetime.strptime(f"{time}", "%H%M").strftime("%H:%M:%S")}.grib, the output filename doesn't evaluate as expected and just outputs the literal string. Can you please test this out?

PR #555, opened 2026-07-29
Added support to provide nocache parameter in Config

DarshanSP19:
Add support to provide a `nocache` to fetch new data instead of a cached one. (This is currently specific to the `CDS` API). The value can only be true or false. The error will be thrown in case of unsupported value. 
Examples for supported values 
```
[selection]
no_cache=true
```

```
[selection]
no_cache=false
```


weather_dl/download_pipeline/config.py
```diff
@@ -123,7 +122,9 @@ def optimize_selection_partition(selection: t.Dict) -> t.Dict:
         del selection_['day']
         del selection_['month']
         del selection_['year']
-    if 'nocache' in selection_:
-        selection_['nocache'] = int(time.time())
+    if 'no_cache' in selection_:
```
mahrsee1997: can we move this to parsers.py ?

mahrsee1997 (changes requested):
Couple of changes ...
(1) We need to do similar changes for weather-dl v2.
(2) Bump appropriate setup version

weather_dl/download_pipeline/config.py
DarshanSP19: Done. Update the versions as well.

mahrsee1997 (approved):
LGTM!

PR #557, opened 2026-08-05
Fix for year-month partitioning

ksic8:
Fix for year-month partitioning


weather_dl/download_pipeline/partition.py
```diff
@@ -138,6 +139,13 @@ def _create_partition_config(option: t.Tuple, config: Config) -> Config:
     for idx, key in enumerate(config.partition_keys):
         copy[key] = [option[idx]]
 
+    # Replace year-month with actual dates.
+    if 'year-month' in copy:
```
DarshanSP19: We should add this change in `optimize_selection_partition`. (Here https://github.com/google/weather-tools/blob/main/weather_dl/download_pipeline/config.py#L80)

weather_dl/download_pipeline/fetcher.py
```diff
@@ -69,6 +69,8 @@ def fetch_data(self, config: Config, *, worker_name: str = 'default') -> None:
 
         client = CLIENTS[self.client_name](config, self.log_level)
         target = prepare_target_name(config)
+        if 'year-month' in config.selection:
```
DarshanSP19: The same can be moved to the same `optimize_selection_partition`. (https://github.com/google/weather-tools/blob/main/weather_dl/download_pipeline/config.py)

weather_dl_v2/fastapi-server/config_processing/partition.py
```diff
@@ -69,6 +70,13 @@ def _create_partition_config(self, option: t.Tuple) -> Config:
         for idx, key in enumerate(self.config.partition_keys):
             copy[key] = [option[idx]]
 
+        # Replace year-month with actual dates.
```
DarshanSP19: Same as `weather_dl`.

DarshanSP19 (changes requested):
Ideally we're modifying this date selections in `optimize_selection_partition` in this [config.py](https://github.com/google/weather-tools/blob/main/weather_dl/download_pipeline/config.py) file. So changing the key of a selection and deletion of the unsupported key can be done via the same method. This will also help us when we do client specific changes in future.

DarshanSP19 (approved):
LGTM

mahrsee1997 (commented):
@ksic8 

Need to update VERSION.txt for license_deployement as well. Please raise a follow up PR for the same.

cc: @DarshanSP19

PR #550, opened 2026-06-23
Fix Metview regrid memory leak

j9sh264:
## Problem
During the execution of the Dataflow regrid pipeline, we observed persistent memory accumulation on the worker nodes, eventually leading to `MemoryLimitExceeded` errors. The memory leak was isolated to the `fs.base_date()` method in the `metview-python` package. When extracting the year for year-wise directories, `fs.base_date()` triggers the allocation of C-level memory structures (via the underlying Metview/ecCodes bindings) that are not properly released or tracked by Python's Garbage Collector.

## Solution
This PR completely bypasses the leaky `base_date()` implementation by directly extracting primitive data types from the GRIB keys.

## Key Changes:

* Added `get_safe_base_date()` helper: Replaces the native `fs.base_date()` call. It utilizes `fs.grib_get(["dataDate", "dataTime"])` to fetch the raw date/time strings directly from the ecCodes engine.
* Because `grib_get` returns standard Python primitives (strings/lists), Python's Garbage Collector can effortlessly track and destroy them when they go out of scope.
* We then manually parse these safe primitives into Python `datetime` objects using Metview's built-in `utils.date_from_ecc_keys(d, t)`.

PR #561, opened 2026-08-27
test(regrid): Add robust OS-level memory leak tracking to Regrid pipeline

j9sh264:
# Description
Previously, we identified a C-level memory leak in our code caused by the native `fs.base_date()` method when processing GRIB files, which was fixed in #550. This PR introduces automated tests to explicitly check for and catch any such instances of C-level memory leaks in the **Regrid** pipeline beforehand.

# Testing Performed
* Verified that reverting the fix from #550 (using `fs.base_date()`) correctly triggers the new `AssertionError` due to ballooning memory.
* Verified that using the patched `get_memory_safe_base_date(fs)` keeps memory growth under the 10.0 MB threshold across multiple test iterations.


mahrsee1997 (approved):
LGTM!

